PR #103, opened 2021-02-25
Feature/input system

LukeStampfli:
This replaces legacy inputs with the new Unity Input System packages.

Inputs are no longer polled in Update but FixedUpdate. It still uses the queue system which we had in place to combine inputs from the UI buttons and queue them up properlyi.. 

- Changes to scene files are only updating the EventSystem. Scenes itself are unchanged.
- Had to fix some of the GameData assets because IntVariables where unassigned.


LukeStampfli:
Closing this as its stale and will probably need at least a rebase once we decide to implement this.

PR #175, opened 2021-03-16
Feature/animation anticipation

dwoodruffsf:
* Introduces Action Anticipation System. Lengthy comment explains the system in ActionVisualization.AnticipateAction
* Changed AnimationOnlyActionFX to implement Cancel rather than End. This ensures that the 2nd animation always fires. This wasn't to fix a specific problem, but I was concerned that if I didn't guarantee "Animation 2" plays, then the character could get stuck in an animation state node. 
* Not all Actions opt in to the anticipation system--actions like ChargedActionFX and AoeActionFX have only been changed to call base.Start, per the new contract. These are todos that I didn't think reasonably fit in the scope of this change. 
* TargetActionFX now actively listens for Action Inputs. If it sees one for a target, it pre-emptively changes its target ID to the new ID. This makes targeting enemies happen instanteously, as far as visual feedback goes. 
* ClientCharacterVisualization can now tell if you're doing something besides just idling. You must just be idling in order to play an animation anticipatively.
* ClientCharacterVisualization now playes the "AnticipateMove" animation queue if not already animating (this does work properly with holding the mouse down).
* The "Rest States" of the Character Controller layers have been changed to the name "Nothing" to make them deducible in code (except for the base layer, which is special). 
* Fixed bug where you can only update your movement input at 2fps. I did leave it capped though (at 20fps). I felt leaving it uncapped didn't represent best practice. 
* (ServerCharacter). Target actions now cancel on move. If you don't have this, there's no way to cancel your target short of killing it, or targeting yourself.


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=175) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=175) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=175) it.</sub>

SamuelBellomo (approved):

pdeschain (approved):

PR #223, opened 2021-03-28
[GOMPS-417][GOMPS-379] mage-bolt improvements

eheimburg:
The description I have for this task is: 

> - If I shoot while walking, the projectile spawns behind you even in host mode
> - When the bolt hits the target, it can hit but the visuals will show it as missed

This PR fixes these issues and cleans up some related timing problems, but there are other bugs underneath it that I couldn't completely address. I've been trying to condense the issue, but it's still kind of complicated, so bear with me while I unpack it!

First, I fixed the graphical issue at the heart of this task: the code assumed that the character would be stationary while firing the projectile, so it spawned the projectile immediately (in other words, before `ExecTimeSeconds` had elapsed) in order to show special "power up particles" prior to actually firing. This meant that if the player had moved since firing, the projectile seemed to start in the wrong place, and didn't go far enough to reach the enemy. We didn't end up using "power up particles", so I ripped all that stuff out, simplifying the code. The projectile now visually starts at the right spot and end at the right spot. This fixes the PR task...

But in testing I realized that there was another major bug: if you issued a character-move command while the projectile is in the air, the projectile aborted. This is because the action was blocking for ExecTime+CooldownTime, and CooldownTime hadn't elapsed yet. **But it didn't cancel the graphics on the client!** So it looked like you were shooting an enemy and it just wasn't doing any damage. Extremely frustrating.

There's no easy fix for this, because the notion of "blocking during cooldown" doesn't work for this action. It can only block for ExecTime, because after that, the projectile is already in the air, and it's a bug for it to be aborted! Once ExecTime has elapsed, it must become a background action so that it can't be aborted. 

So I've changed the action's blocking mode to "Only During Exec Time". This means the mage can shoot a bolt every 0.5 seconds, though (the ExecTime), and I don't have a way to make the mage-bolt have a slower re-casting time. (And having an ExecTime longer than 0.5s looks dumb.)

(To be fair, I don't this notion of "block on cooldown" has ever really worked well for player actions, because the player can always just enqueue a different action, which aborts the action on cooldown, and then they can use that action again immediately. So I think we need a more complex cooldown system for players regardless... although the current system **does** work for NPCs, because they only try to enqueue new actions when they're out of blocking actions.)

* I fixed the scenario where the action is canceled on the server and doesn't get canceled on the client: it now cancels the client graphics also. (That is, if you changed the blocking mode back so that it started cancelling in-the-air bolts again, it would at least cancel them on both server and client.)
* I also sped up the mage-bolt graphic so that it moves at the intended speed. It just looks much better

Edit: addendum: to better address this problem, I added a new field, `ReuseTimeSeconds`, which acts as a cooldown timer on the action. The mage-bolt has a reuse time of 1 second. 

Edit: Also added fix for the "mage bolts go through walls" issue.


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=223) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=223) before we can accept your contribution.<br/>**0** out of **2** committers have signed the CLA.<br/><br/>:x: dwoodruffsf<br/>:x: eheimburg<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=223) it.</sub>

SamuelBellomo:
Updated the base branch to develop, please update your branch accordingly

eheimburg:
I removed fenando-cortez's suggested workaround and instead added a new field to ActionDescription: `ReuseTimeSeconds`. The server checks this, and discards any repeat actions that happen faster than this. I set the mage-bolt's `ReuseTimeSeconds` to 1. This prevents players from starting mage-bolt attacks more often than once per second.

(We may need to use this field on other abilities, like the archer's power-shot, to prevent infinite-spamming.)

eheimburg:
Also, since this has become the "make mage-bolt work" PR, I also fixed [GOMPS-379]: Mage ranged attack passes through walls. The fix was to do a line-of-sight check at the start of the action.

eheimburg:
Okay I think this one's ready for re-review.

Assets/BossRoom/Scripts/Server/Game/Action/ActionPlayer.cs
```diff
@@ -109,24 +112,37 @@ private void StartAction()
         {
             if (m_Queue.Count > 0)
             {
+                float reuseTime = m_Queue[0].Description.ReuseTimeSeconds;
+                if (   reuseTime > 0
```
fernando-cortez, line 116: I'm not 100% sure if this is the correct place to check but I did encounter a bug with the mage bolt.

If you spam left & right click simultaneously you'll see that the animation & sound fire. but no bolt is created.

eheimburg: Nice find! This is a bug in the input-processing code path, and not this PR per se. It happens with all classes' basic attacks, most easily noticed with the archer, who does the animation but no projectile comes out. I think this is also happening to the tank and rogue's melee attacks, and contributes to why those attacks don't always seem to "hit": they are actually being cancelled almost instantly after they begin, but the animation trigger has already fired. 

(Prior to this PR, the client didn't correctly cancel the mage's graphics, so the bolt was rendered on the client even though it had already been canceled on the server.)

Can you open a bug on this? (I don't have direct bug access.) I'd like to fix this in a separate PR if possible. I hacked in a quick-fix locally, but I'm not confident that it doesn't cause other problems. Will need to get David's feedback on the proper solution.

eheimburg: oh right, @dwoodruffsf ^^ FYI on a bug that needs fixing

fernando-cortez: I made a bug [task](https://unity3d.atlassian.net/browse/GOMPS-467) to track this.

fernando-cortez (dismissed):

pdeschain (dismissed):

pdeschain (approved):

fernando-cortez (approved):

PR #263, opened 2021-04-07
Refactor for stealth state, should use bool, not byte

SamuelBellomo:
This missing fix slipped through and gives the wrong message that NetworkVariableBool is broken (which is not the case). Removing that usage now to not mislead users.


LukeStampfli (approved):
Confirm that this works no in MLAPI, the refactor is correct and the comments have been correctly removed.

pdeschain (approved):

eheimburg:
@SamuelBellomo Did this get merged into the intended branch? I think it should be in `develop` as well?

SamuelBellomo:
> @SamuelBellomo Did this get merged into the intended branch? I think it should be in `develop` as well?

@fernando-cortez will look into merging master back to develop.

PR #271, opened 2021-04-12
(Cleanup) (feedback needed) Remove redundant list of ServerCharacters

eheimburg:
Edit: moving the static list of `ServerCharacters` into a separate class that ONLY tracks the `ServerCharacters` for players. Refactored code that iterates players to use this new list.

---
`ServerCharacter` had a static list of all active `ServerCharacters`. This turns out to be basically redundant with MLAPI's `NetworkSpawnManager` which already provides a list of these. The static list had already been phased out in all but one place, so I've refactored the last remaining spot.

But note that this code is slower than before: `NetworkSpawnManager` gives us a listing of `NetworkObject`s, so we have to call `GetComponent<ServerCharacter>()` on each one to get the actual character info. And unfortunately the last remaining use of this list is in code that gets called many times every frame. 

So I refactored the code to do other checks first before calling `GetComponent<>()`, to minimize the number of calls needed. But even so, it's definitely not a performance improvement. But it does make the code cleaner, and this is supposed to be demo code, so that's important.

So feedback is needed! Should we go this route? Alternatively we could remove the comments that say the `ServerCharacter` list is "temporary" and commit to using it in the various places that the server needs a list of `ServerCharacter`s. (There are only a few other places.)

### DIFF
--- a/Assets/BossRoom/Scripts/Server/Game/AIState/IdleAIState.cs
+++ b/Assets/BossRoom/Scripts/Server/Game/AIState/IdleAIState.cs
@@ -1,3 +1,6 @@
+using MLAPI;
+using MLAPI.Connection;
+using System.Collections.Generic;
 using UnityEngine;
 
 namespace BossRoom.Server
@@ -31,13 +34,25 @@ namespace BossRoom.Server
             float detectionRange = m_Brain.CharacterData.DetectRange;
             // we are doing this check every Update, so we'll use square-magnitude distance to avoid the expensive sqrt (that's implicit in Vector3.magnitude)
             float detectionRangeSqr = detectionRange * detectionRange;
-            Vector3 position = m_Brain.GetMyServerCharacter().transform.position;
+            Vector3 myPosition = m_Brain.GetMyServerCharacter().transform.position;
 
-            foreach (ServerCharacter character in ServerCharacter.GetAllActiveServerCharacters())
+            // in this game, NPCs only attack players (and never other NPCs), so we can just iterate over the connected players to see if any are nearby
+            foreach (var networkClient in NetworkManager.Singleton.ConnectedClientsList)
             {
-                if (m_Brain.IsAppropriateFoe(character) && (character.transform.position - position).sqrMagnitude <= detectionRangeSqr)
+                if (networkClient.PlayerObject == null)
                 {
-                    m_Brain.Hate(character);
+                    // skip over any connection that doesn't have a PlayerObject yet
+                    continue;
+                }
+
+                if ((networkClient.PlayerObject.transform.position - myPosition).sqrMagnitude <= detectionRangeSqr)
+                {
+                    // they're in range! Make sure that they're actually an enemy
+                    var serverCharacter = networkClient.PlayerObject.GetComponent<ServerCharacter>();
+                    if (serverCharacter && m_Brain.IsAppropriateFoe(serverCharacter))
+                    {
+                        m_Brain.Hate(serverCharacter);
+                    }
                 }
             }
         }
--- a/Assets/BossRoom/Scripts/Server/Game/AIState/IdleAIState.cs
+++ b/Assets/BossRoom/Scripts/Server/Game/AIState/IdleAIState.cs
@@ -33,11 +33,18 @@ namespace BossRoom.Server
             float detectionRangeSqr = detectionRange * detectionRange;
             Vector3 position = m_Brain.GetMyServerCharacter().transform.position;
 
-            foreach (ServerCharacter character in ServerCharacter.GetAllActiveServerCharacters())
+            foreach (var spawnedObject in MLAPI.Spawning.NetworkSpawnManager.SpawnedObjectsList)
             {
-                if (m_Brain.IsAppropriateFoe(character) && (character.transform.position - position).sqrMagnitude <= detectionRangeSqr)
+                if (!spawnedObject) { continue; } // must have been Destroy()ed very recently
+                if ((spawnedObject.transform.position - position).sqrMagnitude <= detectionRangeSqr)
                 {
-                    m_Brain.Hate(character);
+                    // they're within range... are they an appropriate foe?
+                    ServerCharacter serverChar = spawnedObject.GetComponent<ServerCharacter>();
+                    if (!serverChar) { continue; } // not even a character at all...
+                    if (m_Brain.IsAppropriateFoe(serverChar))
+                    {
+                        m_Brain.Hate(serverChar);
+                    }
                 }
             }
         }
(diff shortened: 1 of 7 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=271) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=271) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=271) it.</sub>

Assets/BossRoom/Scripts/Server/Game/AIState/IdleAIState.cs
```diff
@@ -33,11 +34,18 @@ protected void DetectFoes()
...
-            foreach (ServerCharacter character in ServerCharacter.GetAllActiveServerCharacters())
+            foreach (var spawnedObject in NetworkSpawnManager.SpawnedObjectsList)
             {
-                if (m_Brain.IsAppropriateFoe(character) && (character.transform.position - position).sqrMagnitude <= detectionRangeSqr)
+                if (!spawnedObject) { continue; } // must have been Destroy()ed very recently
+                if ((spawnedObject.transform.position - position).sqrMagnitude <= detectionRangeSqr)
                 {
-                    m_Brain.Hate(character);
+                    // they're within range... are they an appropriate foe?
+                    ServerCharacter serverChar = spawnedObject.GetComponent<ServerCharacter>();
```
fernando-cortez, line 43: So just with a quick deep profile, this change allocates ~50 KB of memory per frame as it's called ~300 times. Let's keep the old approach with the static collection of `ServerCharacter`s and refactor the places in code to use this to be consistent.

eheimburg: sounds good

eheimburg: funnily enough I don't see any other uses that need to be switched to use this list. The other uses were scenarios where we needed to iterate over players (not just all characters), so after refactoring, they use `NetworkManager.Singleton.ConnectedClients` instead. And since that list is much shorter than the list of all characters, that's probably the more efficient approach for them.

So in the end this PR is just some comment clean up!

SamuelBellomo: Why is it iterating over all characters? Why not just players?

eheimburg: Hm... actually yeah, it can be. Back when we started, I wanted to support the notion of "pets" (e.g. a "sentient projectile" that could attract enemy attention), so enemies search for "foes" rather than just players. But none of that has ended up being relevant. 

Unless you object I'll go ahead and change it to a player-list iteration. That obsoletes the list of ServerCharacters, so I'll just remove that.

SamuelBellomo: thank you

eheimburg: ok

eheimburg:
Okay, it was pointed out that we don't need to iterate on `ServerCharacter`s at all. I've made `IdleAIState` iterate over player connections, and removed the cached `ServerCharacter` list.

This does mean that in the worst case, imps perform up to 8 `GetComponent<>` calls during Update (if 8 players all came within range at the same time). But it only happens once: as soon as the imp has found a foe, the AI stops being in Idle mode, and no longer searches for enemies. 

So I think overall this is a performance improvement, just because we're doing so many fewer distance-checks total. And we get rid of the weird `ServerCharacter` list.

Assets/BossRoom/Scripts/Server/Game/AIState/IdleAIState.cs
fernando-cortez: We want to eliminate all possible GC allocations wherever possible. There may be a solution here that fits both needs.

How about enforcing composition here? We attach a `PlayerServerCharacter` component (that requires a `ServerCharacter`) on our player prefab. This defines a publicly static collection like before. This component specifically adds and removes players to this collection on OnEnable()/OnDisable() respectively. That way you'd be getting a reference to all & only player `ServerCharacter`s without any `GetComponent<>`s.

eheimburg: Okay, I've added the `PlayerServerCharacter` component to the players' prefab and refactored `IdleAIState` (and a few other places) to use this list.

Assets/BossRoom/Scripts/Server/Game/Character/PlayerServerCharacter.cs
```diff
@@ -0,0 +1,44 @@
...
+    [RequireComponent(typeof(ServerCharacter))]
+    public class PlayerServerCharacter : NetworkBehaviour
+    {
+        static List<ServerCharacter> s_ActivePlayers = new List<ServerCharacter>();
+
+        ServerCharacter m_CachedServerCharacter;
+
+        void OnEnable()
+        {
+            m_CachedServerCharacter = GetComponent<ServerCharacter>();
```
fernando-cortez, line 27: Just last comment to serialize & cache this and we're good to go.

dwoodruffsf (approved):

Assets/BossRoom/Scripts/Server/Game/Character/PlayerServerCharacter.cs
```diff
@@ -0,0 +1,44 @@
...
+using MLAPI;
+using UnityEngine;
+
+namespace BossRoom.Server
+{
+    /// <summary>
+    /// Attached to the player-characters' prefab, this maintains a list of active ServerCharacter objects for players.
+    /// </summary>
+    /// <remarks>
+    /// This is an optimization. In server code you can already get a list of players' ServerCharacters by
```
SamuelBellomo, line 12: should add quick comment on how this should be done with SOs in a second pass

Assets/BossRoom/Scripts/Server/Game/Character/PlayerServerCharacter.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// iterating over the active connections and calling GetComponent() on their PlayerObject. But we need
+    /// to iterate over all players quite often -- the monsters' IdleAIState does so in every Update() --
+    /// and all those GetComponent() calls add up! So this optimization lets us iterate without calling
+    /// GetComponent().
```
fernando-cortez, line 16:
```suggestion
    /// GetComponent(). This will be refactored with a ScriptableObject-based approach on player collection.
```

Assets/BossRoom/Scripts/Server/Game/Character/PlayerServerCharacter.cs
dwoodruffsf: done. I realized this class is also not properly disabling itself on client, so I changed that as well.

SamuelBellomo (approved):

fernando-cortez (approved):

PR #272, opened 2021-04-12
(Cleanup) Isolate testing commands

eheimburg:
Moves the testing commands that were hacked into `ServerBossRoomState` into their own separate component on the same prefab.

### DIFF
--- a/Assets/BossRoom/Scripts/Server/ServerTestingHotkeys.cs
+++ b/Assets/BossRoom/Scripts/Server/ServerTestingHotkeys.cs
@@ -11,7 +11,7 @@ namespace BossRoom.Server
     /// </summary>
     public class ServerTestingHotkeys : NetworkBehaviour
     {
-#if DEBUG
+#if UNITY_EDITOR || DEVELOPMENT_BUILD
         [SerializeField]
         [Tooltip("Enemy to spawn. Make sure this is included in the NetworkManager's list of prefabs!")]
         private NetworkObject m_EnemyPrefab;
--- a/Assets/BossRoom/Scripts/Server/ServerTestingHotkeys.cs
+++ b/Assets/BossRoom/Scripts/Server/ServerTestingHotkeys.cs
@@ -1,0 +1,67 @@
+using System.Collections;
+using System.Collections.Generic;
+using MLAPI;
+using UnityEngine;
+
+namespace BossRoom.Server
+{
+    /// <summary>
+    /// Provides various special commands that the Host can use while developing and
+    /// debugging the game. Remove this component before releasing your game! (So that
+    /// regular players can't spawn extra bosses etc.)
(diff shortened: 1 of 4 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=272) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=272) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=272) it.</sub>

Assets/BossRoom/Scripts/Server/ServerTestingHotkeys.cs
```diff
@@ -0,0 +1,67 @@
...
+
+namespace BossRoom.Server
+{
+    /// <summary>
+    /// Provides various special commands that the Host can use while developing and
+    /// debugging the game. Remove this component before releasing your game! (So that
+    /// regular players can't spawn extra bosses etc.)
+    /// </summary>
+    public class ServerTestingHotkeys : NetworkBehaviour
+    {
```
fernando-cortez, line 14: What I did for previous projects was wrap everything inside a testing class with a #UNITY_DEBUG preprocessor directive. So essentially everything inside the class compiles only inside the editor and development builds, but not on release builds. The class itself always compiles, it's just empty on release. @pdeschain any objections?

eheimburg: It doesn't seem like UNITY_DEBUG is pre-defined. Did you mean DEBUG? That seems to be auto-defined by Unity in the editor and in development builds, but not in release builds. (I've checked in that change for consideration.)

eheimburg: Reading a bit more, seems the DEBUG symbol is coming from C# itself based on lower-level compilation directives. The best way I see to make our intent really explicit is to do `#if UNITY_EDITOR || DEVELOPMENT_BUILD`   so I switched to that.

Tested in editor, in develop builds, and in release builds, and works as intended.

fernando-cortez: Ah yes, I did mean #DEBUG. But I like how explicit this is to a user.

fernando-cortez (approved):

SamuelBellomo (approved):

PR #273, opened 2021-04-12
Bugfix: Add move-check timeout to untargeted Actions [GOMPS-467]

eheimburg:
Edit: after various back-and-forth, the crux of this PR is that we now allow players to move while performing regular actions: movement doesn't clear the action queue. The exception is "Interruptible" actions (like rogue's stealth or archer's charged shot). Those actions are now interruptible by movement as well as by other actions. We also abort any movement that's active when interruptible actions start, to ensure that the player can't perform that action while moving.

---

This addresses a problem @fernando-cortez found during testing of mage-bolt: if you click both the left and right mouse buttons at the same time, the character appears to use their base-attack, but doesn't really. 

(This bug happened for all classes, but is most easily visible with the Archer class: pressing left+right mouse buttons made the archer play an animation but did not fire an arrow.)

At the time I feared it was a deep server timing bug and asked to do it in a separate PR. But now that I've explored it more, it's just an input-timing problem: pressing both buttons sent two Actions at once, and one of them got cancelled. The fix is to use an input timeout like we do elsewhere in the same class. Edit: now we're using a server-side fix; see comments below.


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=273) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=273) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=273) it.</sub>

Assets/BossRoom/Scripts/Client/Game/Character/ClientInputSender.cs
```diff
@@ -219,15 +219,22 @@ private void PerformSkill(ActionType actionType, SkillTriggerStyle triggerStyle,
...
             else if(actionType != ActionType.GeneralTarget )
             {
-                // clicked on nothing... perform a "miss" attack on the spot they clicked on
+                // clicked on nothing... perform an "untargeted" attack on the spot they clicked on.
+                // (Different Actions will deal with this differently. For some, like archer arrows, this will fire an arrow
+                // in the desired direction. For others, like mage's bolts, this will fire a "miss" projectile at the spot clicked on.)
+
+                // Don't trigger our move logic for a while. This protects us from moving immediately after
+                // starting the untargeted action (which would abort the action on the server)
+                m_LastSentMove = Time.time + k_TargetMoveTimeout;
```
fernando-cortez, line 236: So while this does take care of the initial bug that was found on the mage bolt PR, I'm finding this change to be a little restrictive on gameplay. Spamming right click and left click will essentially render you stationary. Right now, as an archer or a mage, I can't pick a target that is not in my path to shoot at. As a tank, I cant swing and choose a new location (in quick successions).

It may be that the cancellation of actions when a new action is requested is the cause of this limitation in gameplay. I would expect _some_ actions to be played along with a move & some to be cancelled outright.

Again, this PR does the thing it's meant to do but we're limiting ourselves with the flow of actions here. I'd want @dwoodruffsf & @SamuelBellomo to chime in to get more perspective.

eheimburg: I see what you mean by the delay. I've removed the input-delay and changed the server's move-handler to only cancel actions if they are marked as interruptible.

This fixes the problem, but I'm not 100% sure that there aren't subtle ramifications that I can't predict. Would like more eyes on this approach in case others see problems with this! @dwoodruffsf

fernando-cortez: After speaking with @dwoodruffsf, we were hoping if this change effectively makes `CooldownSeconds` obsolete and if it can be removed entirely. I'm thinking it can be, and `ReuseTimeSeconds` should be updated for the Boss' charge action. If you want to look that over for another PR, I'll approve this for now but keep this comment up to come back to.

fernando-cortez (approved):

Cosmin-B (dismissed):
Looks good to me!

Assets/BossRoom/Scripts/Client/Game/Character/ClientInputSender.cs
eheimburg: There's one counter-example that's really important: rogues should not be allowed to move while performing their Stealth action. If they can move while going into stealth, they're basically invincible. (Someone has marked Revive and Charged Shot as Interruptible too, which probably make sense for the same reason.)

So I don't think we can just go to a "you can move while doing any actions" solution. But the existing code is definitely weird in how it deals with enqueued actions.

The proposed code happens to work for Stealth, but I now realize that's only because Stealth explicitly issues a "Stop moving" command when it first starts. Then the code above correctly interrupts it if you try to move while casting. I think this general idea makes the most sense: any interruptible action should also be interrupted by movement, and thus we should implicitly call `CancelMove()` on interruptible actions when they start.

That way the above code effectively says "Want to move? Sure, but if your active action is interruptible, it's getting tossed. Not interruptible? Fine, we'll move while we play it. Have other actions enqueued, regardless of their type? Well, if they're interruptible, movement will be canceled when they begin."

This also lets me remove the explicit `CancelMove()` from StealthAction, since it's implicit.

eheimburg:
After some starts and stops, the current point of this PR is that we now allow players to move while performing regular actions: movement doesn't clear the action queue. 

The exception is "Interruptible" actions (like rogue's Stealth or archer's Charged Shot). Those actions are now interruptible by movement as well as by other actions. When an interruptible action starts, we abort any active movement, and if they perform a new move while the action is going, it's interrupted.

Assets/BossRoom/Scripts/Client/Game/Character/ClientInputSender.cs
eheimburg: > After speaking with @dwoodruffsf, we were hoping if this change effectively makes `CooldownSeconds` obsolete and if it can be removed entirely. I'm thinking it can be, and `ReuseTimeSeconds` should be updated for the Boss' charge action. If you want to look that over for another PR, I'll approve this for now but keep this comment up to come back to.

(Forgot to address this!) I suspect it can be removed also. Let me tackle that in a separate PR, just to make the data-wrangling a little cleaner. Please let me know the task # if you add one for it!

Cosmin-B (approved):

Assets/BossRoom/Scripts/Client/Game/Character/ClientInputSender.cs
dwoodruffsf: Yeah, it's a little surprising that "non-interruptible" actions can happen during movement, but interruptible actions can't. Maybe it shouldn't be, though. Non-interruptibles are short--things like a basic melee attack--compared to long actions like charged shot. Thinking about it that way it's easier to see why this pattern makes sense. 

In any case, I'm OK with it. check from me.

dwoodruffsf (approved):

Assets/BossRoom/Scripts/Client/Game/Character/ClientInputSender.cs
fernando-cortez: I'm okay with this as well. My only final thought is, is everything discussed here captured in comments/documentation/architecture clearly enough for a dev to understand this? If it's not then, we need to update scripts with these insights, namely the great summary:

> Edit: after various back-and-forth, the crux of this PR is that we now allow players to move while performing regular actions: movement doesn't clear the action queue. The exception is "Interruptible" actions (like rogue's stealth or archer's charged shot). Those actions are now interruptible by movement as well as by other actions. We also abort any movement that's active when interruptible actions start, to ensure that the player can't perform that action while moving.

eheimburg: In terms of documenting it I think we're in decent shape; the variable has this explanation:

```
[Tooltip("Is this Action interruptible by other action plays or by movement? (Implicitly stops movement when action starts.) Generally, actions with short exec times should not be interruptible in this way.")]
public bool ActionInterruptible;
```

And where we call `CancelMove()`, we have 

`  // if this Action is interruptible, that means movement would interrupt it. So stop any active movement!`

eheimburg: Actually I made a very small tweak to the latter comment:

`// if this Action is interruptible, that means movement should interrupt it. So stop any movement that's already happening before we begin
`

Cosmin-B (approved):
Approving this again, as I suppose as the conversations are resolved.

fernando-cortez (commented):
Great additions to the comments. Fix that merge conflict, and we're good to go.

fernando-cortez (approved):

PR #280, opened 2021-04-14
Feature/Rogue Dash Attack

eheimburg:
Implements rogue's second special ability, the Dash Attack: the rogue dashes to the target spot and performs a melee attack. While dashing, the rogue cannot suffer damage. (But they can still be damaged while "winding up" the attack, as indicated by the action's `ExecTimeSeconds`.)

~~This is implemented using the same "charge forward" functionality that the Trample action uses: the character moves in a straight line toward their target, as opposed to navmesh pathing. So if the player clicks on the other side of a wall, their character charges into the wall (and then performs a harmless melee attack against the wall.)~~

Edit: On the server this is implemented as a delayed teleport followed by a melee attack at the new location. On the client, we move the local visualization so that the character appears to be zooming across the screen.

*Implementation Notes*
- This uses final artwork that was already in the project. (Added `SpecialFXGraphic` components to each prefab, but no other changes)
- Since this performs the same melee attack as `MeleeAction`, it should ideally use the same code. Added a static reusable function to `MeleeAction` that other actions can call. (It's a little bit weird to stick it here, but I'm not sure where else to put it. It can't go in the `ActionUtils` class because that's in shared code, and we need to work with `IDamageable` which is server-only.)
- This new utility function also addresses the problem that melee attacks can be "wasted" on dead enemies (if a living and dead enemy are both in the targeted area, one was arbitrarily chosen even if it was dead).
- To help address this, added a simple "is it dead" function to IDamageable

*Merging Notes: CharacterSetController*
- Added two new triggers: "DashAttackStart" and "DashAttackEnd"
- Added three nodes to Attacks Layer: "Dash Attack (start)", "Dash Attack (loop)", "Dash Attack (end)". 
- Transition from Nothing->Dash Attack (start) on trigger DashAttackStart. No exit time
- Transition from Dash Attack (start)->Dash Attack (loop). Uses exit time
- Transition from Dash Attack (loop)->Dash Attack (end) on trigger DashAttackEnd. No exit time
- Transition from Dash Attack (end)->Nothing. Uses exit time
- Edit: Added another trigger, "DashAttackCanceled"
- Edit: Added additional transitions from all three new nodes, sending us back to the start node if DashAttackCanceled. No exit time


eheimburg:
One of my last "important things to fix" was the problem with melee hits ending up being consumed by already-dead enemies. Since this new Action also performs a melee-hit, I added that functionality here, just to make merging and maintenance easier.

Assets/BossRoom/Scripts/Server/Game/Action/MeleeAction.cs
```diff
@@ -64,36 +64,47 @@ public override bool Update()
...
+        /// Utility used by Actions to perform Melee attacks. Performs a melee hit-test
+        /// and then looks through the results to find an alive target, preferring the provided
+        /// enemy.
+        /// </summary>
+        /// <param name="isNPC">true if the attacker is an NPC (and therefore should hit PCs). False for the reverse.</param>
+        /// <param name="attacker">The collider of the attacking GameObject.</param>
+        /// <param name="range">The range in meters to check for foes.</param>
+        /// <param name="preferredTargetNetworkId">The NetworkObjectId of our preferred foe, or 0 if no preference</param>
+        /// <returns></returns>
+        public static IDamageable GetIdealMeleeFoe(bool isNPC, Collider ourCollider, float meleeRange, ulong preferredTargetNetworkId)
```
fernando-cortez, line 86: Make sure all params are included in the summary params and are contextual.

Assets/BossRoom/Scripts/Server/Game/Action/DashAttackAction.cs
```diff
@@ -0,0 +1,127 @@
...
+        public DashAttackAction(ServerCharacter parent, ref ActionRequestData data) : base(parent, ref data)
+        {
+            Assert.IsTrue(Description.MoveSpeed > 0, $"ActionDescription for {Description.ActionTypeEnum} needs a MoveSpeed assigned!");
+            Assert.IsTrue(Description.Radius > 0, $"ActionDescription for {Description.ActionTypeEnum} needs a Radius assigned!");
+        }
+
+        public override bool Start()
+        {
+            // snap to face our destination. This is the direction we'll charge in
+            m_Parent.transform.LookAt(GetTargetSpot());
```
fernando-cortez, line 37: Finding some weird behaviour when this is fired when you are stationary and press the dash button on the UI. You don't have a target, so this forces you to turn towards `Data.Position` which turns out to be just your root transform's position, which will force the player to turn around and do a micro dash. I would expect the character to dash forward in the direction they're facing to some minimum distance, no? 

I also just played as the archer and this is the case too with their Attack1 skill. They turn around and shoot an arrow when stationary. Tested with a mage and this happens with their Attack1 skill as well. Seems like a shared bug with ranged attacks.

Assets/BossRoom/Scripts/Server/Game/Action/DashAttackAction.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            // tell clients to visualize this action
+            m_Parent.NetState.RecvDoActionClientRPC(Data);
+            return true;
+        }
+
+        public override bool Update()
+        {
+            if (!m_StartedDash && TimeRunning >= Description.ExecTimeSeconds)
```
SamuelBellomo, line 46: This should be an instant action. An RPC is sent to the client. This is in opposition to the boss charge which moves your character over time, here an RPC that can't be corrected teleports your player.

Cosmin-B:
Personally, I don't like some architectural decisions here, but there is nothing that could be done in regards to that in this PR. Overall, the implementation of Dash Attack looks solid to me, including what has been mentioned above!

Assets/BossRoom/Scripts/Server/Game/Action/DashAttackAction.cs
eheimburg: My understanding is that if you click the action buttons it will use the last clicked location. So once you've clicked anywhere once, the attacks will aim at that spot. I'm not sure how we'd address the corner-case of not having clicked on anything yet... it would seem to me to be non-trivial for many actions (e.g. should we just find the nearest target for mage-bolt? otherwise wherever we pick, it can't possibly connect...)

Assets/BossRoom/Scripts/Server/Game/Action/DashAttackAction.cs
eheimburg: That isn't what the art is set up for, and it's not what's is in the GDD. Since reimplementing it as a teleport doesn't improve its gameplay behavior in any way I can figure, I assume you want this rewritten in order to have an example of a networked teleport in the demo?

I'm happy to do so, but... I'm not sure how it's supposed to work. To avoid further delays I'd like to get answers to some design questions first! (Paging @dwoodruffsf, but figured I'd do this conversation here in the PR, since I assume @SamuelBellomo has expectations also.)

1- The artwork is designed to show the character dashing across the screen. Do we need new artwork, or do we show the character "pretend running" across the screen? (Presumably by having the ActionFX disable normal visualization of networked movement, and determining the visualization's location by lerping between start and end spot.)
2- Is this a fixed-duration teleport, or should we calculate the distance based on a speed variable, as we do with mage-bolt? (Fixed duration is much simpler in terms of implementing the visualization part.)
3- Can the rogue teleport through walls? (If not, presumably we implement the same LOS check as mage-bolt, and the character abruptly stops at the wall)
4- What does a new player see if they log in right as the action is happening? (Moving the visualization via an Action will mean that if a character joins right after a rogue begins, they will see the rogue standing in the old spot, before instantly blipping to the new spot. This is an unusual edge-case and I'd assume it's fine to ignore it, but want to confirm.)
5- Is it fine to teleport to the exact spot that the target is in, or do we need to try to figure out a spot "beside" the target? (E.g. teleporting to 1 meter away from the spot.) I can see both scenarios making sense: if you to teleport to the floor switch, for instance, you don't want to teleport to "near" the floor switch. But teleporting on top of the boss would be visually confusing.

eheimburg: Final thought: if we end up wanting to use new artwork for this, can I suggest we give the teleport to the mage instead? (Since they're down an action anyway.) 

Implementing a "pure teleport" would be a simpler demonstration of teleportation: presumably it would let you teleport through walls (like in Diablo 2), and it wouldn't need to do all the hacky "pretend to move the visualization" stuff; the character could just perform an animation at the start spot, then blink to the new spot after the RPC.

eheimburg: After internal communication, this is my understanding of what's desired:

1- we will show the character dash across the screen (lerping position)
2- fixed duration seems fine (so all dashes will take e.g. 1.5 secs)
3- ??? (will assume rogue cannot teleport through walls, for safety, as removing it is easier than adding it)
4- ignore edge case of players joining game in the 1.5secs that a teleport is happening
5- we will teleport to "near" the spot, not "on top of" the spot.

If Unity folks want alterations to this plan, please let me know ASAP! Will begin implementation of this tomorrow

Assets/BossRoom/Scripts/Server/Game/Action/MeleeAction.cs
eheimburg: ok

Assets/BossRoom/Scripts/Server/Game/Action/DashAttackAction.cs
eheimburg: Okay, I've revised this to be a teleport followed by a melee attack at the destination spot. (And the aforementioned client-side visualization of the character dashing.)

Using a teleport instead of a dash has some problematic gameplay behavior: prior to teleporting, the enemies will continue to perceive the character as being in their original spot. So if the "dash" part takes a long time, the imps seem to not understand where the character really is.

To minimize the impact of this, I've made the "dash" part extremely short, so that you zip across the screen in 0.7 seconds. During this 0.7 seconds the character is immune to damage, which also minimizes chances for weirdness. Prior to dashing, though, the character performs a wind-up animation for 1.3 seconds, during which they are NOT immune to damage.

I have not attempted to tweak the action's particles to match these new timings, so they may appear to be off by a few tenths of a second. But we can tweak those when the action's times are finalized.

Assets/BossRoom/Scripts/Server/Game/Action/DashAttackAction.cs
fernando-cortez: I was testing the new teleport dash and it's opened the way for speed hacks for sure. Spamming left click and 2 lets you move around the map in a very undesirable manner (I'm seeing the player get snapped to many different positions). I'm hoping `ReuseTime` can help us here. Also, is the action abiding by the range or is that set to a fairly high number? I'm able to zoom out and pick the farthest point on the map and traverse there.

Right, one solution will not work for all actions. If some actions require a target, is it worth displaying the last target position on the map, perhaps? This might help alleviate some confusion on where you're dashing, shooting at, etc.

Assets/BossRoom/Scripts/Client/Game/Action/DashAttackActionFX.cs
```diff
@@ -0,0 +1,98 @@
+using MLAPI;
```
SamuelBellomo, line 1: when I click to move the character left, hit the dash shortcut, then while ramping up I click to the left of the character I get a weird boomerang movement

Assets/BossRoom/Scripts/Server/Game/Character/ServerCharacterMovement.cs
```diff
@@ -128,6 +128,26 @@ public void CancelMove()
...
+            {
+                // warping failed! We're off the navmesh somehow. Weird... but we can still teleport
+                Debug.LogWarning($"NavMeshAgent.Warp({newPosition}) failed!", gameObject);
+                transform.position = newPosition;
+            }
+
+            m_Rigidbody.position = transform.position;
+            m_Rigidbody.rotation = transform.rotation;
+
+            m_NetworkCharacterState.RecvTeleportClientRpc(newPosition);
```
SamuelBellomo, line 148: I'm curious, why isn't this part of the recvDoClientRPC in the action? Why do we need a separate RPC for the teleport?
Why is the action client RPC not just action anticipation?

SamuelBellomo: So in other words, why two RPC calls for one action?

Assets/BossRoom/Scripts/Client/Game/Action/ActionVisualization.cs
```diff
@@ -66,6 +66,14 @@ public void OnStoppedChargingUp(float finalChargeUpPercentage)
             }
         }
 
+        public void OnTeleport()
```
SamuelBellomo, line 69: Adding one single action has a lot of rippling effect. Why is the generic ActionVisualization aware of teleportation? Same for the other changes in this PR, why does it change so many files?

Assets/BossRoom/Scripts/Client/Game/Character/ClientCharacterVisualization.cs
```diff
@@ -339,5 +361,27 @@ public bool IsAnimating
                 return false;
             }
         }
+
+        /// <summary>
+        /// Turns off the normal auto-movement mode, allowing ActionFXes to manually move
+        /// the visualization around. To return to normal mode, call <see cref="StartFollowingNetworkTransform"/>().
+        /// </summary>
+        public void StopFollowingNetworkTransform()
```
SamuelBellomo, line 369: Is this still needed?

Assets/BossRoom/Scripts/Server/Game/Action/DashAttackAction.cs
```diff
@@ -0,0 +1,85 @@
+using MLAPI;
```
SamuelBellomo, line 1: As a user, the first thing I did was press the dash without setting a target. I saw my character do some vfx, but nothing else. Simple screen message or setting a default target destination would help

Assets/BossRoom/Scripts/Client/Game/Action/DashAttackActionFX.cs
eheimburg: What's happening there is that the action is canceled due to movement. I've cleaned up the action-cancel code to explicitly cancel the FX

Assets/BossRoom/Scripts/Server/Game/Action/DashAttackAction.cs
eheimburg: I've give the action a ReuseTime of 2 seconds (which can be higher if desired).

When testing this I found it to be very unfortunate that canceling the action doesn't reset the reuse time. (Meaning that even if you cancel the teleport action, you still have to wait 2 seconds before you can start again.) I've fixed `ActionPlayer` to clear the timestamp so that the action can be started again immediately if desired.

Assets/BossRoom/Scripts/Client/Game/Character/ClientCharacterVisualization.cs
eheimburg: (This is new code, I added it for this version of the action.) Since the visualization normally follows the networked transform's position, I wanted an explicit way to cleanly stop it from doing so while being driven by an ActionFX.

In this case it's not _technically_ needed because of the long `ExecTimeSeconds` on this action: after 1.3 seconds, all smooth-movement will have finished, so `ClientCharacterVisualization` won't happen to be trying to move the character at the same time the Action is moving it.

I still like having this here as a way to make the behavior contract explicit, but can remove if desired.

Assets/BossRoom/Scripts/Server/Game/Character/ServerCharacterMovement.cs
eheimburg: (Sorry, I didn't realize this comment had been marked as Resolved before I started working on it -- but I think it's worth addressing anyway)

Since the point of this RPC was the teleport, I started with a clean version of teleportation and then made an Action that shows how to teleport with a unique visualization. This way it's obvious how to add something like a teleport platform - just call `ServerCharacterMovement.Teleport()`. 

But it's true the visualization doesn't need the extra RPC. I've changed `ServerCharacterMovement.Teleport()` to have a second parameter indicating whether the RPC is needed, and set it to false here.

I've also removed the code that passed the teleportation event to the ActionFX as it isn't necessary.

Assets/BossRoom/Scripts/Client/Game/Action/ActionVisualization.cs
eheimburg: (see earlier comment)

removed `OnTeleport()` from `ActionVisualization`, `ActionFX`, and `DashAttackActionFX`

Assets/BossRoom/Scripts/Server/Game/Action/DashAttackAction.cs
eheimburg: > Also, is the action abiding by the range or is that set to a fairly high number? I'm able to zoom out and pick the farthest point on the map and traverse there.

The action's range was set pretty high at 20m (I've lowered it to 15m) but there was also a missing distance check if the action had no target. This has been fixed so that if it's too far away, we pick a new teleport spot that's closer. (As opposed to fixing it by having the character walk until they're in range, which is what happens if they have a target.)

eheimburg: If the teleport-destination seems unset (is precisely Vector3.zero or is extremely close to the character's current location), we now pick a new spot that's `ActionDescription.Range` away. See what you think about this implementation.

Assets/BossRoom/Scripts/Server/Game/Action/DashAttackAction.cs
eheimburg: Added check so that if the teleport-destination seems unset (is precisely Vector3.zero) or is extremely close to the character's current location, we now pick a new spot that's ActionDescription.Range away. This ensures that clicking the action at startup has a noticeable effect.

eheimburg:
I've tried to address all the PR comments, so I think this is ready for another review.

Cosmin-B (dismissed):

Cosmin-B:
Most of the feedback seems to have been addressed. @SamuelBellomo @fernando-cortez please confirm!

Assets/BossRoom/Scripts/Server/Game/Action/DashAttackAction.cs
```diff
@@ -0,0 +1,98 @@
...
+        private Vector3 m_TargetSpot;
+
+        public DashAttackAction(ServerCharacter parent, ref ActionRequestData data) : base(parent, ref data)
+        {
+            Assert.IsTrue(Description.Radius > 0, $"ActionDescription for {Description.ActionTypeEnum} needs a Radius assigned!");
+        }
+
+        public override bool Start()
+        {
+            // if we don't have a meaningufl stop-spot, try to find one that's a decent distance away
```
fernando-cortez, line 34: Spelling.

Assets/BossRoom/Scripts/Server/Game/Action/DashAttackAction.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            // tell clients to visualize this action
+            m_Parent.NetState.RecvDoActionClientRPC(Data);
+
+            return ActionConclusion.Continue;
+        }
+
+        public override bool Update()
+        {
+            return ActionConclusion.Continue;
```
fernando-cortez, line 55: This might complicate some things with the need of a secondary RPC, but one case where it's not practical to use the dash is the case when the target moves after choosing it as the destination. An example is targetting an imp who is already chasing you: this action will move towards the initial target spot, not the current imp position so the action ends behind the imp thus dealing no damage. 

Wondering if we can get around this by applying damage while we dash?

Assets/BossRoom/Scripts/Server/Game/Action/DashAttackAction.cs
fernando-cortez: Much better than before. Maybe it feels a little too far when stationary but I don't object to leaving it as it is.

Cosmin-B: Yes, i like this too.

Assets/BossRoom/Scripts/Server/Game/Character/ServerCharacterMovement.cs
```diff
@@ -125,6 +125,34 @@ public void CancelMove()
...
+        /// <summary>
+        /// Instantly moves the character to a new position. NOTE: this cancels any active movement operation!
+        /// Typically this includes an RPC that will let clients know that this movement happened due to a teleport,
+        /// so that they don't try to animate the character traveling to the new location. But if the teleport occurs
+        /// as part of an Action that already handles the visualization (such as in DashAttackActionFX), this extra
+        /// RPC isn't necessary and can be omitted.
+        /// </summary>
+        /// <param name="newPosition">new coordinates the character should be at</param>
+        /// <param name="sendClientRpc">whether we should notify clients with an RPC</param>
+        public void Teleport(Vector3 newPosition, bool sendClientRpc)
```
SamuelBellomo, line 137: from what I can see, sendClientRpc is always false. YAGNI would apply here

Assets/BossRoom/Scripts/Client/Game/Character/ClientCharacterVisualization.cs
```diff
@@ -65,6 +65,13 @@ public class ClientCharacterVisualization : NetworkBehaviour
 
         event Action Destroyed;
 
+
+        /// <summary>
+        /// When true, the visualization moves itself to keep in sync with the network transform.
+        /// (This can be disabled by special actions that want to reposition the visualization manually.)
+        /// </summary>
+        bool m_IsFollowingNetworkTransform = true;
```
SamuelBellomo, line 73: Why the name "network transform"? I thought we weren't using MLAPI's NetworkTransform yet? (let's make sure we don't confuse our users and use different words than this in all of this PR)

Assets/BossRoom/Scripts/Client/Game/Character/ClientCharacterVisualization.cs
SamuelBellomo: YAGNI
This is code lying around that's not used and so not tested. This will rot whenever we change the way we're doing movements in the future.
Yes please remove

Assets/BossRoom/Scripts/Server/Game/Action/DashAttackAction.cs
SamuelBellomo: I think it's acceptable to keep it as is. It'd make it a "surprise attack" kind of dash and make it less OP (it's already pretty OP). 
I wouldn't add too much complexity to our solution just for gameplay reasons.

Assets/BossRoom/Scripts/Client/Game/Character/ClientCharacterVisualization.cs
eheimburg: (Will revise when responding to your other feedback)

Assets/BossRoom/Scripts/Server/Game/Action/DashAttackAction.cs
eheimburg: ok. actually that if-statement is redundant (it was moved into the utility function the next line down, so that both client and server can use it). removed the if-statement and comment

Assets/BossRoom/Scripts/Server/Game/Character/ServerCharacterMovement.cs
eheimburg: ok

Assets/BossRoom/Scripts/Client/Game/Character/ClientCharacterVisualization.cs
eheimburg: this variable and all related code was removed (by request of another comment) so issue is moot!

Assets/BossRoom/Scripts/Client/Game/Character/ClientCharacterVisualization.cs
eheimburg: ok

Cosmin-B (dismissed):
An interesting clean-up, I like it.

Assets/BossRoom/Scripts/Client/Game/Action/DashAttackActionFX.cs
```diff
@@ -9,6 +9,13 @@ namespace BossRoom.Visual
     /// But on the client, we visualize this as the character dashing across the screen. The dashing begins after
     /// ExecTimeSeconds have elapsed.
     /// </summary>
+    /// <remarks>
+    /// This script moves the character visualization during the "dash", which is normally performed by
+    /// <see cref="ClientCharacterVisualization"/>. If both scripts try to move the visualization at the
+    /// same time it would result in unpredictable movement stuttering, but it's fine in this case because
+    /// the character has been stationary for ExecTimeSeconds before we start moving it (so the regular
+    /// visualization code will be dormant.)
```
fernando-cortez, line 17: These last two lines leave me a little confused. Shouldn't there be a clash? `ClientCharacterVisualization` should be applying SmoothMove towards the ServerCharacter position on every Update() right? So is it really dormant?

Assets/BossRoom/Scripts/Client/Game/Action/DashAttackActionFX.cs
```diff
@@ -96,10 +100,6 @@ public override void End()
             {
                 m_Parent.OurAnimator.SetTrigger(Description.Anim2);
             }
-            if (m_StartedDash)
-            {
-                m_Parent.StartFollowingNetworkTransform();
-            }
         }
 
         public override void Cancel()
```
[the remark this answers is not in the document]
eheimburg: There's a .3 second window where that can happen. Basically, we render you zipping across the screen before we teleport you... and if you cancel the action during that tiny time frame, the action is canceled, so we have no recourse except to put your character back where it was.

Assets/BossRoom/Scripts/Client/Game/Action/DashAttackActionFX.cs
eheimburg: Hmm, I think you're right -- the `ClientCharacterVisualization` will continue to pull the visualization back toward the server character position. But the movement isn't strong enough to be noticeable due to the very short time we spend moving the visualization.

Should I add the visualization-disabling boolean back? Or just adjust the comment?

Assets/BossRoom/Scripts/Client/Game/Action/DashAttackActionFX.cs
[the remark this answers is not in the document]
fernando-cortez: Yes, I was just going to comment that the snap I think is okay. The lasting animation should be cleared.

Assets/BossRoom/Scripts/Client/Game/Action/DashAttackActionFX.cs
fernando-cortez: That's a good question. Since the movement workaround removal was suggested by @SamuelBellomo, let's get his opinion here. It doesn't feel right to have two concurrent movement behaviours, however.

SamuelBellomo: SmoothMove is called before ActionFX.Update() in ClientCharacterVisualisation
It's normal this lerp here overrides the interpolation. (let's make sure we update our comment to reflect this)

SamuelBellomo, line 17:
```suggestion
    /// This script moves the character visualization during the "dash", which is normally performed by
    /// <see cref="ClientCharacterVisualization"/>. Since SmoothMove is called before ActionFX.Update, action fx position changes override interpolated positions
```

SamuelBellomo: I added a suggestion change. Once we take this in, I'm good

SamuelBellomo:
I'll let Fernando resolve his conversations, I'm good on my side once my suggestion has been commited

Assets/BossRoom/Scripts/Client/Game/Action/DashAttackActionFX.cs
eheimburg: works for me

ok

Assets/BossRoom/Scripts/Client/Game/Action/DashAttackActionFX.cs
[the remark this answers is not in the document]
eheimburg: This is caused by Anticipation, which makes it possible to run two copies of the ActionFX at the same time. Since the action resets its end-triggers, that second instance of the action can screw up the first. I've removed the explicit trigger-reset, which fixes the problem. (I liked explicitly resetting the triggers as antibugging, but it obviously won't work here. I think Anticipation might be changed to not run multiple stacks of the same Action, but that's outside the scope of this PR.)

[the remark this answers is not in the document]
fernando-cortez: I see. This works for now but I'll follow up on a possible follow-up to this PR.

Cosmin-B (approved):

fernando-cortez (approved):

PR #285, opened 2021-04-15
Feature/Action Bar Improvements

eheimburg:
Improves the action bar in two ways:
- adds tooltips to action buttons
- when another player is selected, the basic-attack button becomes a Revive button (so that its behavior matches what happens if you right-click another player)

*Notes:*
- The "Revive" icon isn't in the repository yet, so right now we're using an unused attack icon. This will be replaced with the correct icon in a later PR
- I made the BossRoom scene's canvas into a new prefab, `BossRoomHudCanvas`. All changes to the scene are encapsulated in this prefab (e.g. the tooltip configuration stuff). This way we won't need to change the scene to edit the hud in the future


Assets/BossRoom/Scripts/Client/UI/HeroActionBar.cs
```diff
@@ -128,10 +113,62 @@ void OnButtonClickedUp(int buttonIndex)
...
+
+            if (m_NetState.TargetId.Value != 0
+                && NetworkSpawnManager.SpawnedObjects.TryGetValue(m_NetState.TargetId.Value, out NetworkObject selection)
+                && selection != null
+                && selection.IsPlayerObject
+                && selection.NetworkObjectId != m_NetState.NetworkObjectId)
+            {
+                m_IsOtherPlayerSelected = true;
+                // we have a different player selected! In that case we want to reflect that our basic Action is a Revive, not an attack
+                UpdateIcon(0, ActionType.GeneralRevive);
```
fernando-cortez, line 136: This could be added now or in a new PR, but I find that we should have a disabled sprite for revival when the target is a player that is still alive. The button should only turn into the active revive action when they're truly actually fainted.

Assets/BossRoom/Prefabs/BossRoomHudCanvas.prefab.meta
```diff
@@ -0,0 +1,7 @@
+fileFormatVersion: 2
```
fernando-cortez, line 1: For this "HowToPlayPopupPanel" prefab I like that the popup panel prefab reference was broken but the PopupPanel script still remains. That can safely go.

Assets/BossRoom/Scripts/Client/UI/HeroActionBar.cs
SamuelBellomo: let's do this in this PR

Assets/BossRoom/Prefabs/BossRoomHudCanvas.prefab.meta
eheimburg: ok

Assets/BossRoom/Scripts/Client/UI/HeroActionBar.cs
```diff
@@ -43,52 +63,17 @@ void OnDisable()
             }
         }
 
-        public void RegisterInputSender(Client.ClientInputSender inputSender)
+        void OnDestroy()
```
Cosmin-B, line 66: Question: You have allocated some memory on OnEnable for m_Tooltips, should we release those back to the system, by explicitly setting `m_Tooltips = null` in an OnDisable function?

Assets/BossRoom/Scripts/Client/UI/HeroActionBar.cs
```diff
@@ -106,9 +91,9 @@ void OnButtonClickedDown(int buttonIndex)
 
             switch (buttonIndex)
             {
-                case 0: m_InputSender.RequestAction(m_CharacterData.Skill1, SkillTriggerStyle.UI); break;
-                case 1: m_InputSender.RequestAction(m_CharacterData.Skill2, SkillTriggerStyle.UI); break;
-                case 2: m_InputSender.RequestAction(m_CharacterData.Skill3, SkillTriggerStyle.UI); break;
+                case 0: m_InputSender.RequestAction(m_IsOtherPlayerSelected ? ActionType.GeneralRevive : m_NetState.CharacterData.Skill1, SkillTriggerStyle.UI); break;
```
Cosmin-B, line 94: Question: How could a new developer that takes a look at this, will be able to understand what those magic index values represent for Boss Room?

Cosmin-B (changes requested):

Assets/BossRoom/Scripts/Client/UI/HeroActionBar.cs
eheimburg: ok

Assets/BossRoom/Scripts/Client/UI/HeroActionBar.cs
eheimburg: I didn't write this code originally, but I found it fairy obvious how things were set up. If you're recommending we create an enum for the buttons instead of using an array, we can definitely do that! (But I don't think it would clarify the code for me.)

Attempted some simple clarifications:
- since the button array is hard-coded to support 4 buttons exactly, I added commentary saying so, and an Assert to confirm it at startup
- Rectored tertiary operator in case 0 switches to use a separate boolean to more explicitly convey intent

Assets/BossRoom/Scripts/Client/UI/HeroActionBar.cs
eheimburg: I've gotten two slightly different instructions on this one: the notes here are to disable the "revive" button when the selected player isn't revivable, but out of band, it sounds like maybe I'm supposed to "hide" the revive button, presumably showing the basic-attack icon if the selected character is alive. I've implemented the "disable" approach, but can easily switch it just showing a non-disabled basic-attack icon instead.

Note that as part of this change we need to check the selected player's NetState during `Update()` to see if their aliveness state has changed. To implement this efficiently we cache the selected player's NetState.

ok (but please ensure this is the intended behavior)

fernando-cortez: I don't object to this implementation. A disabled hit action might be the more "correct" implementation, but I do see the argument for having it like this. It subtly shows the user that you "can" revive another player, just not now. I'll let others weigh in.

Assets/BossRoom/Scripts/Client/UI/HeroActionBar.cs
fernando-cortez: I can't speak for @Cosmin-B, but I think he meant the indices used in the switch statement, which come from the UIHUDButton class. Would defining each Button out explicitly (instead of having an array) help in readability?

Cosmin-B: I think they will, Fernando.

eheimburg: ok, assigned constants for all hard-coded values in the class.

eheimburg: oh, I should mention that I tried implementing it with the arrays completely removed, but it was ugly for two reasons:

- we need to keep track of each button's cached ref to their `UITooltipDetector` component. That's currently done by just having a second array with identical indices. Without the arrays, we need a different way to associate the two fields (e.g. a struct for each button+tooltip combo, or a switch-statement that returns the tooltip for a specific button). That's fine but it involves lots of ugly setup/teardown code.
- the reason those switch statements exist is because all buttons call the same callback events with different IDs. Either we need separate callback events for each button (eight total: four button-down and four button-up), or we have the same situation as we have now, where we need constants. 

So ultimately I think just using constants are the simplest solution here

Cosmin-B: I do like this idea, a lot more, although the current implementation with constants, also looks way better than before.
`if this still seems bad, I'd next try an enum approach: using a private enum to index an array of structs, each struct containing the button, the tooltip, and any other associated data. (That's how I would probably have written this class if I'd made it originally.) That gives each button its own serialized variable, and hides the array internally. More code overall, but let me know if that is preferable.`

Cosmin-B: Would you be able to do this now?

FPaizSF:
Please let us know if other code changes are needed for this. Otherwise we will fix merge conflicts soon so it can be merged.

Assets/BossRoom/Scripts/Client/UI/HeroActionBar.cs
eheimburg: Here is that version. As you can see it's a little bit more code. But one of the benefits is that we can remove the `ButtonID` int from `UIHUDButton`, which I think is a win.

eheimburg:
Made requested changes to `HeroActionButton` (and associated minor changes to `UIHUDButton`) which need to be re-reviewed. Other than that, I think we are ready

Assets/BossRoom/Scripts/Client/UI/HeroActionBar.cs
```diff
@@ -147,18 +208,18 @@ void OnButtonClickedDown(int buttonID)
...
             }
 
-            switch (buttonID)
+            switch (buttonType)
             {
-                case k_BasicActionButtonID: m_InputSender.RequestAction(basicAction, SkillTriggerStyle.UI); break;
-                case k_SpecialAction1ButtonID: m_InputSender.RequestAction(m_NetState.CharacterData.Skill2, SkillTriggerStyle.UI); break;
-                case k_SpecialAction2ButtonID: m_InputSender.RequestAction(m_NetState.CharacterData.Skill3, SkillTriggerStyle.UI); break;
-                default: throw new System.Exception($"Unknown button ID {buttonID}");
+                case ActionButtonType.BasicAction: m_InputSender.RequestAction(basicAction, SkillTriggerStyle.UI); break;
```
Cosmin-B, line 213: This implementation does look good. But I was expecting that these switch cases would vanish.

eheimburg: Oh, right. Here is that version. 

Since I added a non-readonly variable to the internal struct, I turned it into a class. This is just because it's less surprising to use. (We pass it as a function parameter, and this way the function can just change the variable without having to re-assign it to the dictionary, or pass with the ref keyword.) I think this is fine because they are allocated during `Awake()`, and are very tiny.

Cosmin-B (approved):
Excellent.

fernando-cortez (approved):

PR #291, opened 2021-04-19
Bugfix: testing commands created monsters that were not destroyed on new scene

eheimburg:
Change the testing commands to explicitly spawn test entities with the "don't destroy on load" parameter set to true.

(This fixes null-ref exceptions when you have a test-spawned monster selected when the game switches to the game over scene.)

### DIFF
--- a/Assets/BossRoom/Scripts/Server/ServerTestingHotkeys.cs
+++ b/Assets/BossRoom/Scripts/Server/ServerTestingHotkeys.cs
@@ -50,12 +50,12 @@ namespace BossRoom.Server
             if (m_SpawnEnemyKeyCode != KeyCode.None && Input.GetKeyDown(m_SpawnEnemyKeyCode))
             {
                 var newEnemy = Instantiate(m_EnemyPrefab);
-                newEnemy.SpawnWithOwnership(NetworkManager.Singleton.LocalClientId);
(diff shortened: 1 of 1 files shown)


Cosmin-B (dismissed):
Looks good to me!

Assets/BossRoom/Scripts/Server/ServerTestingHotkeys.cs
```diff
@@ -50,12 +50,12 @@ private void Update()
             if (m_SpawnEnemyKeyCode != KeyCode.None && Input.GetKeyDown(m_SpawnEnemyKeyCode))
             {
                 var newEnemy = Instantiate(m_EnemyPrefab);
-                newEnemy.SpawnWithOwnership(NetworkManager.Singleton.LocalClientId);
```
fernando-cortez, line 53: Not a comment on this PR, but it would resolve an oversight on the previous PR. As part of guidelines, we are not allowed to be using #if UNITY_EDITOR preprocessor directives in the project. And as far as testing goes, it's okay for users to keep this in their release builds so let's eliminate line 14 & 66. Thanks!

eheimburg: ok

Cosmin-B (approved):

fernando-cortez (approved):

PR #293, opened 2021-04-20
Bugfix/broken pieces should not be hit by attacks/projectiles [GOMPS-461]

eheimburg:
Adds a new Layer to the project, "Debris". The "broken" versions of pots, pillars, and crystals are now in the Debris layer. Since projectiles and attacks check for the "PC" "NPC" and "Ground" layers, this fixes the issue. 

(Previously these "broken versions" were using the NPCs layer.)


Cosmin-B (approved):
Looks okay to me!

fernando-cortez (approved):

PR #294, opened 2021-04-20
Bugfix: Fix mage-bolt friendly fire [GOMPS-299]

eheimburg:
Many actions can cause the player to "look like" they're hitting another player (e.g. selecting an ally and clicking the Attack button will make your ally play a hit-react animation) but that's just amusing fluff... the problem is that mage bolt did actual damage to other players. This is fixed.

Also fixed: if the target of the mage bolt had disappeared since the attack was queued, it was simply aborted because it had no target. Now, the attack is not aborted, and turns into a "missed" bolt. (To do this, the client sends the click-position along with all mage-bolt attacks.)

### DIFF
--- a/Assets/BossRoom/Scripts/Client/Game/Action/FXProjectileTargetedActionFX.cs
+++ b/Assets/BossRoom/Scripts/Client/Game/Action/FXProjectileTargetedActionFX.cs
@@ -102,10 +102,16 @@ namespace BossRoom.Visual
                 return null;
             }
 
-            NetworkObject obj;
-            if (NetworkSpawnManager.SpawnedObjects.TryGetValue(Data.TargetIds[0], out obj) && obj != null)
+            if (NetworkSpawnManager.SpawnedObjects.TryGetValue(Data.TargetIds[0], out NetworkObject targetObject) && targetObject != null)
             {
-                return obj;
+                // make sure this isn't a friend (or if it is, make sure this is a friendly-fire action)
+                var targetable = targetObject.GetComponent<ITargetable>();
+                if (targetable != null && targetable.IsNpc == (Description.IsFriendly ^ IsParentAnNPC()))
+                {
+                    // not a valid target
+                    return null;
+                }
+                return targetObject;
             }
             else
             {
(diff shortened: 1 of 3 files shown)


Cosmin-B (dismissed):
Personally, I don't see any red flags with this fix, so, it looks good to me.

Assets/BossRoom/Scripts/Client/Game/Action/FXProjectileTargetedActionFX.cs
```diff
@@ -100,14 +95,6 @@ private void PlayHitReact()
...
 
-        /// <summary>
-        /// Do we even have a target? (If false, it means player clicked on nothing, and we're rendering a "missed" fake bolt.)
-        /// </summary>
-        private bool HasTarget()
-        {
-            return Data.TargetIds != null && Data.TargetIds.Length > 0;
-        }
-
         private NetworkObject GetTarget()
```
fernando-cortez, line 98: Shouldn't this match the Action's implementation? As in, if these two are matching, we won't see the target react, correct? I feel that it's weird that on the End() for the FX, we have the target trigger their hit react animation when it's a "missed" bolt.

eheimburg: > As in, if these two are matching, we won't see the target react, correct?

Sorry, I can't figure out which two things you're referring to?

(If we have a target, it's never a missed bolt, so I was trying to simplify things.)

fernando-cortez: Right. So this PR fixes the friendly fire damage, but it's still firing the hit react animation on an ally, which is what I feel can be removed. What I suggested was to align both the Action and FX GetTarget() methods so that the FX is also verifying that it's not an ally, so that you can't possibly trigger a hit react animation on your target (since target would be null on an ally).

eheimburg: Oh, gotcha. Revised to address this

Cosmin-B (approved):
The changes looks good to me and the feedback has been addressed!

fernando-cortez (approved):

PR #295, opened 2021-04-20
Feature/Spawner Improvements

eheimburg:
Adds new settings to the ServerWaveSpawner. These new settings will be tuned and tweaked in the final "balance pass". This is just the code pass, so the values I entered are placeholders. 

(The ultimate goal is to make the spawners relevant to the demo! Right now you can simply ignore the spawners and attack the boss, because they're off in the corners and the imps don't aggro. This is also one of the places we can dynamically tune the game difficulty based on number of players, so we want to demonstrate that.)

*Feature Additions*
- The max # of simultaneous spawns is now determined based on the number of players
- Spawner can now override the spawned entity's detection range
- New option makes spawner's player-detector skip over players who are Stealthy

*Other*
- Added more spawn-positions to the prefab for a total of 10 (to allow a max of 10 spawned imps without any of them overlapping each other when they are all at idle)
- Removed the "unlimited spawns" option because it's dangerous (it could easily overwhelm the Host with infinite imps) and it's unlikely to be what is actually desired. (But if designers do want effectively unlimited spawns, they can just set the cap to a million!)

### DIFF
--- a/Assets/BossRoom/Scripts/Server/Game/Character/ServerCharacter.cs
+++ b/Assets/BossRoom/Scripts/Server/Game/Character/ServerCharacter.cs
@@ -42,6 +42,12 @@ namespace BossRoom.Server
         // Cached component reference
         private ServerCharacterMovement m_Movement;
 
+        /// <summary>
+        /// If we are created by a spawner, the spawner might override our detection radius
+        /// -1 is a sentinel value meaning "no override"
+        /// </summary>
+        private float m_DetectRadiusOverride = -1;
+
         private void Awake()
         {
             m_Movement = GetComponent<ServerCharacterMovement>();
@@ -212,6 +218,33 @@ namespace BossRoom.Server
         public float GetBuffedValue(Action.BuffableValue buffType)
         {
             return m_ActionPlayer.GetBuffedValue(buffType);
+        }
+
+        /// <summary>
+        /// Returns the range at which this character can detect enemies. Only applicable to NPCs.
+        /// This is usually the same value as is indicated by our game data, but it can be
+        /// dynamically overridden.
+        /// </summary>
+        /// <returns>our entity detection range (in meters)</returns>
+        public float GetDetectRadius()
+        {
+            if (m_DetectRadiusOverride == -1)
+            {
+                return GameDataSource.Instance.CharacterDataByType[NetState.CharacterType].DetectRange;
+            }
+            else
+            {
+                return m_DetectRadiusOverride;
+            }
+        }
+
+        /// <summary>
+        /// Changes this creature's detection radius. Only meaningful for NPCs.
+        /// </summary>
+        /// <param name="radius">new detection radius for this entity (in meters)</param>
+        public void SetDetectRadius(float radius)
+        {
+            m_DetectRadiusOverride = radius;
         }
 
         /// <summary>
--- a/Assets/BossRoom/Scripts/Server/Game/Character/ServerCharacter.cs
+++ b/Assets/BossRoom/Scripts/Server/Game/Character/ServerCharacter.cs
@@ -256,7 +256,7 @@ namespace BossRoom.Server
         }
 
         /// <summary>
-        /// This character's AIBrain. Will be null if this is not an NPC, or if this NPC has no "brain".
+        /// This character's AIBrain. Will be null if this is not an NPC.
         /// </summary>
         public AIBrain AIBrain { get { return m_AIBrain; } }
     }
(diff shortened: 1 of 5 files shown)


Assets/BossRoom/Scripts/Server/Game/Entity/ServerWaveSpawner.cs
```diff
@@ -63,8 +67,19 @@ public class ServerWaveSpawner : NetworkBehaviour
...
         [SerializeField]
-        [Tooltip("The spawner won't create more than this many entities at a time. 0 = don't track spawn count")]
-        int m_MaxActiveSpawns;
+        [Tooltip("When looking for players within proximity distance, should we count players in stealth mode?")]
+        bool m_DetectStealthyPlayers;
+
+        [Header("Spawn Cap (i.e. number of simultaneously spawned entities)")]
+        [SerializeField]
+        [Tooltip("The minimum number of entities this spawner will try to maintain (regardless of player count)")]
+        int m_MinSpawnCap;
```
Cosmin-B, line 76: Could we initialize these values to some default ones in code, please?

Assets/BossRoom/Scripts/Server/Game/Entity/ServerWaveSpawner.cs
```diff
@@ -217,20 +229,45 @@ NetworkObject SpawnPrefab()
...
+        }
+
+        /// <summary>
+        /// Returns the current max number of entities we should try to maintain.
+        /// This can change based on the current number of living players; if the cap goes below
+        /// our current number of active spawns, we don't spawn anything new until we're below the cap.
+        /// </summary>
+        int GetCurrentSpawnCap()
+        {
+            int numPlayers = 0;
```
Cosmin-B, line 261: Just to confirm, we are not despawning the surplus of Active Spawns, when the player leaves the server/host session?

Cosmin-B (changes requested):

Cosmin-B (dismissed):
Looks good to me!

Assets/BossRoom/Scripts/Server/Game/Entity/ServerWaveSpawner.cs
eheimburg: - Assigned defaults (based on the settings in the spawner prefab, but with a few modifications)
- Moved some non-serialized variables down below the massive list of serialized variables (just for clarity)

ok

Assets/BossRoom/Scripts/Server/Game/Entity/ServerWaveSpawner.cs
eheimburg: Correct, my understanding is the monsters shouldn't de-spawn, we just don't create more while we're over the cap.

Cosmin-B: Thanks for confirming!

Assets/BossRoom/Scripts/Server/Game/Character/ServerCharacter.cs
```diff
@@ -214,6 +220,33 @@ public float GetBuffedValue(Action.BuffableValue buffType)
...
+
+        /// <summary>
+        /// Changes this creature's detection radius. Only meaningful for NPCs.
+        /// </summary>
+        /// <param name="radius">new detection radius for this entity (in meters)</param>
+        public void SetDetectRadius(float radius)
+        {
+            m_DetectRadiusOverride = radius;
+        }
+
```
fernando-cortez, line 249: Could this be placed someplace else that's pertinent only to NPCs that have detecting capabilities? This ServerCharacter component shouldn't be aware of this, I feel.

eheimburg: I think it belongs in `AIBrain`, but because that's a private object owned by `ServerCharacter` (and not a component) I felt weird exposing it. Here's what that looks like with `ServerCharacter` providing an `AIBrain` accessor, and `AIBrain` having the radius.

I think we had discussed making AIBrain its own component eventually, and that's an existing jira task, so I'm fine with this approach given that it will likely become a component in the future

Cosmin-B: i'm okay with that for now!

Cosmin-B (approved):

Assets/BossRoom/Scripts/Server/Game/Character/ServerCharacter.cs
fernando-cortez: This works great for now.

fernando-cortez (approved):

PR #296, opened 2021-04-20
Feature/Show full particle on imp death

eheimburg:
Imp monsters' GameObjects are being destroyed before their full "poof" particle can play. (This results in the particle looking like a "splat" rather than cool purple smoke as intended.)

To address this, I added a new boolean that lets a particle be de-parented from the character. This option is enabled on the imp's death particle. I double-checked that the particle's `SpecialFXGraphics` has a timeout, so that it will self-destruct itself.

(At the moment, This is the only planned use of this new bool.)

Merging note:: the only actual change to `ImpGraphics.prefab` was to enable the new bool. All other changes shown here were automatic


Cosmin-B (approved):
Looks good to me!

fernando-cortez (approved):

PR #302, opened 2021-04-21
fixing NRE in RoomNameBox

dwoodruffsf:
We didn't have a repro for this, but I'm sure it's some odd boundary condition either on starting or stopping. We also didn't have exact knowledge of which object was null, but the line number showed it happening on the if statement I modified, so I changed the statement to be very conservative. 
fixing: GOMPS-458


Cosmin-B (approved):
Looks good

fernando-cortez (approved):

PR #308, opened 2021-04-26
Feature/Shield-Charge attack has anticipation anim [GOMPS-472]

eheimburg:
The tank's "Charged Shield" action did not support Anticipation (i.e. playing a client-side action before the server tells us to, for more reactive gameplay). This is fixed by this PR.

Actions with special input-handlers didn't trigger the same client-side events that other actions did. To ensure that all actions sent to the server also send appropriate events to the client, I've routed a new callback into `BaseActionInput`, which derived versions can use to send input correctly. 

After this is done, the actual custom Anticipation logic for `ChargedShieldActionFX` is trivial, and just like other Actions.

Also fixed issue where you could sometimes get the animations out of sync if you spam-clicked the shield charge button. (By resetting the end-animation trigger before we begin.)

### DIFF
--- a/Assets/BossRoom/Scripts/Client/Game/Action/AoeActionInput.cs
+++ b/Assets/BossRoom/Scripts/Client/Game/Action/AoeActionInput.cs
@@ -59,7 +59,7 @@ namespace BossRoom.Visual
                         ShouldQueue = false,
                         TargetIds = null
                     };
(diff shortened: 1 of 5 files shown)


Cosmin-B (approved):

Cosmin-B:
Looks good to me, but let's get some more eyes on this @fernando-cortez

fernando-cortez (approved):

PR #309, opened 2021-04-26
Feature/remove cooldown time variable

eheimburg:
The `CooldownTimeSeconds` variable of `ActionDescription` was obsoleted by `ReuseTimeSeconds`, and this PR removes the old variable.

Only one action was still using the `CooldownTimeSeconds` feature: the boss's trample attack. To fix this, the AI is changed to instead pick actions whose `ReuseTimeSeconds` are up.

There was a subtle side-effect to the boss's use of cooldown time: the boss would stand still for a few seconds after the trample ended (waiting for the cooldown to run out). This made the boss more vulnerable than was probably intended -- he should only stand still while stunned! But it turns out our animator controller relied on this behavior: the animation loop only stopped when the boss's movement became zero. Since the boss doesn't *necessarily* stop moving anymore, we now just use a trigger to end the animation sequence. (Which in retrospect is probably how it should have been done in the first place.)

*Notes*
- to let the boss use his trample as often as before, I've given `BossImpTrampleAttack` a `ReuseTimeSeconds` of 7.5. (The `DurationSeconds` is 3.5, and the `CooldownTimeSeconds` was previously 4, so a reuse time of 3.5+4=7.5.) 
- there was some weird timing calculation going on in the old TrampleAction.cs which doesn't look like it was accurate -- but it happened to work. It's all removed now though, so problem solved
- Merging note: added one trigger, and changed one transition to use the trigger
- Testing note: remember that you can spawn a boss imp at the dungeon entrance by pressing "B"

### DIFF
--- a/Assets/BossRoom/Scripts/Server/Game/AIState/AttackAIState.cs
+++ b/Assets/BossRoom/Scripts/Server/Game/AIState/AttackAIState.cs
@@ -12,6 +12,11 @@ namespace BossRoom.Server
 
         List<ActionType> m_AttackActions;
 
+        /// <summary>
+        /// a temporary list used by GetCurrentAttackInfo() so that it doesn't have to allocate a list every call
+        /// </summary>
+        List<ActionType> m_TempListOfAttackActions = new List<ActionType>();
+
         public AttackAIState(AIBrain brain, ActionPlayer actionPlayer)
         {
             m_Brain = brain;
@@ -88,8 +93,15 @@ namespace BossRoom.Server
                 }
             }
 
-            // Choose whether we can attack our foe directly, or if we need to get closer first
+            // choose the attack to use
             var attackInfo = GetCurrentAttackInfo();
+            if (attackInfo == null)
+            {
+                // no actions are usable right now
+                return;
+            }
+
+            // attack!
             var attackData = new ActionRequestData
             {
                 ActionTypeEnum = attackInfo.ActionTypeEnum,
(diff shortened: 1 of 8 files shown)


Assets/BossRoom/Scripts/Server/Game/AIState/AttackAIState.cs
```diff
@@ -122,9 +134,27 @@ private ServerCharacter ChooseFoe()
...
         }
 
+        /// <summary>
+        /// Chooses a usable attack and returns an ActionDescription representing it. If no actions are usable, returns null.
+        /// </summary>
+        /// <returns>Action to use, or null</returns>
         private ActionDescription GetCurrentAttackInfo()
         {
-            m_CurAttackAction = m_AttackActions[Random.Range(0, m_AttackActions.Count)];
+            m_TempListOfAttackActions.Clear();
```
Cosmin-B, line 143: The only thing that I don't like here is the fact that Getter's (in my opinion) should not modify the members of a class, it just violates multiple principles. I propose to rename this function.

Cosmin-B (changes requested):

Assets/BossRoom/Scripts/Server/Game/AIState/AttackAIState.cs
eheimburg: Renamed `GetCurrentAttackInfo()` to `ChooseAttack()` and cleaned up some leftover code. (The function can now just return an enum value instead of needing to set a member variable.)

Cosmin-B (approved):

Assets/BossRoom/Scripts/Server/Game/AIState/AttackAIState.cs
```diff
@@ -122,18 +134,27 @@ private ServerCharacter ChooseFoe()
...
+        /// <returns>Action to attack with, or ActionType.None</returns>
+        private ActionType ChooseAttack()
         {
-            m_CurAttackAction = m_AttackActions[Random.Range(0, m_AttackActions.Count)];
+            m_TempListOfAttackActions.Clear();
+            foreach (var actionType in m_AttackActions)
+            {
+                if (m_ActionPlayer.IsReuseTimeElapsed(actionType))
+                {
+                    m_TempListOfAttackActions.Add(actionType);
```
fernando-cortez, line 148: To reduce the data complexity of this class, we can remove m_TempListOfAttackActions. We could generate a random index from 0 to (m_AttackActions.Length - 1) and iterate through `m_AttackActions` , checking for `IsReuseTimeElapsed(..)` and return that Action if true, correct? If we have a random start index, we can just iterate through `m_AttackActions.Length` times and wrap around if necessary. If no Actions were found during that iteration, we would return `ActionType.None`.

fernando-cortez (commented):
Loving the cleanup.

Assets/BossRoom/Scripts/Server/Game/AIState/AttackAIState.cs
eheimburg: ok

fernando-cortez (approved):

Cosmin-B (approved):

PR #311, opened 2021-04-27
Feature/Anticipation logic tweaks [GOMPS-505]

eheimburg:
- if an anticipated action's `Start()` returns false, we should remove the action
- do not call `Update()` on an anticipated action (until it stops being anticipated)

This addresses [GOMPS-505]: Archer's Volley generates "This should not execute" exceptions

### DIFF
--- a/Assets/BossRoom/Scripts/Client/Game/Action/ActionVisualization.cs
+++ b/Assets/BossRoom/Scripts/Client/Game/Action/ActionVisualization.cs
@@ -30,7 +30,7 @@ namespace BossRoom.Visual
             for (int i = m_PlayingActions.Count - 1; i >= 0; --i)
             {
                 var action = m_PlayingActions[i];
-                bool keepGoing = action.Update();
+                bool keepGoing = action.Anticipated || action.Update(); // only call Update() on actions that are past anticipation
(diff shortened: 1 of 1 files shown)


eheimburg:
Want to get a quick check from @dwoodruffsf on this one, specifically on the notion that we don't call `Update()` on anticipated actions. That seems fine for our existing ActionFX scripts, and makes the code-flow much less weird (because otherwise we are calling `Update()` before `Start()`. 

But it's possible that future actions will need to be able to do custom anticipation logic during `Update()`. I'm considering that a "cross that bridge when you get there" problem, but maybe there's a cleverer solution that I'm missing

Cosmin-B (approved):
To me, it looks good.  I don't see a problem yet for not calling Update on an anticipated function.

dwoodruffsf:
Well, I thought about it, and I agree it is potentially confusing to get an Update before Start (particularly from the perspective of a Unity programmer, who wouldn't expect to get an Update between Awake and Start). It does feel like losing a bit of functionality, but I suppose you could start a coroutine if you really needed an update during the anticipation phase. 

I'm OK with this change.

dwoodruffsf (approved):

fernando-cortez (approved):

PR #314, opened 2021-04-28
Misc particle/sound hookups and related visual tweaks

eheimburg:
This PR is a last push to get all the particles and sounds wired up. Also did some data cleanup, removing redundant audio files from the tree and replacing some sounds. 

Also replaced the existing wav files with higher bit-rate versions. (That part was supposed to happen in an earlier PR, but there were some data mishaps.)

*There are some minor code changes:*
- Made it possible for a MeleeAction to play a particle on the target. (We use this to play a heal particle on the Mage's heal target.)
- Altered an ActionFX utility function in order to support playing particles on other entities (this is why other ActionFX besides MeleeAction were changed: they just pass another parameter to a function now)
- Removed the artificial restriction that only one particle could play per animation node. (It was a safety check because we didn't need that functionality until now.)
- Removed an unused feature of `AnimatorTriggeredSpecialFX`, the on-node-exit hook. We never ended up needing it and the code is easier to follow without it

*Animation changes:*
- Archer now plays correct animation when performing the AoE attack

*Particle changes:*
- added three particles to mage heal: the staff glows, particles play on caster, and particles play on the (anticipated) target
- added particles to boss's charge-attack sequence
- added hum sound to tank's shield particle
- torches have sound, dungeon has dripping noises as well
- Imp death plays out (with imp scream)
- Arrow volley has new sound that plays on victims (the sound is part of the arrows-hit-the-target particle)
- new wav files for all sounds
- removed the "Audio" folder which was redundant with the "Sounds" folder, made sure all sounds are coming from the Sounds folder
- revised archer powershot sound
- Boss has tears and sound effects when defeated
- volume changes for many sounds, most notably footstep sounds
- minor particle repositioning tweaks


Cosmin-B (dismissed):

Assets/BossRoom/Scripts/Client/Game/Action/MeleeActionFX.cs
```diff
@@ -27,6 +34,19 @@ public override bool Start()
             }
 
             base.Start();
+
+            // we can optionally have special particles that should play on the target. If so, add them now.
+            // (don't wait until impact, because the particles need to start sooner!)
+            if (Data.TargetIds != null && Data.TargetIds.Length > 0 && NetworkSpawnManager.SpawnedObjects.TryGetValue(Data.TargetIds[0], out var targetNetworkObj))
```
fernando-cortez, line 40: We may need a null check here on the returned targetNetworkObject right?

Assets/BossRoom/Scripts/Client/AnimationCallbacks/AnimatorTriggeredSpecialFX.cs
```diff
@@ -282,18 +213,12 @@ private void OnValidate()
...
-            }
 
             if (m_AudioSources == null || m_AudioSources.Length == 0) // if we have AudioSources handy, plug them in automatically
+            {
                 m_AudioSources = GetComponents<AudioSource>();
+            }
         }
-#endif
 
     }
```
fernando-cortez, line 224: I did my own little investigation, and there's a neat way to get around serializing properties within the editor with [SerializedProperty](https://docs.unity3d.com/ScriptReference/SerializedProperty.html).

So my suggestion is to move code within these blocks to a new Editor script that lives within Scripts/Editor/. With some tests, I also had to add BossRoom.Client to the Editor's asmdef. Anyways, I'll follow up with @dwoodruffsf offline with the suggested changes, as I got a little too deep into this and I resulted with a working prototype.

Cosmin-B: That sounds great!

Assets/BossRoom/Scripts/Client/Game/Action/MeleeActionFX.cs
eheimburg: It occurred to me that we use this gross if-statement a lot (now with four separate `&&` checks!), and we could just add a method to `ActionRequestData`, something like `TryGetFirstTarget()`. That way we could just write 

`if (Data.TryGetFirstTarget(out var targetNetObj)` 

I was tempted to add it here, but this is a PR for particles, so only minimal code-changes are appropriate... perhaps a cleanup for another time.

ok

Assets/BossRoom/Scripts/Client/AnimationCallbacks/AnimatorTriggeredSpecialFX.cs
fernando-cortez: This will go in another PR. Seems like an architectural change whereas this PR should be dedicated to visuals.

Assets/BossRoom/Scripts/Client/Game/Action/MeleeActionFX.cs
fernando-cortez: Thanks!

fernando-cortez (approved):

Cosmin-B (approved):

PR #317, opened 2021-04-29
Fix boss-death helmet jiggle by deparenting helmet [GOMPS-463]

eheimburg:
The boss's helmet appears to "jiggle" on the ground while the boss throws his temper-tantrum. This isn't a problem with the animation per se -- the helmet is stationary in the FBX file. Most of the jitter comes from animation compression, but even when turning that off, there is still a slight jiggle, which is just caused by 32-bit-float roundoff. The helmet is parented deep in the bone hierarchy, so when the boss is moving around, it's very difficult to keep the child transform perfectly still.

The best fix we could see is to de-parent the boss's helmet when it hits the ground. The death animation now generates an event at the proper moment, and a new component on the boss handles that event by disconnecting the helmet (and cleaning it up on destroy).

### DIFF
--- a/Assets/BossRoom/Scripts/Client/AnimationCallbacks/BossDeathHelmetHandler.cs
+++ b/Assets/BossRoom/Scripts/Client/AnimationCallbacks/BossDeathHelmetHandler.cs
@@ -1,0 +1,46 @@
+using UnityEngine;
+
+namespace BossRoom.Visual
+{
+    /// <summary>
+    /// De-parents the boss's helmet when he is defeated. This prevents the helmet from jittering during the
+    /// animation loop. The "boss defeat start" animation has an event that calls OnAnimEvent("HelmetLanded")
+    /// right as the helmet hits the ground. That's what we listen for here.
+    /// </summary>
+    /// <remarks>
+    /// Without this code, the boss's helmet appears to "jiggle" a bit while the boss throws his temper-tantrum.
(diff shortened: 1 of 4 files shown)


Assets/BossRoom/Scripts/Client/AnimationCallbacks/BossDeathHelmetHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// animation loop. The "boss defeat start" animation has an event that calls OnAnimEvent("HelmetLanded")
+    /// right as the helmet hits the ground. That's what we listen for here.
+    /// </summary>
+    /// <remarks>
+    /// Without this code, the boss's helmet appears to "jiggle" a bit while the boss throws his temper-tantrum.
+    /// The animation in the FBX keeps the helmet staying stationary, but it moves in-game due to animation
+    /// compression and floating-point round off. Since the helmet is parented deep in the transform hierarchy,
+    /// it's difficult to keep the helmet precisely still while all of its parent transforms are moving around wildly.
+    /// 
+    /// We could get rid of the majority of the jiggle by disabling animation-compression on the FBX, but that could
```
Cosmin-B, line 16: I appreciate that you've put this handler into it's own MonoBehavior

Cosmin-B (approved):

fernando-cortez (approved):

PR #319, opened 2021-04-30
Minor gameplay polish tasks

eheimburg:
Some easy-to-implement polish tasks that came about from feature-complete playtesting. These are intended to improve the game's first impression.

- crystals are too shiny and weird-looking when they're broken 
  - (broken bits now use a different material)
- if you're standing still and trying to melee an enemy, you should turn to face the enemy so you can actually hit them
  - note that this is overridden by movement, so if player is already charging forward when they perform a melee action, they'll still melee in the direction they're moving
- when using the archer's AoE ability, the reticule should wait for a real mouse-click, not just a mouse-release
  - (it already worked this way if you used the keyboard key to activate it, so this is just making mouse and keyboard orthogonal)
- make the 1 key fire a basic-attack (same as right-clicking the mouse), and make 2 and 3 keys perform the first and second action. (4 key continues to toggle the Emote panel on/off.)
- add numbering to the ability bar and emote bar so players know what keys do what thing

### DIFF
--- a/Assets/BossRoom/Scripts/Client/Game/Action/AoeActionInput.cs
+++ b/Assets/BossRoom/Scripts/Client/Game/Action/AoeActionInput.cs
@@ -20,6 +20,12 @@ namespace BossRoom.Visual
         Camera m_Camera;
         int m_GroundLayerMask;
         Vector3 m_Origin;
+
+        //The general action system works on MouseDown events (to support Charged Actions), but that means that if we only wait for
+        //a mouse up event internally, we will fire as part of the same UI click that started the action input (meaning the user would
+        //have to drag her mouse from the button to the firing location). Tracking a mouse-down mouse-up cycle means that a user can
+        //click on the ground separately from the mouse-click that engaged the action (which also makes the UI flow equivalent to the
+        //flow from hitting a number key). 
         bool m_ReceivedMouseDownEvent;
 
         RaycastHit[] m_UpdateResult = new RaycastHit[1];
--- a/Assets/BossRoom/Scripts/Client/Game/Action/AoeActionInput.cs
+++ b/Assets/BossRoom/Scripts/Client/Game/Action/AoeActionInput.cs
@@ -20,6 +20,7 @@ namespace BossRoom.Visual
         Camera m_Camera;
         int m_GroundLayerMask;
         Vector3 m_Origin;
+        bool m_ReceivedMouseDownEvent;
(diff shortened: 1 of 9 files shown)


fernando-cortez (approved):

Assets/BossRoom/Scripts/Client/Game/Action/AoeActionInput.cs
```diff
@@ -48,7 +49,13 @@ void Update()
...
             m_OutOfRangeVisualization.SetActive(!isInRange);
 
-            if (Input.GetMouseButtonUp(0))
+            // wait for the player to click down and then release the mouse button before actually taking the input
+            if (Input.GetMouseButtonDown(0))
+            {
+                m_ReceivedMouseDownEvent = true;
+            }
+
+            if (Input.GetMouseButtonUp(0) && m_ReceivedMouseDownEvent)
```
SamuelBellomo, line 58: Why not just MouseButtonUp? Why do you need to make sure down was pressed first?

dwoodruffsf: *context (flow of action input)*

* UIHUDButton for "Rain of Arrows" is pressed, and OnPointerDownEvent is fired.
* HeroActionBar.OnButtonClickedDown is called, and ClientInputSender.RequestAction is invoked
* in next Fixed Update, ClientInputSender creates an AOEActionInput to manage the input.
* At this point, the user's finger is still down (from his press on the button). In the old code, he would have to drag his mouse off the button to the spot where he wanted the arrow rain, and then release it, triggering the MouseUp behavior in the AOEActionInput.
* In the new code, he can release his finger, since the AOEActionInput itself requires a full button click to fire the skill.

*question* 
This logic really starts in the UIHUDButton and HeroActionBar, but the same question you raised could be applied to this system as well: why can't HeroActionBar only fire the RequestAction on mouse up? Then the AOEActionInput's logic wouldn't have to change.

*answer*
The answer is that the HeroActionBar system needs to work for a bunch of different skills, and some of them are "charged" skills, where you hold down a button to charge up an effect. These actions must be Requested on mouse down, not mouse up. This was the change that kept things contained to the AOEInputAction, without having to rethink either the action flow, or other skills.

SamuelBellomo (approved):

fernando-cortez (approved):

PR #326, opened 2021-05-10
[GOMPS-522] fixes character jitter in high packet loss scenarios

dwoodruffsf:
I have moved position and rotation to the MLAPI PositionUpdate channel, which is {Unreliable,Sequenced}. This significantly improves the user experience in high packet loss scenarios. 

I tested with clumsy, 50ms of bi-directional latency (measuring 250ms RTT in game), and 10% packet loss. Without this change, movement and camera were both horribly janky. After the change, the experience was largely tolerable, with occasional subtle disruptions (not too bad, considering 10% packet loss is pretty rough!)

### DIFF
--- a/Assets/BossRoom/Scripts/Shared/Game/Entity/NetworkCharacterState.cs
+++ b/Assets/BossRoom/Scripts/Shared/Game/Entity/NetworkCharacterState.cs
@@ -45,7 +45,8 @@ namespace BossRoom
         /// <summary>
         /// The networked position of this Character. This reflects the authoritative position on the server.
         /// </summary>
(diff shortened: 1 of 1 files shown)


fernando-cortez (approved):
Tested this before + after this addition with LiteNetLib's injected latency (min 80ms, max 120ms, 10% packet loss).

What was really noticeable was that the big teleports essentially disappeared. We need to document this for easy improvements to movement syncing.

Cosmin-B (approved):

Cosmin-B:
@lkrell See above!

PR #329, opened 2021-05-12
[GOMPS-524] fixing issue where you wade through the treasure chest dais

dwoodruffsf:
This was less simple than it seemed. Adding a CollisionBox to the dais did not solve the problem. From experimentation I found that the Player would only collide against the dais if it was non-kinematic (and even then, step-up didn't work properly). This is literally the only place in the demo where the floor is uneven, and I don't want to make a major change to the movement model just to accommodate this problem, so I have done the following:

1. Sunk the dais down to floor level. 
2. Still added a CollisionBox to the dais (this is just so that you see the move-click particle effect on top of it). 
3. Added a PathObstacle to the treasure chest--before it was elevated off the pathing plane, and that was somehow blocking navmesh from generating under it, but when the treasure chest was actually dropped down to the ground you started to be able to walk through it. 
4. Regenerated the navmesh. 

I did a bit of testing to make sure I didn't create any regressions with movement/pathing etc., but this is admittedly a bigger change than we thought it would be.


Cosmin-B (dismissed):

Assets/BossRoom/Prefabs/Dungeon/Dungeon Pieces/env_boss_floor_2.prefab
```diff
@@ -11,6 +11,7 @@ GameObject:
   - component: {fileID: 3325143161705299546}
```
fernando-cortez, line 11: I wasn't able to see the click feedback on top of the floor near the chest (around stage left area). Might need to play with heights here.

dwoodruffsf: Well that's quite mysterious. It's flat, so I don't know why it worked for only one half of the dais--and click feedback was already basing itself off the y height of the object that was clicked on. I have nudged up the click feedback hover height to to 0.15 (from 0.1), and now it appears everywhere.

fernando-cortez (approved):

Cosmin-B (approved):

PR #332, opened 2021-05-17
Feature/player data persistence

fernando-cortez:
Jira task [here](https://unity3d.atlassian.net/browse/GOMPS-253?atlOrigin=eyJpIjoiZWQ1ZTk0NjEzYjA3NDliYjhiMjA0NzVmNGJhYzI2YTAiLCJwIjoiaiJ9).

Notable changes:
- Player prefab is now PlayerCharacter prefab. Player prefab is now a new prefab and _is_ MLAPI's default player prefab. Player contains a `BossRoomPlayer` NetworkBehaviour component whereas PlayerCharacter contains a `BossRoomPlayerCharacter` NetworkBehaviour component.

- New player prefab contains NetworkBehaviours that should persist through the entirety of a connection (win, name, type states). To accomplish this, `NetworkCharacterState` has been trimmed & refactored.

- Now, lobby data is passed to each Player by the server and we've eliminated the need for passing lobby data through a GameStateRelay object. Similarly, win state is passed by the server to players on game completion, removing the need for another GameStateRelay object. Win state is fetched through the Player ScriptableObject runtime list reference. 

- Introduction of ScriptableObject runtime sets, namely `BossRoomPlayerRuntimeCollection` and `BossRoomPlayerCharacterRuntimeCollection`. These get populated on both clients and server. This is referenced in several state prefabs, and the Player/PlayerCharacter prefabs themselves. This has helped remove some hard references between systems.

- Due to a possible race condition on the initialization of NetworkObjects on late-join scene loads, a callback-based framework has been introduced for `BossRoomPlayerCharacter`. As soon as its associated `BossRoomPlayer` NetworkObject has been created, it invokes an event for all other sibling NetworkBehaviours to listen for before initializing behaviours. An example of this is seen inside of `ClientInputSender` (it needs to verify that its BossRoomPlayer _is_ the local player).

- Single-NetworkVariable NetworkBehaviour components (eg. `NetworkNameState`) have been consolidated in terms of visibility within the editor and OnValueChanged add/remove methods created. For protection, the NetworkVariable cannot be accessed externally, whereas only its value can now be retrieved/set.

- `NetworkBehaviourLookup`, a utility component for quick lookups on sibling NetworkBehaviours. Has an associated Editor script for quick population.

- Controversial, but I exercised code cleanup on any file I may have modified so there's quite a bit of refactoring here.


Assets/BossRoom/Scripts/Server/Net/ServerGameNetPortal.cs
```diff
@@ -312,25 +278,34 @@ private IEnumerator WaitToDisconnectClient(ulong clientId, ConnectStatus reason)
...
         {
-            m_ClientData.Add("host_guid", new PlayerData(m_Portal.PlayerName, m_Portal.NetManager.LocalClientId));
-            m_ClientIDToGuid.Add(m_Portal.NetManager.LocalClientId, "host_guid");
-        }
+            var localClientID = NetworkManager.Singleton.LocalClientId;
 
+            m_ClientData.Add("host_guid", new PlayerData(m_Portal.PlayerName, localClientID));
+            m_ClientIDToGuid.Add(localClientID, "host_guid");
+
+            var newPlayer = Instantiate(m_PlayerPrefab, Vector3.zero, Quaternion.identity, null);
```
SamuelBellomo, line 302: let's make sure we can't use MLAPI's default player spawn flow before spawning ourselves. And please document in code anything you find

Assets/BossRoom/Scripts/Shared/Game/Entity/BossRoomPlayerCharacter.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            if (!m_BossRoomPlayer)
+            {
+                m_BossRoomPlayers.ListChanged += TryGetPlayer;
+            }
+
+            m_BossRoomPlayerCharacters.Add(this);
+        }
+
+        void TryGetPlayer()
```
SamuelBellomo, line 47: as discussed, let's try to have one single consolidated "all objects spawned for the first time" callback

Assets/BossRoom/Scripts/Shared/Game/NetworkBehaviourLookup.cs
```diff
@@ -0,0 +1,31 @@
...
+using MLAPI;
+using UnityEngine;
+
+namespace BossRoom
+{
+    /// <summary>
+    /// A utility class that can be added to a NetworkObject to collect its NetworkBehaviours for quick lookups.
+    /// </summary>
+    [RequireComponent(typeof(NetworkObject))]
+    public class NetworkBehaviourLookup : MonoBehaviour
```
SamuelBellomo, line 11: as discussed, let's validate GetComponent only allocates in editor

SamuelBellomo: this might not be needed

Assets/BossRoom/Scripts/Shared/NetworkNameState.cs
```diff
@@ -7,9 +7,25 @@ namespace BossRoom
     /// <summary>
     /// NetworkBehaviour containing only one NetworkVariableString which represents this object's name.
     /// </summary>
-    public class NetworkNameState : NetworkBehaviour
+    public class NetworkNameState : NetworkBehaviour, INetworkSubscribable<string>
```
SamuelBellomo, line 10: As discussed, netvars are meant to be used directly in gameplay. We shouldn't wrap them just to hide netvars from gameplay code. Because this is such a simple class here that's meant as a container for these values, it might not make sense to have encapsulation for this.
If you have other cases for it being dangerous to expose to users, please do tell, it might be a need for a refactor netvar side.

PR #345, opened 2021-05-28
radio button introduced for main menu UX improvements (default is still IP)

fernando-cortez:
Jira task [here](https://unity3d.atlassian.net/browse/GOMPS-543?atlOrigin=eyJpIjoiZTZkYzViYzQzNDVmNDhkNGEyNjU0ZmMzNGJjZTg0NTciLCJwIjoiaiJ9).

The attached screenshot should provide more information than the summary :).
image

Changes:
- Added a `Toggle Group` parent to `PopupPanel` prefab, with toggles for each connection type (IP & relay).
- Added a bit more instructive text to hosting (both on IP and Relay).


Assets/BossRoom/Scripts/Client/UI/UIRadioButton.cs
```diff
@@ -0,0 +1,51 @@
+using System;
```
SamuelBellomo, line 1: As discussed, let's try using toggle groups instead of having our own implementation

fernando-cortez: Done.

Cosmin-B (approved):

SamuelBellomo (approved):

PR #356, opened 2021-07-19
Feature/player persistence visuals

fernando-cortez:
Part 3/3 of Player Persistence. Jira [here](https://unity3d.atlassian.net/browse/GOMPS-550?atlOrigin=eyJpIjoiYWEzMmMxN2Y2MzE5NGRiYmJlM2U0N2JmMzQ4YWFhODciLCJwIjoiaiJ9).

This PR fixes any visuals that were omitted from Part 2. This re-introduces names/health/portrait to `PartyHud`, and re-introduces the world-space displayed name for a player avatar.

Notable changes:
* Graphics variants have been created for every avatar, 8 in total. These get spawned (& cached for reuse) inside `CharSelect` scene.
* `GameStateRelay` has been removed, as character select data can be passed to `PersistentPlayer`. A late-joiner gets assigned a random avatar class.
* A `NetworkGameState` NetworkObject is spawned alongside the host. This keeps track of the win state. Server sets this component's win state, and clients read from this state value when presenting win/loss screen.
* `PartyHud` refactor to initialize state as avatars are spawned inside BossRoom scene. This removed a fair amount of hard referencing inside of `ClientCharacterVisualization`. Systems are decoupled.


fernando-cortez:
Placed on hold to verify potential refactorings from MLAPI features.

Assets/BossRoom/Scripts/Client/Game/State/ClientCharSelectState.cs
```diff
@@ -227,24 +228,39 @@ private void UpdateCharacterSelection(CharSelectData.SeatState state, int seatId
...
+
                 m_ClassInfoBox.ConfigureForNoSelection();
             }
             else
             {
-                if ( seatIdx != -1 )
+                if (seatIdx != -1)
                 {
-                    // TODO: Re-implement CharSelect preview (GOMPS-550)
-                    //m_InSceneCharacter.gameObject.SetActive(true);
```
Cosmin-B, line 239: I was wondering when this will be done

Cosmin-B (dismissed):

pdeschain:
Had a call with Fernando - we discussed changing PlayerAvatar structure now instead of later, incorporating NetworkAnimator. This way we won't need to deal with another PR that changes structure of how player gameobjects are structured.

pdeschain (changes requested):
see prior comment

VALERE91Unity (approved):

Assets/BossRoom/Scripts/Client/Game/Action/MeleeActionFX.cs
```diff
@@ -44,7 +45,17 @@ public override bool Start()
...
+                if (PhysicsWrapper.TryGetPhysicsWrapper(Data.TargetIds[0], out var physicsWrapper))
+                {
+                    targetPosition = physicsWrapper.Transform.position;
+                }
+                else
+                {
+                    targetPosition = targetNetworkObj.transform.position;
+                }
+
+                if ((m_Parent.transform.position - targetPosition).sqrMagnitude < (padRange * padRange))
```
pdeschain, line 58: On a stylistic note - I would much rather prefer this boolean value to be captured into a variable with an explicit name VS having a comment. Has to do with comments being fragile code smells almost in all cases.

pdeschain: It's not a request to change anything - more so a general observation about our codebase. We have these comments that describe what's happening in the logic, where a properly placed and named variable would do a better job in the readability and maintainability department.

pdeschain (approved):

Assets/BossRoom/Scripts/Client/Game/Action/MeleeActionFX.cs
fernando-cortez: Agree on this. Best to just be explicit.

hsdgd (commented):
-

PR #361, opened 2021-09-01
Feature/session manager

ryantwarner:
Session manager skeleton


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=361) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=361) before we can accept your contribution.<br/>**2** out of **3** committers have signed the CLA.<br/><br/>:white_check_mark: dwoodruffsf<br/>:white_check_mark: eheimburg<br/>:x: lkrell<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=361) it.</sub>

Assets/BossRoom/Scripts/Client/Net/ClientSessionManager.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        private void OnDestroy()
+        {
+            m_Portal.NetManager.OnClientConnectedCallback -= OnClientConnected;
+            m_Portal.NetManager.OnClientDisconnectCallback -= OnClientDisconnected;
+        }
+
+        private void OnClientConnected(ulong clientID)
+        {
+            if (!m_ClientIDs.Contains(clientID))
```
SamuelBellomo, line 43: Issue with ClientIDs is that they are reused if players reconnect. So clientA and clientB could have client IDs 1 and 2. if both disconnect and reconnect, clientA could have id 2 and clientB could have id 1.
We already have a session guid used to save your character class, I would have consolidated those into your current solution (which seems cleaner :) )

ryantwarner: I'm not sure how to get the clientGUIDs, I tried changing it to OnConnectionApproval and processing the payload, but I'm not sure how to get that info.

Assets/BossRoom/Scripts/Client/Net/ClientSessionManager.cs
```diff
@@ -0,0 +1,41 @@
...
+using MLAPI.Transports.UNET;
+using Photon.Realtime;
+
+namespace BossRoom.Client
+{
+    /// <summary>
+    /// Client side logic for a GameNetPortal. Contains implementations for all of GameNetPortal's S2C RPCs.
+    /// </summary>
+    [RequireComponent(typeof(GameNetPortal))]
+    public class ClientSessionManager : MonoBehaviour
```
SamuelBellomo, line 17: looking at the empty ClientSceneChanged, does this class do anything anymore?

Assets/BossRoom/Scripts/Server/Net/ServerSessionManager.cs
```diff
@@ -0,0 +1,272 @@
...
+        /// to send back more than a "true/false", which means we have to work a little harder to provide a useful error return to the client. To do that, we invoke a
+        /// client RPC in the same channel that MLAPI uses for its connection callback. Since that channel ("MLAPI_INTERNAL") is both reliable and sequenced, we can be
+        /// confident that our login result message will execute before any disconnect message.
+        /// </remarks>
+        /// <param name="connectionData">binary data passed into StartClient. In our case this is the client's GUID, which is a unique identifier for their install of the game that persists across app restarts. </param>
+        /// <param name="clientId">This is the clientId that MLAPI assigned us on login. It does not persist across multiple logins from the same client. </param>
+        /// <param name="callback">The delegate we must invoke to signal that the connection was approved or not. </param>
+        private void ApprovalCheck(byte[] connectionData, ulong clientId, MLAPI.NetworkManager.ConnectionApprovedDelegate callback)
```
SamuelBellomo, line 173: wait, there's two ApprovalChecks?
mmm looking further, looks like callback isn't used. So basically, we're using this one here to insert ourselves during the connection approval process? Isn't there another callback we could use for this, like OnClientConnectedCallback?

Assets/BossRoom/Scripts/Server/Net/ServerSessionManager.cs
```diff
@@ -0,0 +1,272 @@
...
+            if (m_ClientData.ContainsKey(connectionPayload.clientGUID))
+            {
+                if (Debug.isDebugBuild)
+                {
+                    Debug.Log($"Client GUID {connectionPayload.clientGUID} already exists. Because this is a debug build, we will still accept the connection");
+                    while (m_ClientData.ContainsKey(connectionPayload.clientGUID)) { connectionPayload.clientGUID += "_Secondary"; }
+                }
+                else
+                {
+                    gameReturnStatus = ConnectStatus.LoggedInAgain;
```
SamuelBellomo, line 201: beside for checking for success, looks like we're not using that status for anything else?

Assets/BossRoom/Scripts/Server/Net/ServerSessionManager.cs
```diff
@@ -0,0 +1,272 @@
...
+        public Vector3 PlayerPosition;
+        public Vector3 PlayerRotation;
+        public bool IsConnected;
+
+        public SessionPlayerData(ulong clientID, string clientGUID, string name, Vector3 position, Vector3 rotation, bool isConnected = false)
+        {
+            ClientID = clientID;
+            ClientGUID = clientGUID;
+            PlayerName = name;
+            PlayerPosition = position;
```
SamuelBellomo, line 24: tested by trying to connect client with host, then having the client disconnect and reconnect.
class was the right one (instead of the default tank), but position was reset to spawn point.
not sure about the name since we seem to have a bug on our hud preventing the display of the name :(

Assets/BossRoom/Scripts/Shared/Game/Entity/NetworkCharacterState.cs
```diff
@@ -113,7 +113,7 @@ public LifeState LifeState
         public override void OnNetworkSpawn()
         {
             if (!IsServer) return;
-            HitPoints = CharacterClass.BaseHP.Value;
+            if (HitPoints == 0) HitPoints = CharacterClass.BaseHP.Value;
         }
```
fernando-cortez, line 117: We should have the health initialized at the same place in code. Since we're doing that now for the late-joined player inside ServerBossRoomState, we might as well just do this there for a fresh player.

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ServerGameNetPortal.cs
```diff
@@ -1,346 +1,265 @@
...
+    {
+        public string m_PlayerName;  //name of the player
+        public ulong m_ClientID; //the identifying id of the client
+
+        public PlayerData(string playerName, ulong clientId)
+        {
+            m_PlayerName = playerName;
+            m_ClientID = clientId;
+        }
+    }
```
fernando-cortez, line 24: This can be completely removed now right?

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/SessionManager.cs
```diff
@@ -0,0 +1,402 @@
...
+        /// </summary>
+        private Dictionary<string, SessionPlayerData> m_ClientData;
+
+        /// <summary>
+        /// Map to allow us to cheaply map from guid to player data.
+        /// </summary>
+        private Dictionary<ulong, string> m_ClientIDToGuid;
+
+        // used in ApprovalCheck. This is intended as a bit of light protection against DOS attacks that rely on sending silly big buffers of garbage.
+        private const int k_MaxConnectPayload = 1024;
```
fernando-cortez, line 67: Seems like this was brought over from ServerGameNetPortal and could be removed.

Assets/BossRoom/Scripts/Server/Game/Character/PlayerServerCharacter.cs
```diff
@@ -37,6 +37,12 @@ void OnDisable()
             s_ActivePlayers.Remove(m_CachedServerCharacter);
         }
 
+        public override void OnNetworkDespawn()
+        {
+            var movementTransform = m_CachedServerCharacter.Movement.transform;
+            SessionManager.Instance.UpdatePlayerCharacterData(OwnerClientId, movementTransform.position, movementTransform.rotation.eulerAngles, m_CachedServerCharacter.NetState.HitPoints);
```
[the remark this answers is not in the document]
LPLafontaineB: I don't think so. The data is saved by a Vec3, but could be by a Quaternion too. I'll change it

Assets/BossRoom/Scripts/Server/Game/State/ServerBossRoomState.cs
```diff
@@ -180,8 +182,18 @@ private void SpawnPlayer(ulong clientId, bool lateJoin)
             // if joining late, assign a random character to the persistent player
             if (lateJoin)
             {
-                persistentPlayer.NetworkAvatarGuidState.AvatarGuid.Value =
-                    m_AvatarRegistry.GetRandomAvatar().Guid.ToNetworkGuid();
+                SessionPlayerData? sessionPlayerData = SessionManager.Instance.GetPlayerData(clientId);
```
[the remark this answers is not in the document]
LPLafontaineB: Actually, on the SessionManager's side, it uses the clientID to get the guid, for which the mapping is updated each time a new client connects. But we could skip that step and simply get the guid here instead via ClientPrefs and pass that to the SessionManager

Assets/BossRoom/Scripts/Server/Game/State/ServerBossRoomState.cs
```diff
@@ -180,8 +182,18 @@ private void SpawnPlayer(ulong clientId, bool lateJoin)
...
+                if (sessionPlayerData != null && sessionPlayerData.Value.IsReconnecting)
+                {
+                    physicsTransform.SetPositionAndRotation(sessionPlayerData.Value.PlayerPosition, Quaternion.Euler(sessionPlayerData.Value.PlayerRotation));
+                    newPlayerCharacter.NetState.HitPoints = sessionPlayerData.Value.CurrentHP;
+                }
+                else
+                {
+                    persistentPlayer.NetworkAvatarGuidState.AvatarGuid.Value =
```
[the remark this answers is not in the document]
LPLafontaineB: You're right I think I can move the random avatar assignment to the SessionManager.

Assets/BossRoom/Scripts/Shared/Game/Entity/NetworkCharacterState.cs
LPLafontaineB: I thought about it too, but this code is used for NPCs as well. So this is needed for them. I'll try to see if I can do the reverse and get the info from the SessionManager here and remove the setting of HP in ServerBossRoomState

fernando-cortez: That's very true. Hopefully that works out though.

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/SessionManager.cs
LPLafontaineB: Yes you're right, I missed this one, thanks

Assets/BossRoom/Scripts/Server/Game/State/ServerBossRoomState.cs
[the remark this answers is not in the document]
LPLafontaineB: Nevermind, it doesn't work that way since this code is executed on the host, so we don't have access to the guid through ClientPrefs, and will have to get it via the client ID anyways. Anyways this is fine since the mapping of client IDs and GUIDs is handled by the SessionManager. I just tested it with 2 clients, one via parrelsync and one via a build (I had to revert commit 7315fae to be able to test this) and no matter the order of connection/disconnection/reconnection they always reconnect as the right character.
However I also tested it with 2 clients via parrelsync and there was an issue, the modified guid (adding _Secondary) is not handled correctly and if both clients disconnect and reconnect, they currently both reconnect as the first client's character. I'll look into this.

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/SessionManager.cs
```diff
@@ -0,0 +1,402 @@
...
+        {
+            if (m_ClientIDToGuid.TryGetValue(clientId, out var guid))
+            {
+                if (m_ClientData[guid].ClientID == clientId)
+                {
+                    var sessionPlayerData = m_ClientData[guid];
+                    sessionPlayerData.IsConnected = false;
+                    sessionPlayerData.IsReconnecting = false;
+                    m_ClientData[guid] = sessionPlayerData;
+                    // Wait a few seconds before removing the clientId-GUID mapping to allow for final updates to SessionPlayerData.
```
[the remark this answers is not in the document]
LPLafontaineB: The final updates to the sessionPlayerData come in the PlayerServerCharacter's and PersistentPlayer's OnDespawn event, which are called after this.

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ServerGameNetPortal.cs
```diff
@@ -1,346 +1,265 @@
...
+                connectionApprovedCallback(true, null, true, null, null);
+                return;
+            }
+
+            ConnectStatus gameReturnStatus = ConnectStatus.Success;
+
+            // Test for over-capacity connection. This needs to be done asap, to make sure we refuse connections asap and don't spend useless time server side
+            // on invalid users trying to connect
+            // todo this is currently still spending too much time server side.
+            if (SessionManager.Instance.IsServerFull())
```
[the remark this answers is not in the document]
LPLafontaineB: I'll move this back to the GameNetPortal then, and use the number of connected clients from NetManager instead.

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/SessionManager.cs
```diff
@@ -0,0 +1,402 @@
...
+            }
+
+            foreach (var id in idsToClear)
+            {
+                m_ClientData.Remove(m_ClientIDToGuid[id]);
+                m_ClientIDToGuid.Remove(id);
+            }
+        }
+
+        static void AssignPlayerName(ulong clientId, string playerName)
```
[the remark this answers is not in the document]
LPLafontaineB: I thought it was the session manager's responsibility to handle player data that it wants to keep and to reassign, but maybe it could simply carry the information instead. I suppose that I could remove these and make the PersistentPlayer assign them to itself, but I would need to make sure that it does so after the session manager has correctly updated its data.

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/SessionManager.cs
```diff
@@ -0,0 +1,403 @@
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
```
SamuelBellomo, line 3: Linq will allocate a bunch and is usually not friendly to game dev. Any reason we're using it?

LPLafontaineB: It is for this check `m_Portal.NetManager.ConnectedClientsIds.Contains(id)` in the `OnGameEnded` method. I'll look for an alternative

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/SessionManager.cs
```diff
@@ -0,0 +1,402 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Updates the player's session persistant data
+        /// </summary>
+        /// <param name="clientId">Id of the client to update</param>
+        /// <param name="playerName">The player's name to save</param>
+        /// <param name="avatarNetworkGuid">The NetworkGuid describing the player's class</param>
+        public void UpdatePlayerPersistantData(ulong clientId, string playerName, NetworkGuid avatarNetworkGuid)
```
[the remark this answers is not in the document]
LPLafontaineB: In the end, I made it a bit more generic and reusable the only thing that would need to change to reuse it in another project would be the struct holding the data. In parallel I worked on a refactor abstracting all session data using generic types, but the refactored version still has some issues and would still need some work. I think this current version is sufficient for the moment, and would prefer to have this one reviewed and then merge it after eventual new comments are addressed. Then later if we deem it worth it I'll work on ironing out the refactored version and create a new PR for it.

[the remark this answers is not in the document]
LPLafontaineB: Finally the refactored version was easier to iron out than expected, so I pushed it to this branch instead

fernando-cortez (commented):
Nitpicking, but I would zero out the position of the session manager prefab and in scene. Will approve once that's updated.

@SamuelBellomo I think your comments have been addressed too.

fernando-cortez (dismissed):

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/SessionManager.cs
```diff
@@ -0,0 +1,277 @@
...
+namespace Unity.Multiplayer.Samples.BossRoom
+{
+    public interface ISessionPlayerData
+    {
+        bool IsConnected { get; set; }
+        ulong ClientID { get; set; }
+        void Reinitialize();
+    }
+
+    public class SessionManager<T> : MonoBehaviour where T : struct, ISessionPlayerData
```
SamuelBellomo, line 15: Any reason why this is a Monobehaviour? Having this as a pure C# class would make this way easier to manipulate

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/SessionManager.cs
```diff
@@ -0,0 +1,277 @@
...
+        /// Handles the case where NetworkManager has told us a client has disconnected. This includes ourselves, if we're the host,
+        /// and the server is stopped."
+        /// </summary>
+        private void OnClientDisconnect(ulong clientId)
+        {
+            if (m_ClientIDToGuid.TryGetValue(clientId, out var guid))
+            {
+                if (GetPlayerData(guid)?.ClientID == clientId)
+                {
+                    var T = m_ClientData[guid];
```
SamuelBellomo, line 81: T --> let's use complete names please

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/SessionManager.cs
```diff
@@ -0,0 +1,277 @@
...
+        {
+            if (m_NetworkManager)
+            {
+                m_NetworkManager.OnServerStarted -= ServerStartedHandler;
+            }
+        }
+
+        public void AddHostData(T sessionPlayerData)
+        {
+            m_ClientData.Add("host_guid", sessionPlayerData);
```
SamuelBellomo, line 67: magic string "host_guid" should be a const somewhere

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/GameNetPortal.cs
```diff
@@ -169,7 +174,7 @@ public void StartHost(string ipaddress, int port)
                 default:
                     throw new Exception($"unhandled IpHost transport {chosenTransport.GetType()}");
             }
-
+            BossRoomSessionManager.Instance.AddHostData(new SessionPlayerData(NetManager.LocalClientId, PlayerName, m_AvatarRegistry.GetRandomAvatar().Guid.ToNetworkGuid(), 0, true));
```
SamuelBellomo, line 177: why do we need to add a random selection on start host? Can't we wait for the player to select once in the char select scene?

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/SessionManager.cs
LPLafontaineB: It is currently placed as a script in a Gameobject, and uses the Awake(), Start() and OnDestroy() events. It also uses the "enabled" field to disable itself on clients. We could make it not a Monobehavior, and have BossRoomSessionManager have this as a field instead of inheriting from it, and refactor it to make it work, but I'm not sure if it would really make it better. I think it would be better to keep most of the logic in SessionManager so that is is easier to reuse

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/GameNetPortal.cs
LPLafontaineB: I added his here mostly to remove the responsibility of adding a random class on late join from BossRoomState. It also avoids [this issue](https://jira.unity3d.com/browse/MTT-1464)

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/SessionManager.cs
SamuelBellomo: Your Awake, Start and OnDestroy become useless once you port sessionManager to a self contained class.
Awake's logic becomes singleton logic, Start logic becomes constructor logic, OnDestroy logic becomes destructor logic.
"Enabled" is only useful if your method has Unity callbacks. Thing is, looking at our logic here, I'm not sure I see how this works (since ServerStartedHandler is registered in Start, which means there's no other monobehaviour callbacks which could benefit from enabling/disabling

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/SessionManager.cs
```diff
@@ -0,0 +1,277 @@
...
+            if (gameReturnStatus == ConnectStatus.Success)
+            {
+                m_ClientIDToGuid[clientId] = clientGUID;
+                m_ClientData[clientGUID] = sessionPlayerData;
+            }
+
+            return gameReturnStatus;
+        }
+
+        public string GetPlayerGUID(ulong clientID)
```
SamuelBellomo, line 175: never used?

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/SessionManager.cs
SamuelBellomo:
```suggestion
    // GUID binds a player to a session
    // Once that player connects to a host, the host associates the current ClientID to the player's session GUID.
    // If the player disconnects and reconnects to the same host, the session is preserved.
    public class SessionManager<T> : MonoBehaviour where T : struct, ISessionPlayerData
```

SamuelBellomo: Does that summarize it?

LPLafontaineB: Oh you're right, it doesn't even need to be inside a gameobject, I didn't think about that

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/SessionManager.cs
LPLafontaineB: indeed, not used anymore, will remove it

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/SessionManager.cs
```diff
@@ -0,0 +1,280 @@
...
+            {
+                //O__O if adding any event registrations here, please add an unregistration in OnClientDisconnect.
+                m_NetworkManager.OnClientDisconnectCallback += OnClientDisconnect;
+            }
+        }
+
+        /// <summary>
+        /// Reinitializes session data from connected players, and clears data from disconnected players, so that if they reconnect in the next game, they will be treated as new players
+        /// </summary>
+        public void OnGameEnded()
```
SamuelBellomo, line 249: "OnSessionEnded"? Just so are consistent in our API

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/SessionManager.cs
```diff
@@ -12,8 +12,16 @@ public interface ISessionPlayerData
         void Reinitialize();
     }
 
+    // GUID binds a player to a session
+    // Once that player connects to a host, the host associates the current ClientID to the player's session GUID.
+    // If the player disconnects and reconnects to the same host, the session is preserved.
+    // GUID binds a player to a session
+    // Once that player connects to a host, the host associates the current ClientID to the player's session GUID.
+    // If the player disconnects and reconnects to the same host, the session is preserved.
```
SamuelBellomo, line 20:
```suggestion
```

SamuelBellomo: I think I hit the "commit" button and you added it as well

SamuelBellomo:
Beside the monobehaviour removal, I'm all good with this!

fernando-cortez (approved):

VALERE91Unity (dismissed):
That's a really cool feature :) we should extract some of that knowledge and create a bite-size sample about Session Management based on that work @LukeStampfli @Cosmin-B

LPLafontaineB:
Sorry for that last commit dismissing your reviews, there was one last thing that I missed during my last refactor

VALERE91Unity (approved):

SamuelBellomo (approved):

PR #366, opened 2021-09-14
Integrating Unity Transport and Relay into Boss Room

Cosmin-B:
Will follow up with a better description!


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=366) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=366) before we can accept your contribution.<br/>**0** out of **2** committers have signed the CLA.<br/><br/>:x: wackoisgod<br/>:x: VALERE91<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=366) it.</sub>

Assets/BossRoom/Scripts/Client/Net/ClientGameNetPortal.cs
```diff
@@ -211,6 +218,91 @@ public static bool StartClientRelayMode(GameNetPortal portal, string roomKey, ou
...
+        {
+
+            var chosenTransport  = NetworkManager.Singleton.gameObject.GetComponent<TransportPicker>().UnityRelayTransport;
+            NetworkManager.Singleton.NetworkConfig.NetworkTransport = chosenTransport;
+
+            switch (chosenTransport)
+            {
+                case UnityTransport utp:
+                    Debug.Log($"Setting Unity Relay client with join code {joinCode}");
+                    Task t = Task.Run( () => {UnityServices.InitializeAsync();});
```
SamuelBellomo, line 231: "Initialize" seems like a one time thing. we initialize each time we connect? What if we finish a game and reconnect to another host, in the same run?

Assets/BossRoom/Scripts/Client/Net/ClientGameNetPortal.cs
```diff
@@ -211,6 +218,91 @@ public static bool StartClientRelayMode(GameNetPortal portal, string roomKey, ou
...
+                    if (!t.Wait(30000))
+                    {
+                        Debug.Log("UnityServices.Initialize did not finish within 30 seconds");
```
SamuelBellomo, line 234: should we throw or early exit here if we didn't initialize properly?

Assets/BossRoom/Scripts/Client/Net/ClientGameNetPortal.cs
```diff
@@ -211,6 +218,91 @@ public static bool StartClientRelayMode(GameNetPortal portal, string roomKey, ou
...
+
+            var chosenTransport  = NetworkManager.Singleton.gameObject.GetComponent<TransportPicker>().UnityRelayTransport;
+            NetworkManager.Singleton.NetworkConfig.NetworkTransport = chosenTransport;
+
+            switch (chosenTransport)
+            {
+                case UnityTransport utp:
+                    Debug.Log($"Setting Unity Relay client with join code {joinCode}");
+                    Task t = Task.Run( () => {UnityServices.InitializeAsync();});
+                    if (!t.Wait(30000))
```
SamuelBellomo, line 232: I'd set a const for this timeout value

Assets/BossRoom/Scripts/Client/Net/ClientGameNetPortal.cs
```diff
@@ -211,6 +218,91 @@ public static bool StartClientRelayMode(GameNetPortal portal, string roomKey, ou
...
+                    {
+                        Debug.Log("UnityServices.Initialize did not finish within 30 seconds");
+                    }
+                    Debug.Log(AuthenticationService.Instance);
+                    if (!AuthenticationService.Instance.IsSignedIn)
+                    {
+                        Task t2 = Task.Run( () => {AuthenticationService.Instance.SignInAnonymouslyAsync();});
+                        t2.Wait(30000);
```
SamuelBellomo, line 240: We don't check for the task result here, like we do for init earlier?

Assets/BossRoom/Scripts/Client/Net/ClientGameNetPortal.cs
```diff
@@ -211,6 +218,91 @@ public static bool StartClientRelayMode(GameNetPortal portal, string roomKey, ou
...
+                        var playerId = AuthenticationService.Instance.PlayerId;
+                        Debug.Log(playerId);
+
+                        var clientRelayUtilityTask = RelayUtility.JoinRelayServerFromJoinCode(joinCode);
```
SamuelBellomo, line 244: we only join at first signin?

Assets/BossRoom/Scripts/Client/Net/ClientGameNetPortal.cs
SamuelBellomo: Also, is there any error messages we could display? Or any actions we could take on these errors? Let's say I get a 503 because some services are overloaded, could it be possible in a first pass to log that error and add a todo to add exponential backoff in a second pass?

Assets/BossRoom/Scripts/Shared/Net/UnityRelayUtilities.cs
```diff
@@ -0,0 +1,82 @@
...
+                throw;
+            }
+
+            return (allocation.RelayServer.IpV4, (ushort) allocation.RelayServer.Port, allocation.AllocationIdBytes,
+                allocation.ConnectionData, allocation.Key, createJoinCodeResponse.Result.Data.JoinCode);
+        }
+
+        async public static
+            Task<(string ipv4address, ushort port, byte[] allocationIdBytes, byte[] connectionData, byte[]
+                hostConnectionData, byte[] key)> JoinRelayServerFromJoinCode(string joinCode)
```
SamuelBellomo, line 57: That util seems very useful, shouldn't it be part of Unity Relay SDK?

Assets/BossRoom/Scripts/Client/Net/ClientGameNetPortal.cs
```diff
@@ -211,6 +218,91 @@ public static bool StartClientRelayMode(GameNetPortal portal, string roomKey, ou
...
+                        clientRelayUtilityTask.Wait(30000);
+                        if (clientRelayUtilityTask.IsFaulted)
+                        {
+                            throw new Exception($"Failed to allocate on relay server {clientRelayUtilityTask.Exception?.Message}");
+                        }
+
+                        var (ipv4address, port, allocationIdBytes, connectionData, hostConnectionData, key) = clientRelayUtilityTask.Result;
+
+                        utp.SetRelayServerData(ipv4address, port, allocationIdBytes, key, connectionData, hostConnectionData);
```
SamuelBellomo, line 253: It seems weird that this is user side code, that seems like the kind of thing that should be automated for us

Assets/BossRoom/Scripts/Client/UI/MainMenuUI.cs
```diff
@@ -78,6 +83,16 @@ public void OnConnectClicked()
                     case OnlineMode.IpHost:
                         ClientGameNetPortal.StartClient(m_GameNetPortal, connectInput, connectPort);
                         break;
+
+                    case OnlineMode.UnityRelay:
+                        Debug.Log($"Unity Relay Client, join code {connectInput}");
+                        ClientGameNetPortal.StartClientUnityRelayModeAsync(m_GameNetPortal, connectInput);
+                        // if (ClientGameNetPortal.StartClientUnityRelayMode(m_GameNetPortal, connectInput, out string failMessage2) == false)
```
SamuelBellomo, line 90: comment should be removed?

Assets/BossRoom/Scripts/Client/UI/MainMenuUI.cs
```diff
@@ -78,6 +83,16 @@ public void OnConnectClicked()
                     case OnlineMode.IpHost:
                         ClientGameNetPortal.StartClient(m_GameNetPortal, connectInput, connectPort);
                         break;
+
+                    case OnlineMode.UnityRelay:
+                        Debug.Log($"Unity Relay Client, join code {connectInput}");
+                        ClientGameNetPortal.StartClientUnityRelayModeAsync(m_GameNetPortal, connectInput);
+                        // if (ClientGameNetPortal.StartClientUnityRelayMode(m_GameNetPortal, connectInput, out string failMessage2) == false)
+                        // {
+                        //     m_ResponsePopup.SetupNotifierDisplay("Connection Failed", failMessage2, false, true);
```
SamuelBellomo, line 92: I don't see any popup in StartClientUnityRelayModeAsync

Assets/BossRoom/Scripts/Shared/Net/GameNetPortal.cs
```diff
@@ -224,6 +227,10 @@ public void StartHost(string ipaddress, int port)
                     unetTransport.ConnectAddress = ipaddress;
                     unetTransport.ServerListenPort = port;
                     break;
+                case UnityTransport UnityTransport:
+                    // UnityTransport. = ipaddress;
```
SamuelBellomo, line 231: ?
Is there anyway to connect to another IP without relay?

Assets/BossRoom/Scripts/Shared/Net/GameNetPortal.cs
```diff
@@ -248,6 +255,47 @@ public void StartRelayHost(string roomName)
...
+                    // TODO: This needs to be removed ?
+                   // Unity.Services.Relay.RelayService.Configuration.BasePath = "https://relay-allocations-stg.services.api.unity.com";
+                    await UnityServices.InitializeAsync();
+                    if (!AuthenticationService.Instance.IsSignedIn)
+                    {
+                        await AuthenticationService.Instance.SignInAnonymouslyAsync();
+                        var playerId = AuthenticationService.Instance.PlayerId;
+                        Debug.Log(playerId);
+                        // we now need to get the joinCode?
+                        var serverRelayUtilityTask = RelayUtility.AllocateRelayServerAndGetJoinCode(10);
```
SamuelBellomo, line 276: magic number 10?

Assets/BossRoom/Scripts/Shared/Net/NetworkVariableGUID.cs
```diff
@@ -0,0 +1,92 @@
+using System;
+using System.IO;
+using Unity.Netcode;
+using UnityEngine;
+
+namespace BossRoom
+{
+    [Serializable]
+    public class NetworkVariableGUID : NetworkVariableBase
```
SamuelBellomo, line 9: I'm guessing this branch isn't up to date with Val's latest changes?

Assets/BossRoom/Scripts/Shared/Net/UnityRelayUtilities.cs
```diff
@@ -0,0 +1,82 @@
+﻿using System;
+using System.Threading.Tasks;
+using UnityEngine;
+using Unity.Services.Relay;
+using Unity.Services.Relay.Allocations;
+using Unity.Services.Relay.Models;
+
+namespace BossRoom
+{
+    public static class RelayJoinCodeThing
```
SamuelBellomo, line 10: lol "thing"
do we need this class?

Assets/BossRoom/Scripts/Shared/Net/UnityRelayUtilities.cs
SamuelBellomo: same for the other one, why aren't these in the Relay SDK?

Assets/BossRoom/Scripts/Shared/Net/UnityRelayUtilities.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            var allocation = allocationResponse.Result.Data.Allocation;
+
+            Debug.Log($"server: {allocation.ConnectionData[0]} {allocation.ConnectionData[1]}");
+            Debug.Log($"server: {allocation.AllocationId}");
+
+            try
+            {
+                createJoinCodeResponse = await RelayService.AllocationsApiClient.CreateJoincodeAsync(
+                    new CreateJoincodeRequest(
```
SamuelBellomo, line 42: Who should we tell there's a typo in their API? It's missing a capital C for Code

Assets/BossRoom/Scripts/Shared/Net/UnityRelayUtilities.cs
```diff
@@ -0,0 +1,82 @@
...
+                        new JoinCodeRequest(allocationResponse.Result.Data.Allocation.AllocationId)));
+            }
+            catch
+            {
+                Debug.LogError("Relay create join code request failed");
```
SamuelBellomo, line 47: no error message available to print/display on screen?

Assets/BossRoom/Scripts/Shared/Net/UnityRelayUtilities.cs
```diff
@@ -0,0 +1,82 @@
...
+            }
+
+            var allocation = allocationResponse.Result.Data.Allocation;
+
+            Debug.Log($"server: {allocation.ConnectionData[0]} {allocation.ConnectionData[1]}");
+            Debug.Log($"server: {allocation.AllocationId}");
+
+            try
+            {
+                createJoinCodeResponse = await RelayService.AllocationsApiClient.CreateJoincodeAsync(
```
SamuelBellomo, line 41: I know this isn't your code, I'm not sure who to send it to. That api is confusing me. Why can't CreateJoincodeasync just take as an argument an allocationID? Why do we need so many allocations?
Also why the double new?
new CreateJoincodeRequest(new JoinCodeRequest(...

Assets/BossRoom/Scripts/Shared/TransportPicker.cs
```diff
@@ -26,12 +29,20 @@ public class TransportPicker : MonoBehaviour
...
     {
-        Assert.IsTrue(m_IpHostTransport == null || (m_IpHostTransport as UNetTransport || m_IpHostTransport as LiteNetLibTransport),
-            "IpHost transport must be either Unet or LiteNetLib transport.");
+        Assert.IsTrue(m_IpHostTransport == null || (m_IpHostTransport as UNetTransport || m_IpHostTransport as LiteNetLibTransport || m_IpHostTransport as UnityTransport),
+            "IpHost transport must be either Unet or LiteNetLib transport or UTP.");
 
         Assert.IsTrue(m_RelayTransport == null || (m_RelayTransport as PhotonRealtimeTransport), "" +
             "Relay transport must be PhotonRealtimeTransport");
+
+        // Assert.IsTrue(m_UnityRelayTransport == null || (m_UnityRelayTransport as UnityTransport), "" +
```
SamuelBellomo, line 45: should be removed?

Packages/packages-lock.json
```diff
@@ -55,14 +55,14 @@
...
       "depth": 1,
       "source": "registry",
       "dependencies": {
-        "com.unity.test-framework.performance": "2.3.1-preview",
-        "com.unity.burst": "1.3.2"
+        "com.unity.burst": "1.5.4",
+        "com.unity.test-framework": "1.1.22"
       },
-      "url": "https://packages.unity.com"
+      "url": "https://artifactory.prd.cds.internal.unity3d.com/artifactory/api/npm/upm-candidates"
```
SamuelBellomo, line 65: referencing private repos?

Packages/manifest.json
```diff
@@ -48,5 +52,19 @@
     "com.unity.modules.vr": "1.0.0",
     "com.unity.modules.wind": "1.0.0",
     "com.unity.modules.xr": "1.0.0"
-  }
+  },
+  "scopedRegistries": [
```
SamuelBellomo, line 56: should this still be here now that UTP is merged?

SamuelBellomo:
I added a few comments that didn't touch your work, not sure who to talk to service side for this, this API is weird.
Also, I'm noticing you have private repos in PackageManagerSetting.asset and other manifest.

Assets/BossRoom/Scripts/Shared/Net/UnityRelayUtilities.cs
andrews-unity: Because the relay SDK is not Unity Specific

Assets/BossRoom/Scripts/Client/Net/ClientGameNetPortal.cs
andrews-unity: This is no different than setting an IP Address for a Transport? As that is the connection data required you set it and then we use it. This is the only entry point for the Relay in the adapter/transport

Assets/BossRoom/Scripts/Shared/Net/GameNetPortal.cs
andrews-unity: So you can set the ip and port on the transport

Assets/BossRoom/Scripts/Shared/Net/GameNetPortal.cs
andrews-unity: It happens to be the max number of players the relay supports. Which is not really documented on the relay side :P

Assets/BossRoom/Scripts/Shared/Net/UnityRelayUtilities.cs
andrews-unity: They auto generate this API so its a tool being used... so that is how the tool generated the code for this API. So no spelling error from the point of their tool.

Assets/BossRoom/Scripts/Client/Net/ClientGameNetPortal.cs
andrews-unity: I mean you have to be signed in order to use relay ?

Assets/BossRoom/Scripts/Shared/Net/UnityRelayUtilities.cs
andrews-unity: This API has been updated to use a newer API that Relay has pushed.

Assets/BossRoom/Scripts/Shared/Net/UnityRelayUtilities.cs
andrews-unity: This has been updated to use the newer simple API (which was not around in pre.3)

Packages/packages-lock.json
andrews-unity: fixed

Packages/manifest.json
andrews-unity: fixed

andrews-unity:
> > I added a few comments that didn't touch your work, not sure who to talk to service side for this, this API is weird.
> 
> Yes, you are right, 99% of the relay code has been extracted from the Relay sample or the UTP Transport adapter. We can inform @bgagnon and @fjgal. Although, i would still think that some of these comments could be addressed.

I have updated the API to using the new simple API which is much cleaner and less weird! With that said we intentionally didn't want Relay stuff in the Adapter because while not implemented in this sample we didn't want it to need to the relay SDK and there is point where users will need to select their region and if they want to use DTLS or not which we (Relay and Server) didn't feel that should be in the adapter.  So the only entry point is setting the connection data.

Assets/BossRoom/Scripts/Client/Net/ClientGameNetPortal.cs
SamuelBellomo: what I mean is Boss Room has a full flow: when you finish the boss or you die, you're brought back to the main menu where you can launch a new game.
With `if (!AuthenticationService.Instance.IsSignedIn)`, that means if you're already signed in, you can't Join, it'll just skip this whole code block.

Assets/BossRoom/Scripts/Shared/Net/UnityRelayUtilities.cs
SamuelBellomo: ah I see, makes sense for Join
for allocation, it could make sense to have something does both "create allocation" and "get join code"? But that's mostly nitpicking at this point, now that the API is simpler, it seems fine

Assets/BossRoom/Scripts/Client/Net/ClientGameNetPortal.cs
SamuelBellomo: I was saying this since we already ask to join earlier with JoinRelayServerFromJoinCode, it seemed like it should have been just one call from a user's perspective.
But I'm guessing if the goal is to have relay be usable standalone, then having that dependency with UTP would break this?
If that's the case, I'll resolve that comment

Assets/BossRoom/Scripts/Shared/Net/GameNetPortal.cs
SamuelBellomo: We have an IP mode in Boss Room, this commented code should actually be something if we want to use UTP as our default transport, no?

Assets/BossRoom/Scripts/Shared/Net/GameNetPortal.cs
SamuelBellomo: ok. one more reason to document it here. We could have a const defined for this.

Assets/BossRoom/Scripts/Shared/Net/UnityRelayUtilities.cs
SamuelBellomo: much cleaner, thank you

Assets/BossRoom/Scripts/Shared/Net/GameNetPortal.cs
VALERE91Unity: I agree with @SamuelBellomo here. We need to still be able to specify the Server address and port. (I can see that they are now private in the adapter) how to support LAN scenarios otherwise? For Relay you need internet.

Assets/BossRoom/Scripts/Client/Net/ClientGameNetPortal.cs
Cosmin-B: This function is not used at all

Assets/BossRoom/Scripts/Client/Net/ClientGameNetPortal.cs
Cosmin-B: That code is not used. Hence, I removed it

Assets/BossRoom/Scripts/Shared/Net/GameNetPortal.cs
Cosmin-B: Yes, we will do this change, but in a following PR, as we first need to update the Adapter, then merge that into Netcode GO develop, then upgrade Boss Room and only after that we can use these fields

Assets/BossRoom/Scripts/Shared/Net/NetworkVariableGUID.cs
Cosmin-B: It is now!

Assets/BossRoom/Scripts/Shared/Net/GameNetPortal.cs
VALERE91Unity: Works for me

VALERE91Unity (approved):
Could we create an issue concerning the join through IP and port for UTP? So we don't forget that we cannot anymore and we need to fix it ASAP?

SamuelBellomo:
> Could we create an issue concerning the join through IP and port for UTP? So we don't forget that we cannot anymore and we need to fix it ASAP?

Created https://jira.unity3d.com/browse/MTT-1238
Thanks for making us think about it :)

Assets/BossRoom/Scripts/Client/UI/MainMenuUI.cs
```diff
@@ -18,6 +19,25 @@ public class MainMenuUI : MonoBehaviour
 
         private Client.ClientGameNetPortal m_ClientNetPortal;
 
+        private static MainMenuUI s_Instance = null;
+
+        /// <summary>
+        /// Small singleton getter, for easy access across other classes, this should be safe,
+        /// as it is not touching any aspect of networking
+        /// </summary>
+        public static MainMenuUI Instance
```
SamuelBellomo, line 28: oooh that'll be useful

Assets/BossRoom/Scripts/Client/Net/ClientGameNetPortal.cs
```diff
@@ -211,6 +219,51 @@ public static bool StartClientRelayMode(GameNetPortal portal, string roomKey, ou
...
+                    await UnityServices.InitializeAsync();
+                    Debug.Log(AuthenticationService.Instance);
+                    if (!AuthenticationService.Instance.IsSignedIn)
+                    {
+                        await AuthenticationService.Instance.SignInAnonymouslyAsync();
+                        var playerId = AuthenticationService.Instance.PlayerId;
+                        Debug.Log(playerId);
+                    }
+
+                    try
```
SamuelBellomo, line 242: I'd extend the try catch around the signin too, I'm guessing we can have signin failures too, in addition to joining

Assets/BossRoom/Scripts/Client/Net/ClientGameNetPortal.cs
```diff
@@ -211,6 +219,51 @@ public static bool StartClientRelayMode(GameNetPortal portal, string roomKey, ou
...
+
+            var chosenTransport  = NetworkManager.Singleton.gameObject.GetComponent<TransportPicker>().UnityRelayTransport;
+            NetworkManager.Singleton.NetworkConfig.NetworkTransport = chosenTransport;
+
+            switch (chosenTransport)
+            {
+                case UnityTransport utp:
+                    Debug.Log($"Setting Unity Relay client with join code {joinCode}");
+
+                    await UnityServices.InitializeAsync();
```
SamuelBellomo, line 233: You mention in another comment this isn't used, not sure I understand. If it's not useful, why call it?
If it is, is there a "if (!IsInitialized)" we could do, just like you did for signin?
It just feels wrong to initialize UnityServices each time we click the "Connect" button. Maybe it's an idempotent call and I'm worrying for nothing?

Cosmin-B: The code that wasn't used was removed. And for the UnitySerivces.Initialize is fine to be called multiple times at runtime

SamuelBellomo (approved):
🚀

VALERE91Unity (approved):
Ship it!!!!

PR #367, opened 2021-09-14
feat: server-authoritative character NetworkAnimator

fernando-cortez:
Jira task [here](https://jira.unity3d.com/browse/MTT-993).

Here, server-authoritative `NetworkTransform` is added to the Characters (players, imps, imp boss). With this change the player hierarchy looks like so:

PlayerAvatar (`NetworkObject`)
- PhysicsObject (`NetworkTransform`)
- PlayerGraphics (`Animator` & `NetworkAnimator`)

Animations are still played on client FX scripts, but only on the host player. This change then gets replicated to other clients through `NetworkTransform`.
Anticipated actions can still played locally through an override.

### DIFF
--- a/Assets/BossRoom/Scripts/Client/Game/Action/AnimationOnlyActionFX.cs
+++ b/Assets/BossRoom/Scripts/Client/Game/Action/AnimationOnlyActionFX.cs
@@ -10,7 +10,7 @@ namespace BossRoom.Visual
 
         public override bool Start()
         {
-            if( !Anticipated )
+            if (!Anticipated)
             {
                 PlayStartAnim();
             }
@@ -19,15 +19,15 @@ namespace BossRoom.Visual
             return true;
         }
 
-        private void PlayStartAnim()
+        private void PlayStartAnim(bool anticipated = false)
         {
-            m_Parent.OurAnimator.SetTrigger(Description.Anim);
+            m_Parent.TrySetTrigger(Description.Anim, anticipated);
         }
 
         public override void AnticipateAction()
         {
             base.AnticipateAction();
-            PlayStartAnim();
+            PlayStartAnim(true);
         }
 
         public override bool Update()
@@ -39,7 +39,7 @@ namespace BossRoom.Visual
         {
             if (!string.IsNullOrEmpty(Description.Anim2))
             {
-                m_Parent.OurAnimator.SetTrigger(Description.Anim2);
+                m_Parent.TrySetTrigger(Description.Anim2);
             }
         }
 
--- a/Assets/BossRoom/Scripts/Client/Game/Action/AoeActionFX.cs
+++ b/Assets/BossRoom/Scripts/Client/Game/Action/AoeActionFX.cs
@@ -12,7 +12,7 @@ namespace BossRoom.Visual
         public override bool Start()
         {
             base.Start();
-            m_Parent.OurAnimator.SetTrigger(Description.Anim);
+            m_Parent.TrySetTrigger(Description.Anim);
             GameObject.Instantiate(Description.Spawns[0], m_Data.Position, Quaternion.identity);
             return ActionConclusion.Stop;
         }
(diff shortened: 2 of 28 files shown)


Assets/BossRoom/Scripts/Client/Game/Action/AnimationOnlyActionFX.cs
```diff
@@ -10,7 +10,7 @@ public class AnimationOnlyActionFX : ActionFX
 
         public override bool Start()
         {
-            if( !Anticipated )
+            if (!Anticipated)
```
SamuelBellomo, line 13: mmm I would have hoped that this hole set of "Animation only" classes would disappear with networkAnimator :(

Assets/BossRoom/Scripts/Client/Game/Action/AoeActionFX.cs
```diff
@@ -12,7 +12,7 @@ public AoeActionFX(ref ActionRequestData data, ClientCharacterVisualization pare
         public override bool Start()
         {
             base.Start();
-            m_Parent.OurAnimator.SetTrigger(Description.Anim);
+            m_Parent.TrySetTrigger(Description.Anim);
```
SamuelBellomo, line 15: why are we still trying to setTrigger client side? I see this in multiple files in this PR.

Assets/BossRoom/Scripts/Client/Game/Character/ClientCharacterVisualization.cs
```diff
@@ -64,6 +64,8 @@ public class ClientCharacterVisualization : MonoBehaviour
 
         public ulong NetworkObjectId => m_NetState.NetworkObjectId;
 
+        static bool IsServer => NetworkManager.Singleton.LocalClientId == NetworkManager.Singleton.ServerClientId;
```
SamuelBellomo, line 67: NetworkManager has an "IsServer" field

Assets/BossRoom/Scripts/Client/Game/Action/FXProjectileTargetedActionFX.cs
```diff
@@ -92,7 +92,7 @@ private void PlayHitReact()
             if (m_Target && m_Target.TryGetComponent(out Client.ClientCharacter clientCharacter) && clientCharacter.ChildVizObject != null )
             {
                 var hitReact = !string.IsNullOrEmpty(Description.ReactAnim) ? Description.ReactAnim : k_DefaultHitReact;
-                clientCharacter.ChildVizObject.OurAnimator.SetTrigger(hitReact);
+                clientCharacter.ChildVizObject.TrySetTrigger(hitReact);
```
SamuelBellomo, line 95: oh, I just saw the TrySetTrigger implementation. So this isn't just client code, it's also authoritative code, correct?
So this works because we have a host that's both a client and a server?
I would have expected the set trigger stuff to happen in the "server" part of this action, no? Since we're server authoritative, I would expect the information sending to happen from server code.

fernando-cortez: So yes that's how this is working. It was the "safer" implementation. What you're suggesting would mean moving some of the client-side animation code to the server-side Actions, which I agree is the place where animations should play, since we're establishing server authority.

I took one more peek at when animations play on clients and they're for the most part triggered when ActionFX's start, end, or hit something. So it seems like it can be converted to server-side code.

One last thing to point out is that we'll still have some actions that play anticipated animations, so we'll still have a few animations playing on client-side ActionFXs.

SamuelBellomo: for animation anticipation, discussed this with Phil, there could be a way to inherit from NetworkAnimator and have a way to add some API for anticipation.
This could be done samples side

fernando-cortez: Following up on this specific item offline.

PR #371, opened 2021-09-27
Port to URP

jilfranco-unity:
Jira task [here](https://jira.unity3d.com/browse/MTT-1067)

Ported Boss Room from built-in render pipeline to universal render pipeline. This port includes

- rewriting shaders in ShaderGraph
- changing all materials over to those shaders
- editing lighting in the scenes
- editing some scene prefabs
- editing VFX


fernando-cortez (commented):
Looks great. Definitely a little darker than before but the moody lighting works for me.

2 things:
* It seems like the boss' charge vfx has not been converted yet. It's coming across as pink so it's probably just a quick material upgrade.
* The breakable crystals appear red/pink as a default. AFAIR this was the colour of their broken state, whereas they would spawn defaulted in a blue colour when unbroken.

SamuelBellomo:
As mentioned IRL, it'd be great to have "before and after" screenshots for the looks and of the profiler

pdeschain:
I was surprised how easy it turned out to port Roystan toon shader (one line change, yay!). I wonder if we would actually want to replace it with our own shader-graph implementation?

pdeschain:
Screen Shot 2021-09-28 at 1 09 01 PM

I wonder if we should enable some anti-aliasing (either here in URP settings or on the camera prefab)?

pdeschain:
Great work Jil, this PR looks great!

jilfranco-unity:
> Looks great. Definitely a little darker than before but the moody lighting works for me.
> 
> 2 things:
> 
> * It seems like the boss' charge vfx has not been converted yet. It's coming across as pink so it's probably just a quick material upgrade.
> * The breakable crystals appear red/pink as a default. AFAIR this was the colour of their broken state, whereas they would spawn defaulted in a blue colour when unbroken.

I'll look into the boss and crystal issues! Also I'd like to do another lighting pass because I agree, it is a bit dark. I figured this could be a separate change though!

jilfranco-unity:
> I was surprised how easy it turned out to port Roystan toon shader (one line change, yay!). I wonder if we would actually want to replace it with our own shader-graph implementation?

I was glad it was an easy change as well! I think I could recreate this shader in Shader Graph, but I'd have to dig into more. I had this as the lowest priority for converting to Shader Graph since it was hand written and pretty well commented. Maybe having both would be cool? 

I am finding a bug in builds though, the first time through character select, the lighting on the toon shader breaks and looks like this:
image

But if you quit (or the imps absolutely wreck you, whichever comes first) and the char select scene is loaded again, it works just fine:
image

I've found a work around to force it to work: adding a realtime reflection probe to the scene that encompasses the whole level, this issue doesn't occur. I hope to find a better solution than that for the long term

Also I agree it would be a good move to add anti-aliasing!

fernando-cortez:
> > Looks great. Definitely a little darker than before but the moody lighting works for me.
> > 2 things:
> > 
> > * It seems like the boss' charge vfx has not been converted yet. It's coming across as pink so it's probably just a quick material upgrade.
> > * The breakable crystals appear red/pink as a default. AFAIR this was the colour of their broken state, whereas they would spawn defaulted in a blue colour when unbroken.
> 
> I'll look into the boss and crystal issues! Also I'd like to do another lighting pass because I agree, it is a bit dark. I figured this could be a separate change though!

I retract what I said about the crystals. I think they're actually fine! Worth rolling back to a way old version to verify though.

jilfranco-unity:
> > > Looks great. Definitely a little darker than before but the moody lighting works for me.
> > > 2 things:
> > > 
> > > * It seems like the boss' charge vfx has not been converted yet. It's coming across as pink so it's probably just a quick material upgrade.
> > > * The breakable crystals appear red/pink as a default. AFAIR this was the colour of their broken state, whereas they would spawn defaulted in a blue colour when unbroken.
> > 
> > 
> > I'll look into the boss and crystal issues! Also I'd like to do another lighting pass because I agree, it is a bit dark. I figured this could be a separate change though!
> 
> I retract what I said about the crystals. I think they're actually fine! Worth rolling back to a way old version to verify though.

Gotcha, that matches with what I saw when comparing to the pre URP port project! :) They're just a lot less bloom-y now, so their pink color shows up better!

SamuelBellomo:
For performance, were those profiler screenshots taken from actual builds or in editor? I wouldn't rely on frame time from the editor

jilfranco-unity:
> There's a few things that aren't as great with the new screenshots. Floor to wall edge doesn't blend as well button seems completely black? Top of walls seem weird There's no more big shadow circle under the flames? This seems over saturated to me, the older version seemed more coherent and blending together. It feels like we're losing some details because of too dark areas

I agree the lighting itself could use some refining! I can brighten everything up a little, as well as down the saturation to match the pre URP look more closely. The button being much darker would be helped by this, as well as some material adjustments (probably is just too metallic). The ambient occlusion could also be turned down a touch!

Unfortunately, realtime shadows aren't supported by Point lights in URP--we could try baking them instead to get the shadow back!

And could you elaborate on "Tops of walls seem weird"? I'm not sure what you mean :P

jilfranco-unity:
> For performance, were those profiler screenshots taken from actual builds or in editor? I wouldn't rely on frame time from the editor

I used the separate process profiler to profile a dev build of the game for those screenshots. From my main post:
>Render thread time decreased in URP by about half, but overall draw time remains about the same at 12-13 ms in conditions like the screen caps above. Captures in Render Doc of a regular non dev build show frame times of around 6-7ms when just hanging out with no imps, and ~10-12ms when a ton of imps and effects are on screen

pdeschain:
@SteveTheodore looks great! 
Maybe we should also consider adding blob shadows under the characters (though it might look weird with the few realtime lights that we have) - otherwise the imps and players look a bit disconnected from the ground.

jilfranco-unity:
Thanks for your feedback @SteveTheodore! :) I'm going to try out all of these in my next PRs! I've merged the latest version of develop in as well

pdeschain (approved):

PR #372, opened 2021-09-29
feat: NetworkObject parenting, PickUp Action [MTT-992]

fernando-cortez:
### Description
This PR introduces the "PickUp" and "Drop" ability for all players. The action is triggered through contextual targeting of "Heavy" pots, and goes through the Action system. A pot has been added to the start of the entrance area. Pots parented to players that disconnect will be unparented through Netcode.

A pot is parented to the root of the PlayerAvatar GameObject, and the `PositionConstraint` component is used to follow the position of a hand bone.

As this is the first, non-visual iteration of the action, a follow-up PR will include the final UI images for the actions, as well as any supplementary visuals for pickup and dropping.

### Issue Number(s)
[MTT-992](https://jira.unity3d.com/browse/MTT-992)

### Contribution checklist


Assets/BossRoom/Scripts/Server/Game/Entity/ServerRemoveChildrenOnFaint.cs
```diff
@@ -0,0 +1,48 @@
+using Unity.Netcode;
+using UnityEngine;
+
+namespace Unity.Multiplayer.Samples.BossRoom.Server
+{
+    public class ServerRemoveChildrenOnFaint : NetworkBehaviour
```
SamuelBellomo, line 6: is this the workaround we were talking about? might be worth talking about the workaround in comments

Assets/BossRoom/Scripts/Client/Game/Action/PickUpActionFX.cs
```diff
@@ -0,0 +1,53 @@
+using UnityEngine;
+
+namespace Unity.Multiplayer.Samples.BossRoom.Visual
+{
+    /// <summary>
+    /// Complimentary FX class for the Pickup Action class. Plays a "Pickup" or "Drop" animation.
+    /// </summary>
+    public class PickUpActionFX : ActionFX
```
SamuelBellomo, line 8: any reason why we decided to still go with the action pattern instead of the 2 class pattern we discussed a few weeks ago?

Assets/BossRoom/Scripts/Server/Game/Entity/ServerDisplacerOnParentChange.cs
```diff
@@ -0,0 +1,71 @@
...
+using Unity.Netcode;
+using UnityEngine;
+
+namespace Unity.Multiplayer.Samples.BossRoom.Server
+{
+    /// <summary>
+    /// Component to simply play a raising/descending animation when this NetworkObject's parent NetworkObject changes.
+    /// </summary>
+    /// <remarks>
+    /// This script will likely be refactored and/or removed removed with a NetworkAnimator-based solution.
```
SamuelBellomo, line 11: current netcode version used by BR already has a NetworkAnimator, anything preventing us from using it? I thought it was just going to be a "drop on my object and it works", would that be different with newer versions?

Assets/BossRoom/Scripts/Shared/Game/Entity/CustomParentingHandler.cs
```diff
@@ -0,0 +1,158 @@
...
+            }
+        }
+
+        void AddChildNetworkObject(ulong networkObjectId)
+        {
+            if (NetworkManager.SpawnManager.SpawnedObjects.ContainsKey(networkObjectId))
+            {
+                if (!OnTrySetParent(NetworkManager.SpawnManager.SpawnedObjects[networkObjectId]))
+                {
+                    //Error message
```
SamuelBellomo, line 65: ?
empty if?

Assets/BossRoom/Scripts/Server/Game/Entity/ServerRemoveChildrenOnFaint.cs
fernando-cortez: No, this is purely gameplay-related behaviour. Once a player dies, they should drop their parented NetworkObject.

Assets/BossRoom/Scripts/Client/Game/Action/PickUpActionFX.cs
fernando-cortez: My thinking was that having two different Action systems would not be too easy to follow along through code. The two class pattern is a viable route, but there's other systems at play that would need to be accounted for (input, cross-play with other actions such as targeting, visuals, etc).

Assets/BossRoom/Scripts/Server/Game/Entity/ServerDisplacerOnParentChange.cs
fernando-cortez: So initially I gave this a go with NetworkAnimator. The thing is, the animation would be applying root motion to the root pot object, which isn't supported AFAIK.

VALERE91Unity (commented):
If I understand correctly, CustomParentingHandler is meant to disappear right? Otherwise I would recommend implement error logs on all your if.

Assets/BossRoom/Scripts/Shared/Game/Entity/CustomParentingHandler.cs
fernando-cortez: @VALERE91Unity yes, this class will ideally be removed once SDK allows for parenting a NetworkObject to a child of another NetworkObject. Error messages have been added.

VALERE91Unity (approved):

Assets/BossRoom/Scripts/Server/Game/Entity/ServerDisplacerOnParentChange.cs
SamuelBellomo: I see. It might be worth it to add a note in the comment about root motion not being supported by NA

Assets/BossRoom/Scripts/Shared/Game/Entity/CustomParentingHandler.cs
SamuelBellomo: @VALERE91Unity will it be fixed for this release? if yes, let's not add a workaround we know will disappear very soon.
if not for this release, then we don't have a choice to implement the workaround

Assets/BossRoom/Scripts/Server/Game/Entity/ServerDisplacerOnParentChange.cs
fernando-cortez: Updated remarks to elaborate on the "why".

SamuelBellomo:
I've put this on hold for now and added something on our calendars to come back to this.

SamuelBellomo:
After coming back to this, we'll need a refactor to our player structure before this gets in

Assets/BossRoom/Scripts/Client/Game/Character/ClientInputSender.cs
```diff
@@ -414,6 +415,11 @@ void Update()
                 RequestAction(ActionType.Emote4, SkillTriggerStyle.Keyboard);
             }
 
+            if (Input.GetKeyDown(KeyCode.Space))
```
SamuelBellomo, line 418: Is this key documented in the initial instruction popup?

Assets/BossRoom/Scripts/Server/Game/Entity/ServerDisplacerOnParentChange.cs
```diff
@@ -0,0 +1,74 @@
...
+using UnityEngine;
+
+namespace Unity.Multiplayer.Samples.BossRoom.Server
+{
+    /// <summary>
+    /// Component to simply play a raising/descending animation when this NetworkObject's parent NetworkObject changes.
+    /// </summary>
+    /// <remarks>
+    /// Currently, Netcode for GameObjects' (Netcode) NetworkAnimator component does not support animations that apply
+    /// Root Motion. This script is a workaround and will be refactored when Root Motion-based animations are supported.
```
SamuelBellomo, line 13:
```suggestion
    /// Root Motion. This script is a workaround and should be refactored if Root Motion-based animations are supported.
```

Assets/BossRoom/Scripts/Server/Game/Entity/ServerDisplacerOnParentChange.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// </remarks>
+    public class ServerDisplacerOnParentChange : NetworkBehaviour
```
SamuelBellomo, line 15: So this class is basically to lerp the object while the animation is running. I'm wondering if there's anything in NetworkTransform's interpolation that's messing with this.

Assets/BossRoom/Scripts/Server/Game/Entity/ServerDisplacerOnParentChange.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            if (parentNetworkObject)
+            {
+                StopAllCoroutines();
+                StartCoroutine(SmoothPositionLerpY(k_PickUpAnimationLength, 1f, true));
+            }
+            else
+            {
+                StopAllCoroutines();
+
```
SamuelBellomo, line 39:
```suggestion
```

Assets/BossRoom/Scripts/Server/Game/Entity/ServerDisplacerOnParentChange.cs
```diff
@@ -0,0 +1,74 @@
...
+                m_NetworkTransform.enabled = true;
+
```
SamuelBellomo, line 41:
```suggestion
```

Assets/BossRoom/Scripts/Server/Game/Entity/ServerDisplacerOnParentChange.cs
```diff
@@ -0,0 +1,74 @@
...
+            var start = local ? transform.localPosition.y : transform.position.y;
+
+            var progress = 0f;
+            var duration = 0f;
+
+            while (progress < 1f)
+            {
+                duration += Time.deltaTime;
+                progress = Mathf.Clamp(duration / length, 0f, 1f);
+                var progressY = Mathf.Lerp(start, targetHeight, progress);
```
SamuelBellomo, line 57: As discussed, let's try to push limits and really have the pot follow the character's hands

fernando-cortez:
State of PR: awaiting [fix](https://github.com/Unity-Technologies/com.unity.netcode.gameobjects/issues/1523) from SDK that was noticed when working on `NetworkObject` parenting.

Parenting should work with a hierarchy of a spawned `NetworkObject` containing a `NetworkObject` child GameObject. PR on hold until fix is in.

SamuelBellomo:

Assets/BossRoom/Scripts/Game/Server/Action/PickUpAction.cs
```diff
@@ -0,0 +1,123 @@
...
+            // found a suitable collider; make sure it is not already held by another player
+            if (heavyNetworkObject.transform.parent != null &&
+                heavyNetworkObject.transform.parent.TryGetComponent(out NetworkObject parentNetworkObject))
+            {
+                // pot already parented; return for now
+                return false;
+            }
+
+            // found a suitable collider; try to child this NetworkObject
+            if (!heavyNetworkObject.TrySetParent(m_Parent.transform))
```
SamuelBellomo, line 72: If you add artificial jitter, do you see issues where the pot doesn't quite follow the character's hands when walking and moving it?
A solution for that is to use local positions (that's an option you can set in NetworkTransform)

Assets/BossRoom/Scripts/Game/Server/Entity/SpawnNetworkObjectAtLocation.cs
```diff
@@ -0,0 +1,36 @@
+using System.Collections.Generic;
+using Unity.Netcode;
+using UnityEngine;
+using UnityEngine.SceneManagement;
+
+namespace Unity.Multiplayer.Samples.BossRoom.Game
+{
+    public class SpawnNetworkObjectAtLocation : NetworkBehaviour
```
SamuelBellomo, line 8: I feel like I've seen this before. For portals and previously for when we didn't have statically placed objects available.
Did we look into consolidating this (if possible of course)?

Assets/BossRoom/Scripts/Game/Shared/Entity/NetworkPickUpState.cs
```diff
@@ -0,0 +1,14 @@
+using Unity.Netcode;
+
+namespace Unity.Multiplayer.Samples.BossRoom
+{
+    /// <summary>
+    /// Shared Network logic for targetable, NPC, pickup objects.
+    /// </summary>
+    public class NetworkPickUpState : NetworkBehaviour, ITargetable
```
SamuelBellomo, line 8: We need a NetworkBehaviour for this?

Assets/BossRoom/Scripts/Game/Client/Character/ClientInputSender.cs
```diff
@@ -446,6 +447,11 @@ void Update()
                 RequestAction(ActionType.Emote4, SkillTriggerStyle.Keyboard);
             }
 
+            if (Input.GetKeyDown(KeyCode.Space))
```
SamuelBellomo, line 450: Is this documented in our popup tutorial?

Assets/BossRoom/Scripts/Game/Server/Entity/SpawnNetworkObjectAtLocation.cs
SamuelBellomo, line 8: Also, we can't have statically placed pickup pots?

Assets/BossRoom/Scripts/Game/Server/Entity/SpawnNetworkObjectAtLocation.cs
```diff
@@ -0,0 +1,36 @@
...
+        GameObject m_NetworkObjectPrefab;
+
+        public override void OnNetworkSpawn()
+        {
+            if (!IsServer)
+            {
+                return;
+            }
+
+            NetworkManager.SceneManager.OnLoadEventCompleted += SpawnNetworkPrefab;
```
LPLafontaineB, line 20: This gets executed every time a new additive scene is loaded as well, which means that if a character goes near the door, then back to the spawn, then back to the door, etc. there will be multiple objects being spawned at the same place. Was that the intended behaviour?

LPLafontaineB (commented):
Looks like the characters carry the objects differently. The archer and rogue seem to have it be over their head, while the tank and mage have it on their right hand. It might look better with it being placed between both hands in front of the body. The way it is constrained, the pot seems to move weirdly when getting hit or when using another action while carrying it. (should actions be disabled when carrying an object?)
image

Also, there is a pot near the first portal that has the same model but is not interactable.

And should we do something similar to what we did with the revive action, and have a button replace the attack action when we select a pot? That would allow mobile users to use that action too

Assets/BossRoom/Scripts/Game/Server/Entity/SpawnNetworkObjectAtLocation.cs
fernando-cortez: Right. I was doing most of my tests inside just the entrance bit and hadn't thought about the scene being reloaded multiple times. This does overlap with the bug that I've logged [here](https://jira.unity3d.com/browse/MTT-3615). We have a workaround now for parenting in-scene placed NetworkObjects, so if they're part of the scene through NGO, we won't have this issue. I'll add this as a comment on that Jira ticket.

fernando-cortez:
> Looks like the characters carry the objects differently. The archer and rogue seem to have it be over their head, while the tank and mage have it on their right hand. It might look better with it being placed between both hands in front of the body. The way it is constrained, the pot seems to move weirdly when getting hit or when using another action while carrying it. (should actions be disabled when carrying an object?) image
> 
> Also, there is a pot near the first portal that has the same model but is not interactable.
> 
> And should we do something similar to what we did with the revive action, and have a button replace the attack action when we select a pot? That would allow mobile users to use that action too

Very good points on the feel & usability of this action. I will look into a better bone solution (maybe creating a separate one on the graphics gameobjects). And of course clean up the other pots.

I will likely throw the action side of this PR first, and a subsequent one will actually add it to the main scene + address UX issues for usage.

fernando-cortez:
Moving to On Hold until [this](https://github.com/Unity-Technologies/com.unity.netcode.gameobjects/issues/2056) issue is addressed SDK side. 

IMO the action will feel broken if joining players don't see the hold animation on late-join. @SamuelBellomo for visibility.

SamuelBellomo:
@fernando-cortez what do you think about merging this on our side and bugging this so we keep track. This way that code base is now part of our dev branch, it can now be playtested and we can close that ticket, since anyway this isn't a samples issue

fernando-cortez:
@SamuelBellomo: merge conflicts have been resolved. The pots haven't been added to the level. There's still an animator issue on pickup which causes the player to rotate. I think we should do the same as the NetworkRigidbody PR, where @jilfranco-unity builds on top of this PR to address those animator issues, as well as add visual improvements on pickup/drop FX.

Assets/BossRoom/Scripts/Game/Server/Action/PickUpAction.cs
fernando-cortez: It's the same jitter that a player would experience since the pot is synced via NetworkTransform as well.

SamuelBellomo: As discussed, using local pos is also a good way to reduce bandwidth consumption (doc for this should be created)

Assets/BossRoom/Scripts/Game/Client/Character/ClientInputSender.cs
fernando-cortez: This can be done in the follow-up, visual enhancement PR.

SamuelBellomo: I don't see a PR, can you link the jira ticket or PR please? Just so this isn't lost.

Assets/Scripts/Gameplay/GameplayObjects/NetworkPickUpState.cs
```diff
@@ -0,0 +1,10 @@
+using Unity.Netcode;
```
SamuelBellomo, line 1: a single component just for this seems a bit excessive? Since pickup should only be doable by players, do we need this to be modular?

Assets/Scripts/Gameplay/GameplayObjects/NetworkPickUpState.cs
```diff
@@ -0,0 +1,10 @@
...
+using UnityEngine;
+
+namespace Unity.Multiplayer.Samples.BossRoom
+{
+    public class NetworkPickUpState : NetworkBehaviour
+    {
+        public NetworkVariable<ulong> heldObject;
```
SamuelBellomo, line 8: NetworkObjectReference could be used instead of ulong

Assets/Scripts/Gameplay/GameplayObjects/NetworkPickUpState.cs
```diff
@@ -0,0 +1,10 @@
+using Unity.Netcode;
+using UnityEngine;
+
+namespace Unity.Multiplayer.Samples.BossRoom
+{
+    public class NetworkPickUpState : NetworkBehaviour
```
SamuelBellomo, line 6: @pdeschain for state merges

Assets/Scripts/Gameplay/GameplayObjects/ServerDropItemOnFaint.cs
```diff
@@ -0,0 +1,43 @@
+using Unity.Netcode;
```
SamuelBellomo, line 1: Since players can carry only while the action is running, couldn't this logic be part of the action while it's running?

Assets/BossRoom/Scripts/Game/Server/Action/PickUpAction.cs
SamuelBellomo: Is there a ticket for doc writing on this? Or a PR? Just so we don't lose track of this

Assets/Scripts/Gameplay/GameplayObjects/ServerDropItemOnFaint.cs
fernando-cortez: Ideally, yes. However, at the moment, there isn't a way to stop an action of the same type to not be added to the action queue, besides playing with the `m_LastUsedTimestamps` which turns out to be quite messy. I'd be setting an unreachable resuse timeout on the action, and check for input inside this action class to drop the item, and it's still not guaranteed to fire after the input checking inside `ClientInputSender`. It would rework a bit of the underlying Action system, so I can definitely add it. But as of writing this, Actions are almost always just added to the action queue, meaning that this class has to do both the pickup and drop logic anytime it's fired.

Assets/Scripts/Gameplay/GameplayObjects/NetworkPickUpState.cs
fernando-cortez: I think here I preferred composition vs. adding this to the `ServerCharacter` and `NetworkCharacterState` god classes we have. It was trivial to just add it to the PlayerAvatar prefab. If we feel like imps/boss/vandal imp should pick up items, we'd just add it to the base Character prefab.

SamuelBellomo: God classes are anti pattern yes, but why does it apply in this case here? We need to be careful about not over engineering things. You have to ask yourself "Does setting this in a separate class make things more simple overall?" You usually want to start simple, then start adding abstractions and extracting things when you start seeing patterns. 
Wouldn't the action already have access to ServerCharacter?

fernando-cortez: Went back and added it to `NetworkCharacterState`. Fits the same pattern as `targetId`.

Assets/Scripts/Gameplay/GameplayObjects/NetworkPickUpState.cs
fernando-cortez: I should note that I've went back and reverted it back to a `NetworkVariable<ulong>` since `NetworkObjectReference` does not have a OnValueChanged callback.

Assets/BossRoom/Scripts/Game/Client/Character/ClientInputSender.cs
fernando-cortez: I've removed it as an action triggered through input. Instead it's contextual on the `targetId`. This enables this ability on mobile too. I've created a [task](https://jira.unity3d.com/browse/MTT-4304) to update the pickup action image with a proper one.

Assets/Scripts/Gameplay/GameplayObjects/NetworkCharacterState.cs
```diff
@@ -36,6 +36,8 @@ public class NetworkCharacterState : NetworkBehaviour, ITargetable
         /// Indicates how the character's movement should be depicted.
         public NetworkVariable<MovementStatus> MovementStatus { get; } = new NetworkVariable<MovementStatus>();
 
+        public NetworkVariable<ulong> heldNetworkObject;
```
SamuelBellomo, line 39: mmm reverted to netvar? This could still be a NetworkObjectReference, no?

Assets/Scripts/Gameplay/Action/PickUpAction.cs
```diff
@@ -0,0 +1,157 @@
+using System;
```
SamuelBellomo, line 1: Note to not forget, add changelog and check PR template for TODOs

Assets/Scripts/Gameplay/GameplayObjects/ServerDropItemOnFaint.cs
fernando-cortez: Update:
* Pickup and drop actions have been separated. This allowed for simplification of action flows and for the addition of drop icon.
* This class has since been removed since this is tracked inside of `PickupAction.cs`. Life state is checked on update to determine when to end action.

CHANGELOG.md
```diff
@@ -9,7 +9,8 @@ Additional documentation and release notes are available at [Multiplayer Documen
 ## [Unreleased] - yyyy-mm-dd
 
 ### Added
-* 
+* Pickup and Drop action added to the Action system. Actionable once targeting a "Heavy"-tagged NetworkObject. (#372)
```
SamuelBellomo, line 12:
```suggestion
* Pickup and Drop action added to the Action system. Actionable once targeting a "Heavy"-tagged NetworkObject. (#372) - This shows NetworkObject parenting with a pattern to follow animation bones (the hands when picking up)
```

SamuelBellomo: Let's try to have the edu value in the release note please :)

SamuelBellomo (dismissed):

CHANGELOG.md
fernando-cortez: Noted!

SamuelBellomo (approved):

Assets/BossRoom/Scripts/Game/Server/Action/PickUpAction.cs
fernando-cortez: Here: [MTT-4466](https://jira.unity3d.com/browse/MTT-4466).

LPLafontaineB (approved):

PR #375, opened 2021-10-13
Update to latest netcode

SamuelBellomo:
Updating to latest netcode.
Added support for UTP direct IP connection, but couldn't test due to error -4. Reverted back to UNET until we have a fix.
Cleaning GameNetPortal, making it clearer to understand and removing legacy code that was needed for previous versions.
OnDestroy now needs to be overriden (since NetworkBehaviour has its own OnDestroy)
Removing custom interpolation to use NetworkTransform's interpolation.
Fixed some jitter issue with UIName
Now using FastBufferWriter for custom serialization
Character select data now uses a NetworkList of LobbyPlayerState. That state was converted so it's a proper value type now.
Removing NetworkVariableLobbyState and NetworkVariableString.
Names now use FixedString32Byte instead of string. This way we can use those directly in netvars.
Using OnSceneEvent properly instead of custom callback in ServerBossRoomState

### DIFF
--- a/Assets/BossRoom/Scripts/Client/Game/Character/ClientCharacterVisualization.cs
+++ b/Assets/BossRoom/Scripts/Client/Game/Character/ClientCharacterVisualization.cs
@@ -241,7 +241,7 @@ namespace Unity.Multiplayer.Samples.BossRoom.Visual
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


SamuelBellomo:
This PR is based a week old Netcode version. There might be some more changes needed, however waiting on photon fix.

PR #382, opened 2021-10-19
Bug Fix: Renaming connection methods in main menu

jilfranco-unity:
Jira ticket [here](https://jira.unity3d.com/browse/MTT-1473)

In main menu, changed labels of connection types "RELAY" and "U-RELAY" to "PHOTON RELAY" and "UNITY RELAY" respectively

Before:
connectionTypeBefore

After:
connectionTypeAfter


jilfranco-unity:
Pushed that label size increase change!

PR #385, opened 2021-10-20
Fix: update ngo & utp package references

LPLafontaineB:
This PR updates the package references to point to the candidates registry


SamuelBellomo:
> It all looks okay, the only one that I am not sure about is the ParrelSync.

ParrelSync is staying at the same version though, what are you not sure about @Cosmin-B ?

PR #387, opened 2021-10-20
Bug Fix: Bossroom logo missing on mac builds

jilfranco-unity:
Jira ticket [here](https://jira.unity3d.com/browse/MTT-1457)

Fixed the issue where the Bossroom logo wasn't appearing on mac builds

image
Now it does! :)


VALERE91Unity:
> The solve was changing the material on the UI image component to None (which just sets it to the default Unity UI/Unlit material) rather than the Bossroom logo material. There's a place where you can set the image itself separate from the material texture.

Damn that's strange... When you get to difference in graphical pipelines between Metal and DirectX you know you are 48H before release 😛

PR #401, opened 2021-11-03
feat: client-authoritative character select

fernando-cortez:
### Description (*)
This PR introduces a more reactive flow for the local player's character select screen.

The local player's character choice is shown instantaneously, instead of previously waiting for a confirmation event from the server. This clears the local player's previous choice and sets the new seat visuals right away.

The server will still override changes for the local player when they are locked in/out, or set to inactive due to another player confirming the same player, and of course any other player's selection.

The confirm button is server-authoritative. I figure a visual update for that when waiting for server confirm can be done as a follow up to this PR.

### Issue Number(s) (*)

Fixes issue(s): [MTT-1438](https://jira.unity3d.com/browse/MTT-1438)
### Manual testing scenarios

1. ...
2. ...
### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Scripts/Client/Game/State/ClientCharSelectState.cs
+++ b/Assets/BossRoom/Scripts/Client/Game/State/ClientCharSelectState.cs
@@ -170,19 +170,14 @@ namespace Unity.Multiplayer.Samples.BossRoom.Client
         /// </summary>
         private void OnLobbyPlayerStateChanged(NetworkListEvent<CharSelectData.LobbyPlayerState> changeEvent)
         {
-            // ignore state changes for the local player, unless the change event is a locked in event (which is
-            // server-authoritative)
+            // ignore state changes for the local player unless the change event is a locked in event, or when seat
+            // has been reset to -1 (both server-authoritative)
             if (changeEvent.Value.ClientId == NetworkManager.Singleton.LocalClientId)
             {
-                if (changeEvent.Value.SeatState == CharSelectData.SeatState.LockedIn && !m_HasLocalPlayerLockedIn)
-                {
+                var isLockedInEvent = (changeEvent.Value.SeatState == CharSelectData.SeatState.LockedIn && !m_HasLocalPlayerLockedIn) ||
+                    (changeEvent.Value.SeatState == CharSelectData.SeatState.Active && m_HasLocalPlayerLockedIn);
 
-                }
-                else if (changeEvent.Value.SeatState == CharSelectData.SeatState.Active && m_HasLocalPlayerLockedIn)
-                {
-
-                }
-                else
+                if (!isLockedInEvent && changeEvent.Value.SeatIdx != -1)
                 {
                     return;
                 }
@@ -390,12 +385,15 @@ namespace Unity.Multiplayer.Samples.BossRoom.Client
         /// <summary>
         /// Called directly by UI elements!
         /// </summary>
+        /// <remarks>
+        /// For a reactive character select screen, the local player's selection is presented instantaneously
+        /// before a NetworkListEvent can be received from the server.
+        /// To accomplish this, the previous seat selected will be cleared or re-populated by a shared player, and
+        /// the new seat will be populated with the local selection.
+        /// </remarks>
         /// <param name="seatIdx"></param>
         public void OnPlayerClickedSeat(int seatIdx)
         {
-            // visually show the change immediately
-
-            // get current seat (m_LastSeatSelected)
             if (m_LastSeatSelected != -1)
             {
                 var otherPlayerSharesSeat = false;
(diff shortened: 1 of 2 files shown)


Assets/BossRoom/Scripts/Client/Game/State/ClientCharSelectState.cs
```diff
@@ -171,6 +170,19 @@ private void UpdatePlayerCount()
...
         private void OnLobbyPlayerStateChanged(NetworkListEvent<CharSelectData.LobbyPlayerState> changeEvent)
         {
+            // ignore state changes for the local player unless the change event is a locked in event, or when seat
+            // has been reset to -1 (both server-authoritative)
+            if (changeEvent.Value.ClientId == NetworkManager.Singleton.LocalClientId)
+            {
+                var isLockedInEvent = (changeEvent.Value.SeatState == CharSelectData.SeatState.LockedIn && !m_HasLocalPlayerLockedIn) ||
+                    (changeEvent.Value.SeatState == CharSelectData.SeatState.Active && m_HasLocalPlayerLockedIn);
+
+                if (!isLockedInEvent && changeEvent.Value.SeatIdx != -1)
```
SamuelBellomo, line 180: magic number, should be a const somewhere (also other -1 should use that const)

Assets/BossRoom/Scripts/Client/Game/State/ClientCharSelectState.cs
```diff
@@ -171,6 +170,19 @@ private void UpdatePlayerCount()
         /// </summary>
         private void OnLobbyPlayerStateChanged(NetworkListEvent<CharSelectData.LobbyPlayerState> changeEvent)
         {
+            // ignore state changes for the local player unless the change event is a locked in event, or when seat
+            // has been reset to -1 (both server-authoritative)
+            if (changeEvent.Value.ClientId == NetworkManager.Singleton.LocalClientId)
+            {
+                var isLockedInEvent = (changeEvent.Value.SeatState == CharSelectData.SeatState.LockedIn && !m_HasLocalPlayerLockedIn) ||
```
SamuelBellomo, line 177: why do we use local values for locked in? With server authority, we should only take the server value as the authoritative value and never rely on local values

Assets/BossRoom/Scripts/Client/Game/State/ClientCharSelectState.cs
```diff
@@ -373,9 +385,50 @@ private void ConfigureUIForLobbyMode(LobbyMode mode)
...
+        /// the new seat will be populated with the local selection.
+        /// </remarks>
         /// <param name="seatIdx"></param>
         public void OnPlayerClickedSeat(int seatIdx)
         {
+            if (m_LastSeatSelected != -1)
+            {
+                var otherPlayerSharesSeat = false;
+
+                // if any other user shares seat, set them as owner; else deactivate seat
```
SamuelBellomo, line 401: "owner"? I thought the lockedIn state was server driven?

Assets/BossRoom/Scripts/Client/Game/State/ClientCharSelectState.cs
```diff
@@ -373,9 +385,50 @@ private void ConfigureUIForLobbyMode(LobbyMode mode)
...
+                    {
+                        // ignore self;
+                        continue;
+                    }
+
+                    if (lobbyPlayer.SeatIdx == m_LastSeatSelected)
+                    {
+                        // populate this seat with a shared player's lobby player state
+                        otherPlayerSharesSeat = true;
+                        m_PlayerSeats[m_LastSeatSelected].SetState(lobbyPlayer.SeatState, lobbyPlayer.PlayerNum, lobbyPlayer.PlayerName);
```
SamuelBellomo, line 414: If we're to show this in doc, let's rename m_PlayerSeats. I had to go to definition to figure out this was UI elements only (since it has a "SetState" method, I thought this was a netvar of some sort)

SamuelBellomo: We should maybe add a comment above saying this is the client driven part, where we just update our UI state without waiting for server values.

SamuelBellomo: Might be worth to note too that we're doing a pass over all seats. This serves as a sort of "merge conflict resolution" where instead of just trying to apply deltas, we're resetting everything.

Assets/BossRoom/Scripts/Client/Game/State/ClientCharSelectState.cs
```diff
@@ -373,9 +385,50 @@ private void ConfigureUIForLobbyMode(LobbyMode mode)
...
+
+                if (!otherPlayerSharesSeat)
+                {
+                    // no other player shared this seat; it is safe to just disable
+                    m_PlayerSeats[m_LastSeatSelected].SetState(CharSelectData.SeatState.Inactive, -1, string.Empty);
+                }
+            }
+
+            // get local player, and populate the seat that is anticipated to be taken
+            TryGetLobbyPlayer(NetworkManager.Singleton.LocalClientId, out var localLobbyPlayerState);
```
SamuelBellomo, line 426: might be worth to note this is the client drive part where we take the seat without waiting for the server driven value

Assets/BossRoom/Scripts/Client/Game/State/ClientCharSelectState.cs
fernando-cortez: So on OnPlayerClickedSeat(..), that's where the local player makes their anticipated selection, but also sends a ServerRpc for that change to be applied by the server to the NetworkList, which is then replicated to other clients.

Since the local player will still receive that same change event, I need to identify if it's just a local character selection change which can be ignored. What shouldn't be ignored is seat status changes.

NetworkListEvent doesn't provide that granularity on exactly what value was modified, so I'd need to compare it against _something_ locally, no?

I could also be comparing it against the current seat from `m_PlayerSeats`, since that is what is actually modified before the ServerRpc is sent. But that's still also something stored locally.

Assets/BossRoom/Scripts/Client/Game/State/ClientCharSelectState.cs
fernando-cortez: When a player chooses a new seat, I need to clear the local player's previous seat selection from the UI. But sometimes that seat is shared by another player.

Whether that previous seat now becomes empty, or "held" by another player is determined by the last update that was received by the server on the seats.

Assets/BossRoom/Scripts/Client/Game/State/ClientCharSelectState.cs
SamuelBellomo: discussed IRL, should have previous value coming from SDK, need following up with Fatih

fernando-cortez:
State of PR: [fix](https://github.com/Unity-Technologies/com.unity.netcode.gameobjects/pull/1528) containing a `NetworkListEvent`'s previous value has been merged to Netcode's develop branch.

Will place on hold until Netcode SDK version has been upgraded.

fernando-cortez:
Moving this to blocked. Awaiting possible refactors to NetworkVariable implementation.

SamuelBellomo:
So coming back to this, an owner netvar solution might still be weird. After discussions with Luke, we'd still need a server driven network list to manage locked in state, with an additional owner netvar for selection. Syncing between the two could be weird and buggy. Owner netvars can't be corrected by the server, so we'd effectively need to sync the network list and the owner netvar's state, with all the latency in between. 
With client driven movements coming, not sure showing this example would still have value? Let's keep it until we have merged other client driven stuff.

SamuelBellomo:

SamuelBellomo:

PR #404, opened 2021-11-04
Adding utilities asmdef.

SamuelBellomo:
### Description
We shouldn't add any dependencies to boss room to this asmdef. asmdefs should depend on this, not the reverse.
We should move our systems here as a first step before sharing with future samples.

### Manual testing scenarios

Tested with second project, see if we could import it as package file in packman.
Tested dependencies to and from this package, works as expected.

### Contribution checklist


SamuelBellomo:
This is the basis for adding utility classes, so we can start moving some of our reusable code here. This way with future samples, we'll be able to add a file dependency to boss room and will be able to reuse that code.

VALERE91Unity (dismissed):

LPLafontaineB (approved):

pdeschain (approved):

PR #408, opened 2021-11-08
Visual Adjustments for Boss Room

jilfranco-unity:
### Description (*)
Visual updates for Boss Room!

Updates include:

- A matte around the border of the map to block the camera from seeing lava, edges of meshes, particles, etc
- Lighting changes and optimizations 
-- Reduced number of realtime lights to 4 (from 27 (oof))
-- Baked 3x1024 and 1x128 lightmaps (about 4MB), these are for area lights to make the lava glow more convincing
- Added little blob shadows to characters (a plane with a transparent soft black circle attached to the physics object in the character prefab)
- Changed the lights for the enemy spawners from point lights to spot lights (still not shadow casting)
- Rotated all of the little pots to be in different directions so they don't look so copy-pasted
- Tightened radii of the torch lights and reduced the amount of realtime lights for each torch from 2 to 1
-- Also added a couple of extra torches
- Upped the contrast and adjusted the ambient lighting of the scene to be a more saturated blue

### Related Pull Requests
Some changes are from notes from [this PR](https://github.com/Unity-Technologies/com.unity.multiplayer.samples.coop/pull/371)

And now a bunch of before/after screenshots!

Matte to block stuff under level:
image
image

Lava glow:
image
image

Enemy spawners:
image

image

Shadows for imps:
image
image

Contrast and ambient lighting adjustment:
image
image


jilfranco-unity:
Oh I just noticed that in these screen shots the torch flames don't show up--they're still there, but I turned off the simulation of particle systems while I was working because I kept clicking on them :P

chrispope:
> Oh I just noticed that in these screen shots the torch flames don't show up--they're still there, but I turned off the simulation of particle systems while I was working because I kept clicking on them by accident :P

Can you add them in and share a screenshot, I would like to see it.

jilfranco-unity:
> Can you add them in and share a screenshot, I would like to see it.

Of course! Here's a few:

image
image
image

fernando-cortez (commented):
Followed up offline and suggested the following changes:
- Removing shadow casting on matte & removing physics colliders
- Look into repositioning of lava from way below level
- Look into reducing particle count on flame torches & volume max range on audiosource component

Also, StreamingAssets meta folder can be removed.

jilfranco-unity:
Updates to this PR

* Removed some unnecessary items from baked global illumination, this brought bake down to 2x1024 maps at about 2.7 MB, and bake time to 10 seconds (from ~25)

* Redid the matte as a single mesh using ProBuilder. It no longer casts shadows, has a collider, or receives any kind of lighting (and is ~250 verts)

* Adjusted overall scene color to be more cyan than purple

* Reduced intensity of the imp door lights 

* Reduction of about 1ms frame time in the area with baked lighting (~6.5 before, ~5.5 after on a dev build)

* Moved lava up in Y a bit to make the lava pools feel fuller

* Turned off some import features on some meshes (like importing lights, importing cameras, that kind of thing)

* Reduced audio radius of the torch flames from 500 (which is basically the whole level) to 10 and capped max particles to 5 or 10, down from 1000

image
image

chrispope (dismissed):
Lgtm - very cool update

fernando-cortez (commented):
image

Noticing the missing reference on the new border GameObject. Is this a ProBuilder mesh component? Seems to be working without it though, so I'm curious where the mesh for this object lies.

Nitpicking once more, but if it can be positioned at (0,0,0), then great.

pdeschain (approved):

pdeschain:
I love how our fire particles now have 5 max particles vs 1000 👍

chrispope:
> I love how our fire particles now have 5 max particles vs 1000 👍

We will miss those additional 995 max particles

jilfranco-unity:
> image
> 
> Noticing the missing reference on the new border GameObject. Is this a ProBuilder mesh component? Seems to be working without it though, so I'm curious where the mesh for this object lies.
> 
> Nitpicking once more, but if it can be positioned at (0,0,0), then great.

I've moved it to 0,0,0! Also, that missing script reference was the probuilder mesh stuff, so I exported the probuilder mesh and brought it in as a proper mesh asset, so now it just uses that and doesn't need the probuilder mesh stuff :)

fernando-cortez (approved):

chrispope (approved):

PR #411, opened 2021-11-11
feat: player hierarchy QoL updates, anim event listener added for player anim events

fernando-cortez:
### Description (*)
This PR reconfigures the hierarchy for the Character prefab, the base for Imp, Imp Boss, & PlayerAvatar. What was once moved over to a child GameObject (PhysicsObject) is now moved back to the root GameObject.

What the hierarchy looks like in-game is:
PlayerAvatar (NetworkObject, NetworkTransform)
-PlayerGraphics (NetworkAnimator)

An added QoL fix here is that anim-based events now are invoked for the player. The PlayerGraphics Animator did not have a component to listen for anim events, since `ClientCharacterVisualization` component is spawned dynamically and childed to PlayerGraphics (important since the component is not in the same hierarchical level as the animator). To hook up animations firing from the animator, a `ClientCharacterVisualization` component has been re-added to PlayerGraphics.

Bonus fix: the boss' despawn time has been reverted back to the proper value of -1. It has been noted that the boss despawns before the game ends. This resolves that issue.

### Issue Number(s) (*)

Fixes issue(s): [MTT-1511](https://jira.unity3d.com/browse/MTT-1511)
### Manual testing scenarios

1. ...
2. ...
### Contribution checklist


LPLafontaineB:
There seems to be an animation issue with this PR. Some abilities seem to have the wrong animation now (i.e. the archer uses the melee attack animation while shooting or using their AoE, as does the mage when attacking)

fernando-cortez:
> There seems to be an animation issue with this PR. Some abilities seem to have the wrong animation now (i.e. the archer uses the melee attack animation while shooting or using their AoE, as does the mage when attacking)

Will look into that!

LPLafontaineB:
> 
> 
> > There seems to be an animation issue with this PR. Some abilities seem to have the wrong animation now (i.e. the archer uses the melee attack animation while shooting or using their AoE, as does the mage when attacking)
> 
> Will look into that!

Could not reproduce the issue anymore

LPLafontaineB (approved):

Assets/BossRoom/Scripts/Client/AnimationCallbacks/ClientAnimListener.cs
```diff
@@ -0,0 +1,23 @@
+using System;
+using UnityEngine;
+
+namespace Unity.Multiplayer.Samples.BossRoom.Client
+{
+    /// <summary>
+    /// Utility class used to listen for Animation-based events through Mecanim. This will invoke an Action when such
```
SamuelBellomo, line 7:
```suggestion
    /// Utility class used to listen for Animation-based events (events coming from the animation itself and propagated to monobehaviour scripts) through Mecanim. This will invoke an Action when such
```

SamuelBellomo: Did I understand this right?

Assets/BossRoom/Scripts/Client/AnimationCallbacks/ClientAnimListener.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// anim event is invoked.
+    /// </summary>
+    /// <remarks>
+    /// This class might in the future be refactored/removed with a Netcode for GameObjects refactor to NetworkAnimator.
```
SamuelBellomo, line 11: Has this been raised with the SDK team? Just to make sure this isn't lost in a comment here.

SamuelBellomo:
> since ClientCharacterVisualization component is spawned dynamically and childed to PlayerGraphics (important since the component is not in the same hierarchical level as the animator).

Any reason why ClientCharViz is not attached directly to PlayerGraphics? Why does it need to be dynamic?

SamuelBellomo:
> > since ClientCharacterVisualization component is spawned dynamically and childed to PlayerGraphics (important since the component is not in the same hierarchical level as the animator).
> 
> Any reason why ClientCharViz is not attached directly to PlayerGraphics? Why does it need to be dynamic?

Trying to simplify instead of adding an additional layer of complexity. I just wanna make sure we've tried removing the cause of the issue, before adding patching.

Assets/BossRoom/Scripts/Client/AnimationCallbacks/ClientAnimListener.cs
fernando-cortez: Yes that's correct.

fernando-cortez: > Trying to simplify instead of adding an additional layer of complexity. I just wanna make sure we've tried removing the cause of the issue, before adding patching.

I'll look into this simplification because you're right, it would remove the need for this one class.

Assets/BossRoom/Scripts/Client/AnimationCallbacks/ClientAnimListener.cs
fernando-cortez: Yup. It is https://github.com/Unity-Technologies/com.unity.netcode.gameobjects/issues/1389.

Assets/BossRoom/Scripts/Client/AnimationCallbacks/ClientAnimListener.cs
fernando-cortez: Class no longer needed! ClientCharacterVisualization handles anim events through Mecanim since it has been moved to the same GameObject as the Animator.

SamuelBellomo: So what was the move? And why is it there in the first place?

fernando-cortez: So the child of the PlayerAvatar prefab, PlayerGraphics, used to only have Animator and NetworkAnimator components. The prefab that was instantiated for the graphics per character class used to contain ClientCharacterVisualization. We're able to move it to the same level as PlayerGraphics, but just had to update some of the references inside of the components of the instantiated prefab.

SamuelBellomo (approved):

LPLafontaineB (approved):

PR #414, opened 2021-11-15
Shader Graph Toon Shader

jilfranco-unity:
### Description (*)
This PR is for a Shader Graph version of the toon shader for all the characters & imps. 
### Issue Number(s) (*)
Jira ticket [here](https://jira.unity3d.com/browse/MTT-1716)

Before:
image

After:
image

Tried to keep the graph pretty organized and commented as well:
image

I swapped all of the existing toon materials over to this new shader graph version, which thankfully was super quick and easy since all of the variable names were the same. I've also done some housecleaning of the Shaders folder, including deleting the old Royston toon shader. So now we have all shader graph shaders! This SG shader also uses a custom function node, which might be nice to have for users to look through as a more "advanced" usage of Shader Graph.

This also fixes the issue of the shader not rendering correctly in builds without reflection captures that was mentioned in [this PR](https://github.com/Unity-Technologies/com.unity.multiplayer.samples.coop/pull/371#:~:text=I%20am%20finding%20a%20bug%20in%20builds%20though%2C%20the%20first%20time%20through%20character%20select%2C%20the%20lighting%20on%20the%20toon%20shader%20breaks%20and%20looks%20like%20this%3A)


jilfranco-unity:
Setting this PR to urgent since it will go in the release Thursday

pdeschain:
I've learned that my shader graphs are poorly organized now :) The shader doesn't look like it would be particularly performance intensive and visually everything is ok. Great work!

pdeschain (approved):

jilfranco-unity:
Did some profiling on this and performance wise, render time is the same between the two versions! There do seem to be consistently more draw batches (3-5 that happen in the same amount of time as the 2-3 before the change) after the change, but I'd like to see how that changes at all after the other visual updates from my other PR come in

chrispope (approved):
Lgtm - nice parity on the visuals and clean shader graph.

fernando-cortez (approved):

PR #420, opened 2021-11-26
Fix: Graphics issue on android

jilfranco-unity:
### Issue Number(s) (*)
[Jira ticket here](https://jira.unity3d.com/browse/MTT-1500)

### Description (*)
Fixed a couple of bugs on low end android devices. The lighting issue was caused by the intensity of the lights being set too high. My thinking is that on the lowest end devices, Unity sets the light buffer to half precision, so we were seeing overflow. The lights in the main menu and char select scenes were set to like 150000, which is well above the limit for halfs, 65000. This also explains why it wasn't happening in the main scene, since the intensity for those lights is <100. 

Long story short, I was able to achieve a similar look by changing the colors and ranges of the lights, while keeping the light intensity down.

The compounding render texture error was caused by the SSAO being set to sample from scene depth and normals, setting it to just depth makes the error go away. 

I also modified the low quality settings a bit and added cheaper URP pipeline assets to be used with said low quality settings. Last but not least, I noticed there was a warning for the device not supporting HDR reflection probes, so I turned that setting off in the probe and also baked it.

It still runs like a slideshow on the device, but these bugs are resolved!

Here's the before images of lighting issues from the PR that Fernando took on the J2Core:

PXL_20211123_162129994
PXL_20211123_161944925

And the after ones that I sceenshoted from the J2Core:
Screenshot_20211126-144305
Screenshot_20211126-144332


LPLafontaineB:
I tested on PC and, slight nitpick, I found the Ready button and the P1 icon during the character selection seemed a lot brighter which makes them harder to read.

PR #427, opened 2021-12-14
feat: loading screen

LPLafontaineB:
### Description (*)

This PR adds a loading screen that appears when changing scenes.
### Issue Number(s) (*)

Fixes issue(s): [MTT-1960](https://jira.unity3d.com/browse/MTT-1960) [MTT-1746](https://jira.unity3d.com/browse/MTT-1746)
### Manual testing scenarios

1. Start a game with a host
2. Join with a client
3. Move the client around
4. Disconnect the client
5. Reconnect the client

The host and the client should see the loading screen when starting the game, then when changing scenes to the CharSelect scene and the BossRoom scene. Those loading screens should hide the slight delay before player characters are spawned, during which the additive scene is not loaded in the BossRoom scene, the camera is at its default position, and the UI is at its default state. This should be more evident during step 5, which should hide the camera movement described in [MTT-1746](https://jira.unity3d.com/browse/MTT-1746)

### Contribution checklist


SamuelBellomo:
mmm the diff contains changes from additive scene loading :( I tried forcing github to do a reload of the diff (by switching base branch), but that didn't fix it. @LPLafontaineB you might want to look into this before we review

LPLafontaineB:
> mmm the diff contains changes from additive scene loading :( I tried forcing github to do a reload of the diff (by switching base branch), but that didn't fix it. @LPLafontaineB you might want to look into this before we review

Yeah, it probably happened because I was working on this in parallel and started from the additive scene branch instead of develop. I'll see how I can fix this

LPLafontaineB:
Should be OK now @SamuelBellomo

SamuelBellomo:
just tested real quick, I still see unloaded maps after the spinning load screen.

LPLafontaineB:
> just tested real quick, I still see unloaded maps after the spinning load screen host side.

I expected that the fade-out could hide that since I didn't notice it during my tests. In that case, to do it properly I might have to create a more complex solution since I don't think we want loading screens each time a new additive scene is loaded after the actual game starts. I'll return this PR to draft while I work on that

Assets/BossRoom/Scripts/Client/UI/PostGameUI.cs
```diff
@@ -76,7 +77,7 @@ void SetPostGameUI(WinState winState)
         public void OnPlayAgainClicked()
         {
             // this should only ever be called by the Host - so just go ahead and switch scenes
-            NetworkManager.Singleton.SceneManager.LoadScene("CharSelect", LoadSceneMode.Single);
+            SceneLoader.Instance.LoadScene("CharSelect", LoadSceneMode.Single);
```
SamuelBellomo, line 80: shouldn't the switch to CharSelect remain synced with the other clients and the host?

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoader.cs
```diff
@@ -0,0 +1,137 @@
...
+
+namespace Unity.Multiplayer.Samples.Utilities
+{
+    public class SceneLoader : NetworkBehaviour
+    {
+        [SerializeField]
+        List<string> m_ScenesTriggeringLoadingScreen;
+
+        [SerializeField]
+        List<string> m_ScenesEndingLoadingScreen;
```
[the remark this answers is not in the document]
LPLafontaineB: That's what I did originally, but it doesn't work as well when loading the BossRoom scene. In that case, it would stop the loading screen before the first additive scene loads, and before the player characters spawn, which is what it was supposed to hide. Basically, I need a way to prevent the loading screen from stopping when the BossRoom scene is loaded and have it wait for all additive scenes. But I agree it is not an elegant solution.

I could get the RTT and add it as a delay before stopping the loading screen, which should give enough time to start loading the additive scenes, but I'm not sure how to do this cleanly yet. Plus I'm not a fan of adding these kinds of delays, it feels like a workaround. I will try to think of something else.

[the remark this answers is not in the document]
LPLafontaineB: Nevermind, it works. The issue I remembered was probably due to me using the wrong events in the beginning of this PR.

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoader.cs
```diff
@@ -0,0 +1,141 @@
...
+using Unity.Netcode;
+using UnityEngine;
+using UnityEngine.SceneManagement;
+
+namespace Unity.Multiplayer.Samples.Utilities
+{
+    public class SceneLoader : NetworkBehaviour
+    {
+        /// <summary>
+        /// This singleton is a utility handling the loading of scenes. It loads scene using the SceneManager, or, on
```
SamuelBellomo, line 11:
```suggestion
        /// Adds loading screens by wrapping scene management APIs. It loads scene using the SceneManager, or, on
```
Just to put the emphasize on the right thing. Users with no loading screens don't need this class, correct?

Assets/BossRoom/Scripts/Client/Game/State/ClientCharSelectState.cs
```diff
@@ -403,7 +403,7 @@ public void OnPlayerExit()
             // first disconnect then return to menu
             var gameNetPortal = GameObject.FindGameObjectWithTag("GameNetPortal").GetComponent<GameNetPortal>();
             gameNetPortal.RequestDisconnect();
-            SceneManager.LoadScene("MainMenu");
+            SceneLoader.Instance.LoadScene("MainMenu", LoadSceneMode.Single);
```
SamuelBellomo, line 406: instead of adding single everywhere, couldn't we add a default value to SceneLoader's LoadScene?

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/GameNetPortal.cs
```diff
@@ -243,6 +244,11 @@ void StartHost()
         {
             SessionManager<SessionPlayerData>.Instance.AddHostData(new SessionPlayerData(NetManager.LocalClientId, PlayerName, m_AvatarRegistry.GetRandomAvatar().Guid.ToNetworkGuid(), 0, true));
             NetManager.StartHost();
+            SceneLoader.Instance.AddOnSceneEventCallback();
+
+            //The "BossRoom" server always advances to CharSelect immediately on start. Different games
+            //may do this differently.
+            SceneLoader.Instance.LoadScene("CharSelect", LoadSceneMode.Single);
```
SamuelBellomo, line 251: I'm curious why this was moved?

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoader.cs
```diff
@@ -0,0 +1,141 @@
+using System.Collections.Generic;
+using Unity.Netcode;
+using UnityEngine;
+using UnityEngine.SceneManagement;
+
+namespace Unity.Multiplayer.Samples.Utilities
+{
+    public class SceneLoader : NetworkBehaviour
```
SamuelBellomo, line 8: Non blocker nitpicking: That name tickles me. When I first saw it used in game code, I was wondering "why isn't Netcode's scene manager used anymore?"
and then I finally understood SceneLoader is a convenience wrapper that adds loading screen management to normal scene loading calls.
Maybe SceneLoaderWrapper? LoadingScreenManager?
I don't know, naming things is hard :/

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoader.cs
```diff
@@ -0,0 +1,141 @@
...
+        {
+            if (Instance != null && Instance != this)
+            {
+                Destroy(this.gameObject);
+            }
+            else
+            {
+                Instance = this;
+            }
+            DontDestroyOnLoad(this);
```
SamuelBellomo, line 31: I know Netcode used to have issues with DontDestroyOnLoad and scene switching. That's solved right? If you finish a game and try to start a new one, this should work fine correct?

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoader.cs
```diff
@@ -0,0 +1,141 @@
...
+        {
+            if (NetworkManager != null && NetworkManager.SceneManager != null)
+            {
+                NetworkManager.SceneManager.OnSceneEvent -= OnSceneEvent;
+            }
+            SceneManager.sceneLoaded -= OnSceneLoaded;
+        }
+
+        // This needs to be called right after initializing NetworkManager (after StartHost, StartClient or StartServer)
+        public void AddOnSceneEventCallback()
```
SamuelBellomo, line 49: public methods should have a quick doc.

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoader.cs
```diff
@@ -0,0 +1,141 @@
...
+            }
+            DontDestroyOnLoad(this);
+        }
+
+        void Start()
+        {
+            SceneManager.sceneLoaded += OnSceneLoaded;
+        }
+
+        public override void OnDestroy()
```
SamuelBellomo, line 39: does this need to be public?

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoader.cs
LPLafontaineB: It is hard :( I probably spent more time trying to name this class than coding this entire PR. It has this name because my original intention was to create a generic wrapper for loading scenes because there is no callback I could attach to to start the loading screen when offline. But in the end, the only use it has is to tell the loading screen when to start and when to stop.

I think SceneLoaderWrapper would make more sense. LoadingScreenManager would create confusion with those wrapper methods I think, since from the outside they still load scenes first.

Assets/BossRoom/Scripts/Client/Game/State/ClientCharSelectState.cs
LPLafontaineB: We could, but I like having it that way, like how it is with NetworkSceneManager. It makes it more deliberate to chose to load a scene in Single mode.

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoader.cs
LPLafontaineB: It does not work fine. If I start a game, then quit, then start a new one, I get an exception on the client that says: `Unhandled RPC Exception: System.InvalidCastException: Specified cast is not valid.` when joining. It comes from `StopLoadingScreenClientRpc`.

I'll try to debug this

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoader.cs
LPLafontaineB: The event function it overrides in NetworkBehavior is public. The IDE gives me an error if I make it private, and if I just remove the override keyword I get a warning because it hides NetworkBehavior's method

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/GameNetPortal.cs
LPLafontaineB: It was because it would execute before AddOnSceneEventCallback could be called, but looking back at it I did not need to move it, just to put AddOnSceneEventCallback in the same place instead. I'll move it back, I think it is better how it was.

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoader.cs
SamuelBellomo: I'm open to other names too, SceneLoaderWrapper isn't great either

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoader.cs
SamuelBellomo: oh I forgot NetworkBehaviour implemented this. NVM nothing to see here

Assets/BossRoom/Scripts/Client/Game/State/ClientCharSelectState.cs
SamuelBellomo: And the static LoadScene method has default parameters, just like Unity's SceneManager does too https://docs.unity3d.com/ScriptReference/SceneManagement.SceneManager.LoadScene.html
This way, users that want simple calls can just use the default behaviour, without preventing them to be explicit about it (those that want to be explicit can still "override" the default param with the same default param)

LPLafontaineB: I don't see any static LoadScene method with default parameters in NetworkSceneManager, only in Unity's SceneManager, am I looking in the wrong place? Either way, I'll add it some to this new method then

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoader.cs
LPLafontaineB: Pre.4 fixes this issue, but there is another one still. In short, if the host is shut down, then restarted, the NetworkObject is not re-spawned. I created a [github issue](https://github.com/Unity-Technologies/com.unity.netcode.gameobjects/issues/1574) for it. I think we should put this PR on hold in the meantime.

SamuelBellomo: When you say restarted and the NO is not respawned, you mean host side right? Cause I wouldn't expect a client to still be connected after a shutdown.
So when starting a 2nd game in the same process, the host can't have that object present a second time? yeah that's bad :(

LPLafontaineB: It actually happens on both the host and the client. After leaving a game, if any of the two game processes try to restart a new game, the NetworkObject will not be spawned. As you intuited, it doesn't make the game unplayable, it simply doesn't activate the loading screen on the client (it does still activate them on the host, though). However, if we shutdown the session a second time, we will be stuck in the main menu with an infinite loading screen. In short, the game is playable once normally, then a second time with no loading screen on the client, but not a third time.

If, however, we have a host or client that has started hosting or joined a game, then quits, then tries to join a game that is hosted by a new process, it will give use exceptions and fail to spawn the NetworkObject, which makes the game unplayable. The reverse case (a host that has a previous session joined by a new client) does not cause this issue.

SamuelBellomo: yeah ok, I agree, let's keep this on hold then

LPLafontaineB:
On hold, waiting for [this issue](https://github.com/Unity-Technologies/com.unity.netcode.gameobjects/issues/1571) to be solved on the sdk

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ClientGameNetPortal.cs
```diff
@@ -234,6 +235,7 @@ void ConnectClient()
             //  If the socket connection fails, we'll hear back by getting an ReceiveServerToClientSetDisconnectReason_CustomMessage callback for ourselves and get a message telling us the reason
             //  If the socket connection succeeds, we'll get our ReceiveServerToClientConnectResult_CustomMessage invoked. This is where game-layer failures will be reported.
             m_Portal.NetManager.StartClient();
+            SceneLoaderWrapper.Instance.AddOnSceneEventCallback();
```
SamuelBellomo, line 238: 1. ClientGameNetPortal shouldn't have to know about this, it's not its responsibility and just adds coupling.
2. method name is a bit weird, add...something usually means you're adding something and would expect a parameter. "Init" something might be better.

LPLafontaineB: The issue with this is that to trigger the loading screen at the beginning of client syncrhonization, we need to register those callbacks before they happen. However, the NetworkSceneManager is created when StartClient / StartHost / StartServer is called, so we have to register them after this. OnClientConnected and OnNetworkSpawn happen after the synchronization has finished so it must be done before those callbacks. We currently don't have access to callbacks when starting clients, which is why I had to do it this way.

SamuelBellomo: I'd create our own callback for this then, instead of having our connection management know about scene management initialization?

LPLafontaineB: If we do that though, that means that we need to add coupling in the other direction, which means we will have to move the loading screen stuff out of the Utilities package. My initial goal was to create something generic enough to be able to be reused in other projects, but if it makes more sense that way, I can move it.

SamuelBellomo: @pdeschain you have the arch cleanup initiative. What do you think of moving connection management to Utilities as well? To me it's the kind of stuff we'd want to reuse in future samples anyway

LPLafontaineB: Should that be part of a future ticket instead? Or should we move all of that to the Shared assembly for now and evaluate later what should be moved to Utilities?

SamuelBellomo: future ticket, just make sure it's tracked :)

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/ClientLoadingScreen.cs
```diff
@@ -0,0 +1,105 @@
...
+                m_LoadOperation = loadOperation;
+                m_ProgressBar.value = 0;
+            }
+        }
+
+        IEnumerator FadeOutCoroutine()
+        {
+            m_ProgressBar.value = 1;
+            float currentTime = 0;
+            while (currentTime < m_DelayBeforeFadeOut)
```
SamuelBellomo, line 87: this should be a WaitForSeconds

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoaderWrapper.cs
```diff
@@ -0,0 +1,146 @@
...
+                    {
+                        m_ClientLoadingScreen.StopLoadingScreen();
+                    }
+                    break;
+                case SceneEventType.SynchronizeComplete: // Client told server that they finished synchronizing
+                    // Only executes on server
+                    if (NetworkManager.IsServer)
+                    {
+                        // Send client RPC to make sure the client stops the loading screen after the server handles what it needs to after the client finished synchronizing
+                        StopLoadingScreenClientRpc(new ClientRpcParams {Send = new ClientRpcSendParams {TargetClientIds = new[] {sceneEvent.ClientId}}});
```
SamuelBellomo, line 134: Why do we need to wait for the server to tell us to stop the loading screen? Do we have an event client side once we're done syncing? Why the back and forth?

SamuelBellomo:
Removing the on hold label for review, however let's make sure we merge this only once we have pre.7 updated

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ClientGameNetPortal.cs
LPLafontaineB: here it is: https://jira.unity3d.com/browse/MTT-3157

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoaderWrapper.cs
```diff
@@ -0,0 +1,146 @@
...
+                    if (NetworkManager.IsClient)
+                    {
+                        m_ClientLoadingScreen.StopLoadingScreen();
+                    }
+                    break;
+                case SceneEventType.SynchronizeComplete: // Client told server that they finished synchronizing
+                    // Only executes on server
+                    if (NetworkManager.IsServer)
+                    {
+                        // Send client RPC to make sure the client stops the loading screen after the server handles what it needs to after the client finished synchronizing
```
SamuelBellomo, line 133:
```suggestion
                        // Send client RPC to make sure the client stops the loading screen after the server handles what it needs to after the client finished synchronizing, for example character spawning done server side should still be hidden by loading screen.
```

SamuelBellomo (approved):

fernando-cortez (dismissed):

SamuelBellomo (dismissed):

LPLafontaineB:
Adding DoNotMerge label until BossRoom is updated to NGO pre.7

SamuelBellomo (approved):

fernando-cortez (approved):

Issue #275, opened 2021-04-13
Timeout or Cancel button needed for join attempts.


JesusLuvsYooh:
You can get stuck on the 'Connecting  Attempting to join...' section.
Needs a timeout variable or cancel button.

To reproduce:
Open game in Editor, change nothing, click join as a client (don't open a host).

![Screenshot 2021-04-13 at 10 41 22](https://user-images.githubusercontent.com/57072365/114532712-037afb80-9c45-11eb-8b39-6f3c3ef78cce.jpg)

PR #428, opened 2021-12-15
feat: cancel during join or host

LPLafontaineB:
### Description (*)

This PR adds a cancel button to allow cancellation of joining or hosting attempts.
### Issue Number(s) (*)

Fixes issue(s): #275 
### Manual testing scenarios

Those are easiest to test by using Unity Relay with Clumsy active, with a high simulated latency for receiving packets. Otherwise it happens quickly and is hard to cancel before it does. It can also be tested easily when trying to join via direct IP using the wrong IP.
1. Start BossRoom and attempt to host or join a game
2. Cancel hosting or joining
3. See that you are back at the main menu and can now try again
### Contribution checklist


Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ClientGameNetPortal.cs
```diff
@@ -239,22 +243,33 @@ public async void StartClientUnityRelayModeAsync(GameNetPortal portal, string jo
...
 
                     break;
                 default:
                     throw new Exception($"unhandled relay transport {chosenTransport.GetType()}");
             }
 
-            ConnectClient(portal);
+            if (!cancellationToken.IsCancellationRequested)
+            {
+                ConnectClient(portal);
```
SamuelBellomo, line 271: Thinking of this, there's no way to cancel the connection itself? Cause that part too could have delays and issues.
If you test with clumsy or even try to connect to an invalid address, there's no way to cancel that once that's launched?

LPLafontaineB: From my investigation, I haven't found a way to cancel the connection itself. However, If we cancel the operation after this part has started, we still call NetworkManager's Shutdown method. On the user side, it seems to properly cancel the connection. I just tested it again with direct IP and with UnityRelay and it looks like it works fine

SamuelBellomo: ah so "Shutdown" would be what cancels the thing. ok
it might be worth it to add a quick comment on this for users investigating the same thing

SamuelBellomo (approved):

SamuelBellomo (approved):

pdeschain (approved):

PR #430, opened 2021-12-20
feat: god mode cheat

LPLafontaineB:
### Description (*)

This PR adds a cheat to enable/disable god mode, preventing player characters from taking damage while active. To use it, open the cheat window with either the host or a client, and click on the "Toggle God Mode" button. This will make your character invulnerable. It will also add an indication on the character's health bar in the HUD, which is also visible for other players.
At the moment, this indicator is simply an image that is shown on top of the health bar that I selected from what we already had, but we might want a custom image for this, or a different kind of indication. See the screenshots below for how it looks on the player and an ally.

GodModeActivatedMainCharacter
GodModeActivatedAlly

### Related Pull Requests
#444
### Issue Number(s) (*)

Fixes issue(s): [MTT-1973](https://jira.unity3d.com/browse/MTT-1973)
### Manual testing scenarios

1. Start a game
2. Open the cheat window using `/` key
3. Press the "Toggle God Mode" button
3. See that your player can no longer be damaged and that an indicator appears on their health bar in the HUD
### Contribution checklist


Assets/BossRoom/Scripts/Server/ServerTestingHotkeys.cs
```diff
@@ -58,6 +65,13 @@ private void Update()
             {
                 NetworkManager.SceneManager.LoadScene("PostGame", LoadSceneMode.Single);
             }
+            if (m_GodModeKeyCode != KeyCode.None && Input.GetKeyDown(m_GodModeKeyCode))
+            {
+                if (m_BossRoomState)
+                {
+                    m_BossRoomState.ToggleGodMode();
+                }
```
fernando-cortez, line 73: So this is the first action test/debug hotkey that will now impact several parts of code, and we haven't really discussed how that should be handled compositionally.

In my experience, game-wide toggles like these were kept in a single class wrapped in a preprocessor directive, such that the game could be tested in non-development mode, without any possible test/debug tools influence.

My suggestion here is that we keep `godMode` toggle inside its own `NetworkBehaviour`, since this toggle would only be modified by a server, and game-side classes like `DamageReceiver` would just read from this global value.

@pdeschain thoughts?

Assets/BossRoom/Scripts/Server/ServerTestingHotkeys.cs
```diff
@@ -58,6 +65,13 @@ private void Update()
...
+            }
```
fernando-cortez, line 74: Another thing which I thought of after reviewing: this new godMode cheat deserves some UI display, since anyone toggling the G key accidentally will now think the game is unresponsive.

Assets/BossRoom/Scripts/Server/ServerTestingHotkeys.cs
LPLafontaineB: Good point! I agree this would be a lot cleaner than what I did

Assets/BossRoom/Scripts/Server/ServerTestingHotkeys.cs
LPLafontaineB: You're right. At the moment I only added a Console.Log and even knowing about it I feel it can be easy to miss.
There is [this ticket](https://jira.unity3d.com/browse/MTT-1994) about adding a popup window for cheats. I think it would be a good solution. Or if we want something specific that is always visible, we could add a label similar to the NetworkLatencyWarning

Assets/BossRoom/Scripts/Server/ServerTestingHotkeys.cs
SamuelBellomo: any reason we're doing god mode game wide and not just per player? That'd simplify this

LPLafontaineB: I chose this way since ServerTestingHotkeys is only enabled on host, but I could create a ClientTestingHotkeys that has RPC calls to this, or something similar. I think it would require adding another class in shared with callbacks though

fernando-cortez: That's definitely another approach. Having god-mode per player gives us some flexibility when testing, and it would simplify things. I just figured host would control debug game-wide settings but this works too.

pdeschain: I would keep the bool godMode where it is (for now), simply because it's the only cheat we have and there is little reason to architect something more readable and clear then that. The state modification is pretty clean imo.

pdeschain: If we're talking architecture - I'd say that considering segregating cheats into a separate class (from game data perspective) is small potatoes in comparison to solving the way our objects resolve their dependencies overall. 

Solving that ^^ would immediately have an effect on how we would segregate our Cheat-manager (and make it a dependency of all the interested classes), our Input manager and our update loop manager, and then wire them all together along with everything else in the game (well, whatever we decide we want to pipe around using this mechanism).

@LPLafontaineB a little nitpick - is it possible for the cheat manager to not be a NetworkBehaviour? I guess not at this moment, as it needs to know isServer to disable itself. I would much prefer classes to not be NetworkBehaviours, or even MonoBehaviours at all if it's possible, but I think that at this point it would require some infrastructure legwork.

LPLafontaineB: If I'm understanding correctly, @pdeschain you're saying that I should wait until we decide how and if we want to refactor the global architecture before refactoring the cheat manager, right? If so, I agree.

About your point on NetworkBehaviours and MonoBehaviours, I 100% agree. Here I wanted to keep this PR lightwieght if possible, but if we plan on refactoring the cheats system, we should try to see if it could not be either of those. At this moment, this uses serialized fields for the keycodes of cheats, and the prefabs to spawn, so it might require some thought. I think that would be another task altogether

Assets/BossRoom/Scripts/Client/UI/PartyHUD.cs
```diff
@@ -158,6 +161,14 @@ void SetUIFromSlotData(int slot, NetworkCharacterState netState)
...
             m_PartyNames[slot].text = GetPlayerName(netState);
 
+#if UNITY_EDITOR || DEVELOPMENT_BUILD
+            m_PartyHealthGodModeImages[slot].gameObject.SetActive(netState.NetworkLifeState.IsGodMode.Value);
+            netState.NetworkLifeState.IsGodMode.OnValueChanged += (value, newValue) =>
+            {
+                m_PartyHealthGodModeImages[slot].gameObject.SetActive(newValue);
+            };
+#endif
+
```
fernando-cortez, line 171: For sanity's sake, include remove the subscription to this callback when the hero or allies are removed.

Assets/BossRoom/Scripts/DebugCheats/DebugCheatsManager.cs
```diff
@@ -75,6 +81,20 @@ void GoToPostGameServerRpc(ServerRpcParams serverRpcParams = default)
...
+        {
+            ulong clientId = serverRpcParams.Receive.SenderClientId;
+            foreach (var playerServerCharacter in PlayerServerCharacter.GetPlayerServerCharacters())
+            {
+                if (playerServerCharacter.OwnerClientId == clientId)
+                {
+                    playerServerCharacter.NetState.NetworkLifeState.IsGodMode.Value = !playerServerCharacter.NetState.NetworkLifeState.IsGodMode.Value;
+                }
+            }
+            LogCheatUsedClientRPC(serverRpcParams.Receive.SenderClientId, "GoToPostGame");
```
fernando-cortez, line 95: Log string param should be "GodMode", correct?

Assets/BossRoom/Scripts/Server/Game/Character/ServerCharacter.cs
```diff
@@ -189,6 +189,11 @@ public void ReceiveHP(ServerCharacter inflicter, int HP)
             }
             else
             {
+#if UNITY_EDITOR || DEVELOPMENT_BUILD
+                // Don't apply damage if god mode is on
+                if (NetState.NetworkLifeState.IsGodMode.Value) return;
```
fernando-cortez, line 194: Just a stylistic, but consistent across the codebase, suggestion to wrap the return in brackets in new lines.

Assets/StreamingAssets.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
```
fernando-cortez, line 1: This keeps getting generated every time we make builds. I'll create a task to see if it's safe to ignore this on git. But remove in the meantime.

fernando-cortez (commented):
Besides the few comments, looks mostly gtg. I do actually like the look of the yellow bar underneath to represent god mode. Clean.

Assets/BossRoom/Scripts/DebugCheats/DebugCheatsManager.cs
LPLafontaineB: Oops, you're right I missed this one

Assets/StreamingAssets.meta
LPLafontaineB: I thought I made sure to not include it this time. It really doesn't want to leave us alone

fernando-cortez (approved):

SamuelBellomo (approved):

PR #431, opened 2021-12-20
feat: button to copy room name to clipboard

LPLafontaineB:
### Description (*)

Added button to copy room name to clipboard in character selection, on host side.
### Issue Number(s) (*)

Fixes issue(s): [MTT-1988](https://jira.unity3d.com/browse/MTT-1988)
### Manual testing scenarios

1. ...
2. ...
### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Scripts/Client/UI/RoomNameBox.cs
+++ b/Assets/BossRoom/Scripts/Client/UI/RoomNameBox.cs
@@ -3,7 +3,6 @@ using Netcode.Transports.PhotonRealtime;
 using UnityEngine;
 using UnityEngine.Assertions;
 using TMPro;
-using Unity.Multiplayer.Samples.BossRoom.Visual;
 using Unity.Netcode;
 using UnityEngine.UI;
 
--- a/Assets/BossRoom/Scripts/Client/UI/RoomNameBox.cs
+++ b/Assets/BossRoom/Scripts/Client/UI/RoomNameBox.cs
@@ -3,11 +3,18 @@ using Netcode.Transports.PhotonRealtime;
 using UnityEngine;
 using UnityEngine.Assertions;
 using TMPro;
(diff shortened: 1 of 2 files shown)


Assets/BossRoom/Scripts/Client/UI/RoomNameBox.cs
```diff
@@ -4,10 +4,16 @@
...
 
 public class RoomNameBox : MonoBehaviour
 {
-    [SerializeField] TextMeshProUGUI m_RoomNameText;
+    [SerializeField]
+    TextMeshProUGUI m_RoomNameText;
+    [SerializeField]
+    Button m_CopyToClipboardButton;
+
+    string m_RoomName;
```
SamuelBellomo, line 16: is there a possibility that CopyToClipboard will be clicked before m_RoomNameText.text is initialized? I'm not a big fan of storing data in gui text, but in this case here, we're literally asking to copy the content of the text box, why not just use m_RoomNameText to store the room name? (I'm asking so we have one less state to maintain)

LPLafontaineB: I don't think there is, since m_RoomNameText.text is initialized in Awake(), and then is set in ConnectionFinished(), before the button is enabled.
As for m_RoomName, I put the code in a different variable to follow the structure we had with Photon for m_RoomNameText.text ($"Room Name: {roomName}") which I find is clearer for players.

SamuelBellomo: oh right I get what you mean. nvm then :)

SamuelBellomo (approved):

jilfranco-unity (approved):
Looks good to me! Built & tested with both Photon and Unity Relay and it worked as expected

PR #432, opened 2021-12-21
feat: disconnect/quit button

LPLafontaineB:
### Description (*)

This PR adds a button to disconnect/quit game to the settings panel. It also removes the exit button from the character selection scene to prevent duplicating code.
### Issue Number(s) (*)

Fixes issue(s): [MTT-2008](https://jira.unity3d.com/browse/MTT-2008)
### Manual testing scenarios

1. ...
2. ...
### Contribution checklist


Assets/BossRoom/Scripts/Client/UI/UIQuitPanel.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using Unity.Multiplayer.Samples.BossRoom;
+using Unity.Netcode;
+using UnityEngine;
+using UnityEngine.SceneManagement;
+using UnityEngine.UI;
+
+public class UIQuitPanel : MonoBehaviour
```
fernando-cortez, line 8: Add to namespace.

Assets/BossRoom/Scripts/Shared/Game/Action/ActionUtils.cs
```diff
@@ -88,7 +88,7 @@ public static bool IsValidTarget(ulong targetId)
             //note that we DON'T check if you're an ally. It's perfectly valid to target friends,
             //because there are friendly skills, such as Heal.
 
-            if (!NetworkManager.Singleton.SpawnManager.SpawnedObjects.TryGetValue(targetId, out var targetChar))
+            if (NetworkManager.Singleton.SpawnManager == null || !NetworkManager.Singleton.SpawnManager.SpawnedObjects.TryGetValue(targetId, out var targetChar))
```
fernando-cortez, line 91: Is this related?

Assets/BossRoom/Scripts/Client/UI/UISettingsCanvas.cs
```diff
@@ -11,13 +11,17 @@ public class UISettingsCanvas : MonoBehaviour
...
+        private GameObject m_QuitPanelRoot;
+
         void Awake()
         {
             // the settings canvas should exist in all scenes!
             DontDestroyOnLoad(gameObject);
 
             // hide the settings window at startup (this is just to handle the common case where an artist forgets to disable the window in the prefab)
             m_SettingsPanelRoot.SetActive(false);
+            m_QuitPanelRoot.SetActive(false);
```
fernando-cortez, line 24: Another case to consider is disabling the canvas when changing scenes, if it has been enabled. I only mention this since the context of the window can change.

Assets/BossRoom/Scripts/Shared/Game/Action/ActionUtils.cs
LPLafontaineB: Oops I forgot to document this change. It is not related, just a bug that happens when disconnecting while in-game. It did not happen often earlier (since the only case where this happened was when a host quit the game while clients are still playing), but when testing with the disconnect button it caused lots of exceptions.

SamuelBellomo (dismissed):
after other comments are solved, I'm good

fernando-cortez (approved):

jilfranco-unity:
I love this feature! Very needed. I have some suggestions for visuals though, which include this menu, but also apply to other related UI elements, so I won't commit my changes in case we want to break them up into different PRs <br /><br /><br />


I think there needs to be more padding above the quit menu label, maybe something like this:

Before / After
image image

And I know that'll create some issues for when trying to quit in game, so instead of "Disconnect from current session?" we could do "Leave Session?" like so:

Before / After
image image

We can then turn off Best Fit and have the text aligned to the middle of the field correctly <br /><br />

Also, something minor, I like that the exit menu is toggled on and off with the exit button, rather than also having its own exit button. So I'd like to bring that change to the settings menu as well

Before / After
image image

And delete the exit button from the how to play popup at the start of a game:

Before / After
image 
image

Which helps eliminate there being so many close buttons at the start of a game, which will look even nicer in non dev builds with no debug menu <br /><br /><br />

I also made some scaling changes to the text in the how to play menu since I was in the neighborhood:

Before / After
image image

Generally speaking, in the world of graphic design, non-uniformly scaling font is a no-no :P 

Lemme know what y'all think!

LPLafontaineB:
> Lemme know what y'all think!

I like those changes, the UI looks much better that way. I think they could be included here without requiring to be broken up into different PRs

fernando-cortez:
> Lemme know what y'all think!

No reason to have multiple close buttons on UI objects, so I'm all for it. Same with the padding.

What about making the font uniform across all pop up text? I know we've heard that remark a couple of times from people with fresh eyes.

Assets/BossRoom/Scripts/Client/UI/UIQuitPanel.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Unity.Multiplayer.Samples.BossRoom.Client
+{
+    public class UIQuitPanel : MonoBehaviour
+    {
+        [SerializeField]
+        Text m_QuitButtonText;
+
+        void OnEnable()
+        {
+            m_QuitButtonText.text = NetworkManager.Singleton.IsListening ? "Disconnect from current session?" : "Exit Game?";
```
fernando-cortez, line 17: This is returning a NullReferenceException when MainMenu scene is loaded.

fernando-cortez (dismissed):

Assets/BossRoom/Scripts/Client/UI/UIQuitPanel.cs
```diff
@@ -0,0 +1,39 @@
...
+                "Leave session?" :
+                "Exit Game?";
+        }
+
+        public void Quit()
+        {
+            if (NetworkManager.Singleton.IsListening)
+            {
+                // first disconnect then return to menu
+                var gameNetPortal = GameObject.FindGameObjectWithTag("GameNetPortal").GetComponent<GameNetPortal>();
```
SamuelBellomo, line 27: GameNetPortal has a static "Instance".
Finding objects with tags is a maintenance nightmare (can't "find all references" on those)

LPLafontaineB: Didn't see it had a static instance, I'll switch to that. I didn't like using tags for that, but from what I understood it was better than FindObjectOfType() in terms of performance.

SamuelBellomo (approved):

fernando-cortez (dismissed):

LPLafontaineB:
Saw a callback that wasn't cleaned up in OnDestroy() while merging

fernando-cortez (approved):

SamuelBellomo (approved):

PR #433, opened 2021-12-21
fix: join code not displayed for clients using Unity Relay

LPLafontaineB:
### Description (*)

This PR makes the Unity Relay join code accessible to clients after joining.
### Issue Number(s) (*)

Fixes issue(s): [MTT-2009](https://jira.unity3d.com/browse/MTT-2009)
### Manual testing scenarios

1. Start a game as a host using Unity Relay
2. Join the game as a client
3. See that now the join code is properly displayed on both client and host
### Contribution checklist


Assets/BossRoom/Scripts/Shared/Net/UnityRelayUtilities.cs
```diff
@@ -36,6 +35,7 @@ public static async
             try
             {
                 joinCode = await Relay.Instance.GetJoinCodeAsync(allocation.AllocationId);
+                RelayJoinCode.Code = joinCode;
```
SamuelBellomo, line 38: I'm thinking, this utility could be a good candidate for the utility asmdef. With this in mind, would there be a way to do this without adding gameplay specific code here?

LPLafontaineB: RelayJoinCode is a static class containing only the static field Code defined in this same file. It used to be set from GameNetPortal using this same value, but I think it makes more sense to set it here directly and remove the unnecessary clutter. I think we could even remove RelayJoinCode and simply add the join code as a static field of RelayUtility.

LPLafontaineB: If we want to keep the static class holding the join code separate though, we could also revert this change, and add a line in ClientGameNetPortal to set RelayJoinCode.Code after the connection succeeded

LPLafontaineB: Which solution do you think would make more sense @SamuelBellomo ?

pdeschain (approved):
Will need to remember to do the same for Lobby!

Assets/BossRoom/Scripts/Shared/Net/UnityRelayUtilities.cs
SamuelBellomo: as discussed, static field in RelayUtility

SamuelBellomo (approved):

SamuelBellomo (approved):

pdeschain (approved):

PR #436, opened 2021-12-22
fix: more responsive actions

LPLafontaineB:
### Description (*)

This PR increases the speed of base attacks, by increasing the speed of the corresponding animations by a factor of 1.66 and reducing the action durations by half. It also increases the speed of the tank shield and mage heal abilities in the same way (half duration, 1.2x animation speed). Finally, character movement for all characters is increased from 4 to 6.
### Issue Number(s) (*)

Fixes issue(s): [MTT-1992](https://jira.unity3d.com/browse/MTT-1992)
### Contribution checklist


jilfranco-unity:
From an art standpoint, I think this change is great! Makes the game feel and look more responsive. I think the only animation that doesn't as good sped up is the archer volley since that was already a pretty fast one. Could we do a check in code somewhere for this animation to stay at 1x speed instead of 1.5 like the others?

LPLafontaineB:
> From an art standpoint, I think this change is great! Makes the game feel and look more responsive. I think the only animation that doesn't as good sped up is the archer volley since that was already a pretty fast one. Could we do a check in code somewhere for this animation to stay at 1x speed instead of 1.5x like the others?

Oops I just checked and turns out that the "Attack2" animation that I also sped up was actually the volley animation. Good catch! I reverted it back to 1x speed and it seems ok now.

SamuelBellomo:
good improvements. I'm wondering what's the max speed we could add to these without it being too weird? I'm asking just so we don't seem like we're cheating too much with Netcode by hiding our issues with long animations.

Also, any thoughts on increasing character movement speed?

LPLafontaineB:
> good improvements. I'm wondering what's the max speed we could add to these without it being too weird? I'm asking just so we don't seem like we're cheating too much with Netcode by hiding our issues with long animations.

As suggested by Jill, I'll try to increase the animations' speed to 1.66x

> Also, any thoughts on increasing character movement speed?

Not sure how that would make the game more responsive? But I can try tweaking it as well, it will probably make the game more fun anyways

LPLafontaineB:
As discussed @SamuelBellomo I increased the attack and movement speed further. I also increased the speed of the tank shield and mage heal. We might need to tweak the animations if we want to keep that kind of speed instead of just playing them faster.

SamuelBellomo (approved):

SamuelBellomo:
looks great!
What's interesting is now players are more OP :P I managed to clear the first room with just the mage

LPLafontaineB:
> looks great! What's interesting is now players are more OP :P I managed to clear the first room with just the mage

Yeah, the mage and archer are both considerably more powerfull now. To balance things out, we could reduce their damage output I suppose, or increase the imp's Hp

jilfranco-unity:
Gameplay feels nice and snappy now, I like it! :) But I think the animations are way too fast now. The main thing that makes these sped up animations stick out is the head movement being way faster than the idle, which is still at 1-- it makes it obvious that these animations weren't supposed to be played at this speed. If we wanted to play them this fast, we'd have to edit the animations themselves. 

So for now, I think taking the speed of attack1 from 2 to 1.66 and the heal/buff animation down from 2 to 1.2 is the way to go. I tested these slower animations with the new sped up gameplay and they still line up nicely.

LPLafontaineB:
> So for now, I think taking the speed of attack1 from 2 to 1.66 and the heal/buff animation down from 2 to 1.2 is the way to go

Tested it as well and I agree. I did not notice the head movement issue, but now that you mentioned it I cannot unsee it. It does seem to line up nicely for pretty much everything, except the rogue's attack which is a little weird, but it's not that bad. It may be because it is slower or because it has more anticipation, but with the animation at 1.66 and the gameplay sped up by 2x it looks like the rogue damages things before they hit them. We do need to look closely to notice it, and unless it is on pots or spawner crystals it is far from obvious, so I believe it's fine.

> Also, after this is merged in, I'll edit the tank's slash effect material so that the timing matches :)

I thought I did by reducing the start delay on the particle system, but you're right it doesn't seem to have fixed it properly.
Also, would it make sense for you to do it in this same PR instead of creating a new one after it is merged?

jilfranco-unity:
> Tested it as well and I agree. I did not notice the head movement issue, but now that you mentioned it I cannot unsee it. 

Ahahaha I am sorry for that :P 

>It does seem to line up nicely for pretty much everything, except the rogue's attack which is a little weird, but it's not that bad. It may be because it is slower or because it has more anticipation, but with the animation at 1.66 and the gameplay sped up by 2x it looks like the rogue damages things before they hit them. We do need to look closely to notice it, and unless it is on pots or spawner crystals it is far from obvious, so I believe it's fine.

Hmm yeah let's leave the rogue speed as is, I've be experimenting with editing some of the animations, so I can edit it in the future. 

> I thought I did by reducing the start delay on the particle system, but you're right it doesn't seem to have fixed it properly. Also, would it make sense for you to do it in this same PR instead of creating a new one after it is merged?

And sure! Just didn't want to step on any toes :P Can I just push my own commits to your PR? Or I can leave screenshots of the settings here if that works better

LPLafontaineB:
> And sure! Just didn't want to step on any toes :P Can I just push my own commits to your PR? Or I can leave screenshots of the settings here if that works better

I think it would be easier if you added your commits directly to this branch

SamuelBellomo (dismissed):

fernando-cortez (commented):
The actions feeling a bit snappier is great, but did we actually need to increase the character speed? I'm usually hesitant on modifying character speeds since it introduces a lot of ramifications on the balance of the game. 

For instance, with the new player speed, I can pass over the switch and it looks like I should be able to go through the door opening, since I can reach it much quicker now.

Arrows: have we accounted for their speed(s) since every enemy is much quicker?

Boss Charge: have we bumped up its trample speed/ speed that you get repelled at?

Rogue dash: is this still useful anymore? (I think it may be faster to just walk to the destination given the rampup time)

The most significant thing to consider is the change to the difficulty. Increasing enemy speed and lowering rampup time means you die **extremely** quick. I don't think we've ever received a comment to make the game harder from playtests.

So even just reducing rampup time makes the game harder. I wonder if there's a way we can counter-balance the game and not touch character speed with changes like increasing player HP & reducing enemy damage.

LPLafontaineB:
>The actions feeling a bit snappier is great, but did we actually need to increase the character speed? I'm usually hesitant on modifying character speeds since it introduces a lot of ramifications on the balance of the game.

Good points, I did not see it that way. With the increased speed I though it made the game more easy, but I did not consider how it affected the imps as well. The increase in movement speed was in part thought of as a counter to the increased efficiency of ranged characters, but it definitely causes more issues as well. If we want to keep the increase in movement speed, I think we would need to address the points you raised, by:

- Increasing the speed of arrows and the boss's charge
- reducing the rampup time of the rogue dash/ increasing its distance?
- Increasing player HP
- Reducing enemy damage
- Reducing enemy spawn rate?

If we decide to keep the original character speeds instead, I think we should reduce the damage output of archers/mages. We might still want to buff players hp / debuff enemy damage to make the game easier though

SamuelBellomo:
The original goal is to make things snappier so we don't hide latency issues with too long animations. As long as we don't swear too much while testing and we can finish a game, I'm happy. We already have a bunch of OP abilities in here :P
I wouldn't spend too much time on this, if you're making the too easy, I'd say that's better than too hard.
I agree with all points above, except speed of arrows. this one has edu links (spawned obj for slow arrows, RPCs for mage fire bolt)

SamuelBellomo (dismissed):
I'll trust those last changes, we can come back to this later if there's anything

fernando-cortez:
Just a suggestion, could we split the difference on move speed? Playing the game at 5 feels pretty good, still.

fernando-cortez (approved):

pdeschain (approved):

PR #437, opened 2021-12-23
fix: boss stuck in trampling animation when player becomes invisible during windup

LPLafontaineB:
### Description (*)

This PR adds a transition between the `Action - Boss Trample` and the `Trample Stop` states in the animation controller, which allows the cancellation of the `Trample` action to bring the boss back to the idle animation if the `Trample Stop` trigger is received before the `Trampling` state has started.

TrampleStopTransitionInAnimationController

### Issue Number(s) (*)

Fixes issue(s): [MTT-2016](https://jira.unity3d.com/browse/MTT-2016)
### Manual testing scenarios

1. Reach the boss with a rogue while invisible
2. Use invisibility skill while near the boss (while already invisible)
3. The boss will initiate the trampling action when the player becomes visible, then cancel it when the player returns to invisibility

This might require a few tries to make it work, as sometimes the boss will use its normal attack instead.
### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Models/Animation Controllers/CharacterSetController.controller
+++ b/Assets/BossRoom/Models/Animation Controllers/CharacterSetController.controller
@@ -925,6 +925,7 @@ AnimatorState:
   m_Transitions:
   - {fileID: 8727437690882463675}
   - {fileID: 5784886232730517633}
+  - {fileID: -3605610604413125243}
   m_StateMachineBehaviours: []
   m_Position: {x: 50, y: 50, z: 0}
   m_IKOnFeet: 0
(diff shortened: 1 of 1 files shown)


fernando-cortez (approved):

SamuelBellomo:
Just to confirm, this isn't netcode related and is just an edge case we didn't cover? This happens on host right? (just checking to see if we need education on how to setup animations for multiplayer games)

LPLafontaineB:
> Just to confirm, this isn't netcode related and is just an edge case we didn't cover? This happens on host right? (just checking to see if we need education on how to setup animations for multiplayer games)

Exactly, not related to netcode, just a really specific edge case I discovered by accident

jilfranco-unity (approved):
This looks good to me! :)

PR #438, opened 2022-01-11
feat: move session manager and relay utilities to utilities package

LPLafontaineB:
### Description (*)

This PR moves the SessionManager and RelayUtilities classes to the Utilities package for further use in other projects
### Issue Number(s) (*)

Fixes issue(s): none
### Contribution checklist


Packages/com.unity.multiplayer.samples.coop/Utilities/Net/SessionManager.cs
```diff
@@ -109,9 +109,9 @@ private void Clear()
         /// <param name="clientGUID">This is the clientGUID that is unique to this client and persists accross multiple logins from the same client</param>
         /// <param name="sessionPlayerData">The player's initial data</param>
         /// <returns></returns>
-        public ConnectStatus OnClientApprovalCheck(ulong clientId, string clientGUID, T sessionPlayerData)
+        public bool OnClientApprovalCheck(ulong clientId, string clientGUID, T sessionPlayerData)
         {
-            ConnectStatus gameReturnStatus = ConnectStatus.Success;
+            bool success = true;
```
SamuelBellomo, line 114: It feels like this method is making less sense now :(
Let's go back to what's the intent here, why do we have this method: to check whether session manager has duplicates right? Why not just rename this to "HasDuplicates" or something along those lines. Then I would agree returning a bool makes sense

SamuelBellomo: It does state changes too, so "HasDuplicates" might not be the right name. But you get my drift

SamuelBellomo: much better

SamuelBellomo (dismissed):

pdeschain (dismissed):

SamuelBellomo (approved):

pdeschain (approved):

PR #442, opened 2022-01-12
fix: host name not showing

LPLafontaineB:
### Description (*)

This PR fixes an issue that caused the host's player name to not be shown on the UI. The name is set in PersistentPlayer's OnNetworkSpawn(), using data from the SessionManager. However, this data was initialized after the call to StartHost(), and the PersistentPlayer was spawned during that call. This was introduced by PR #434 which aimed to solve a different issue, because the host data initialization used data from NetworkManager that was not properly initialized.

Here, the initialization of the host session data has been moved to the connection approval callback, so that it happens before the spawning of NetworkBehaviours, but after the server has been properly initialized.

### Related Pull Requests

#434
### Issue Number(s) (*)

Fixes issue(s): [MTT-2157](https://jira.unity3d.com/browse/MTT-2157) & [MTT-2010](https://jira.unity3d.com/browse/MTT-2010)
### Manual testing scenarios

For MTT-2157:
1. Start a game as a host
2. Select a character, then load the BossRoom scene
3. See that the host's name is now visible on the UI

For MTT-2010:
1. Start a game as a host
2. Join the game as a client
3. As the client, disconnect while still in the character selection scene
4. As the former client, try to host a new game
### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/GameNetPortal.cs
+++ b/Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/GameNetPortal.cs
@@ -242,8 +242,8 @@ namespace Unity.Multiplayer.Samples.BossRoom
 
         void StartHost()
         {
+            SessionManager<SessionPlayerData>.Instance.AddHostData(new SessionPlayerData(NetManager.ServerClientId, PlayerName, m_AvatarRegistry.GetRandomAvatar().Guid.ToNetworkGuid(), 0, true));
             NetManager.StartHost();
-            SessionManager<SessionPlayerData>.Instance.AddHostData(new SessionPlayerData(NetManager.LocalClientId, PlayerName, m_AvatarRegistry.GetRandomAvatar().Guid.ToNetworkGuid(), 0, true));
         }
 
         /// <summary>
--- a/Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/GameNetPortal.cs
+++ b/Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/GameNetPortal.cs
@@ -242,7 +242,6 @@ namespace Unity.Multiplayer.Samples.BossRoom
 
         void StartHost()
         {
-            // First initialize the host data so that objects that need it during their initial spawn can use it, using ServerClientId as the host's ClientId
             NetManager.StartHost();
         }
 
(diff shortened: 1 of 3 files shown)


fernando-cortez (approved):
Works when disconnecting in CharSelect scene, but fails when shutting down while in BossRoom scene.

I'm positive it's not related to this PR, but instead to the soft sync issue noted [here](https://jira.unity3d.com/browse/MTT-2159). Looking into that at the moment.

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/GameNetPortal.cs
```diff
@@ -242,8 +242,9 @@ public async void StartUnityRelayHost()
 
         void StartHost()
         {
+            // First initialize the host data so that objects that need it during their initial spawn can use it, using ServerClientId as the host's ClientId
+            SessionManager<SessionPlayerData>.Instance.AddHostData(new SessionPlayerData(NetManager.ServerClientId, PlayerName, m_AvatarRegistry.GetRandomAvatar().Guid.ToNetworkGuid(), 0, true));
```
SamuelBellomo, line 246: Do we have a guarantee that ServerClientId will be the same before and after StartHost?

SamuelBellomo: In the sense --> does the SDK team commit to keeping this behaviour in the future?

LPLafontaineB: Currently it's a readonly property. I don't know if the SDK team commit to keep it that way, but I don't see why that would change. I'll ask just to be sure

LPLafontaineB: I was told that it was not safe to assume that it would not change, so I'll be implementing a different solution

SamuelBellomo (approved):

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/SessionManager.cs
```diff
@@ -179,7 +186,7 @@ public ConnectStatus OnClientApprovalCheck(ulong clientId, string clientGUID, T
                 return GetPlayerData(clientGUID);
             }
 
-            Debug.Log("No client guid found mapped to the given client ID");
+            Debug.LogError("No client guid found mapped to the given client ID");
```
fernando-cortez, line 189: For this and the other two error logs below, I'd also log the client Id passed in.

fernando-cortez (approved):

SamuelBellomo (approved):

PR #444, opened 2022-01-13
feat: cheat window

LPLafontaineB:
### Description (*)

This PR replaces the keyboard hotkeys with a popup window for debug cheats. The window can be opened or closed by the host or a client, using the `/` key, or with 4 fingers on a mobile device. The three cheats that we currently have can now be accessed via buttons.
The window and its cheats are only accessible in-editor and in development builds.

DebugCheatsPopupWindowV2

### Issue Number(s) (*)

Fixes issue(s): [MTT-1994](https://jira.unity3d.com/browse/MTT-1994)
### Manual testing scenarios

1. Start a game
2. Select a character and go to BossRoom scene
3. On host or client, press `/` key to open the debug cheats window
4. Press any button to test the corresponding cheat
### Contribution checklist


Assets/BossRoom/Prefabs/UI/BossRoomHudCanvas.prefab
```diff
@@ -1286,7 +1286,7 @@ RectTransform:
   m_Children:
```
fernando-cortez, line 1286: Just a general nitpicky cleanup comment: for UI elements, keep "Raycast Target" enabled only for the necessary components. Some text items, and nested images can have that field disabled. TextMeshPro components have these hidden under "Extra Settings" btw.

Assets/BossRoom/Scripts/Client/UI/UIDebugCheats.cs
```diff
@@ -0,0 +1,46 @@
...
+using Unity.Netcode;
+using UnityEngine;
+
+#if UNITY_EDITOR || DEVELOPMENT_BUILD
+namespace Unity.Multiplayer.Samples.BossRoom.Client
+{
+    public class UIDebugCheats : MonoBehaviour
+    {
+        [SerializeField]
+        DebugCheatsMediator m_DebugCheatsMediator;
```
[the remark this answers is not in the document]
LPLafontaineB: Great idea! It is much cleaner now, and I could remove the DebugCheatsMediator altogether!

Assets/BossRoom/Scripts/Client/UI/UIDebugCheats.cs
```diff
@@ -0,0 +1,56 @@
+using System;
+using Unity.Netcode;
```
fernando-cortez, line 2: Can remove now 👍.

fernando-cortez (commented):
Just needs an unused namespace removed, but LGTM!

fernando-cortez (dismissed):

Assets/BossRoom/Scripts/Server/DebugCheatsManager.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        public void SpawnBoss()
+        {
+            SpawnBossServerRpc();
+        }
+
+        public void GoToPostGame()
+        {
+            GoToPostGameServerRpc();
+        }
```
SamuelBellomo, line 35: do you need those wrappers? Couldn't you call the RPCs themselves?

Assets/BossRoom/Scripts/Server/DebugCheatsManager.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        [ServerRpc(RequireOwnership = false)]
+        void SpawnEnemyServerRpc(ServerRpcParams serverRpcParams = default)
```
SamuelBellomo, line 38: Since your call sites for these RPCs don't use serverRpcParams, I believe you can just remove "default"

Assets/BossRoom/Scripts/Server/DebugCheatsManager.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        [ServerRpc(RequireOwnership = false)]
+        void GoToPostGameServerRpc(ServerRpcParams serverRpcParams = default)
+        {
+            NetworkManager.SceneManager.LoadScene("PostGame", LoadSceneMode.Single);
+            LogCheatUsedClientRPC(serverRpcParams.Receive.SenderClientId, "GoToPostGame");
+        }
+
+        [ClientRpc]
+        void LogCheatUsedClientRPC(ulong clientId, string cheatUsed)
```
SamuelBellomo, line 61: (not a review comment) lol the snitching method :P

Assets/BossRoom/Scripts/Client/UI/UIDebugCheats.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace Unity.Multiplayer.Samples.BossRoom.Client
+{
+    /// <summary>
+    /// Provides a way to open/close a popup window and GameEvents to raise debug cheat commands to be caught by
+    /// listeners. This class is only available in the editor or for development builds
+    /// </summary>
+    public class UIDebugCheats : MonoBehaviour
+    {
+        [SerializeField]
+        GameObject m_DebugCheatsPanel;
```
SamuelBellomo, line 14: My understanding is you're putting all those events in between because of the client/server separation correct?
What do you think of putting everything in Shared and hooking up the buttons directly to the RPCs? It'd make things way simpler and we could put this cheat code in a self contained "cheat assembly", something like that

Assets/BossRoom/Scripts/Server/DebugCheatsManager.cs
LPLafontaineB: The `UnityEventGameEventListener`s uses parameterless `UnityEvent`s, so I could not use the RPC calls directly

Assets/BossRoom/Scripts/Server/DebugCheatsManager.cs
LPLafontaineB: They do use it, for the SenderClientId. This is used for the "snitch log", but it is required anyways for the god mode cheat, if we want to apply it to only a single player.

Assets/BossRoom/Scripts/Client/UI/UIDebugCheats.cs
LPLafontaineB: > My understanding is you're putting all those events in between because of the client/server separation correct?
Yes, that and to keep the UI and logic separate.

> What do you think of putting everything in Shared and hooking up the buttons directly to the RPCs?
It would work for the cheats that are in this PR, but we would lose access to everything in the Server assembly. This could be a problem for some new cheats as well. For example, my current implementation of God Mode uses PlayerServerCharacter to find the character on which to apply the toggle. I could move it to a new assembly with a dependency on Server and Shared, though. It would remove the need for events

Assets/BossRoom/Scripts/Server/DebugCheatsManager.cs
SamuelBellomo: I see, that's how C# method signature works :( 
nvm

Assets/BossRoom/Scripts/Server/DebugCheatsManager.cs
SamuelBellomo: aaand that's what happens when you read too fast.
nvm

SamuelBellomo (dismissed):

LPLafontaineB:
Updated PR description with screenshot of new window with instructions, which is now active at the start of the game.

SamuelBellomo (approved):

fernando-cortez (commented):
Just needs StreamingAssets meta file removed, but gtg.

fernando-cortez (approved):

SamuelBellomo (approved):

PR #453, opened 2022-01-25
fix: cannot replicate network life state

LPLafontaineB:
### Description (*)

This PR removes the preprocessor directive around the IsGodMode NetworkVariable. This caused an issue when playing with a host on the editor/dev build and a client with a regular build (or vice-versa) as the NetworkLifeState could not be proprely synchronized because it did not have the same NetworkVariables in both instances.
### Issue Number(s) (*)

Fixes issue(s):
### Manual testing scenarios

1. Create a regular build
2. Start hosting on the build
3. Join with a client on the editor (or a development build)
4. Proceed to the BossRoom scene
5. See the issue is no longer there
### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Scripts/Shared/NetworkLifeState.cs
+++ b/Assets/BossRoom/Scripts/Shared/NetworkLifeState.cs
@@ -13,11 +13,9 @@ namespace Unity.Multiplayer.Samples.BossRoom
 
         public NetworkVariable<LifeState> LifeState => m_LifeState;
 
-#if UNITY_EDITOR || DEVELOPMENT_BUILD
         /// <summary>
         /// Indicates whether this character is in "god mode" (cannot be damaged).
         /// </summary>
         public NetworkVariable<bool> IsGodMode { get; } = new NetworkVariable<bool>(false);
-#endif
     }
 }


Assets/BossRoom/Scripts/Shared/NetworkLifeState.cs
```diff
@@ -13,11 +13,9 @@ public class NetworkLifeState : NetworkBehaviour
 
         public NetworkVariable<LifeState> LifeState => m_LifeState;
 
-#if UNITY_EDITOR || DEVELOPMENT_BUILD
```
SamuelBellomo, line 16: mmm so that's interesting.
1) education: It might be worth adding this behaviour to our doc.
It kind of makes sense, we were basically trying to have two different replication protocols talk to each other, one including a netvar the other without it.
2) dogfooding: What was the error? Connection time --> impossible to connect due to Netcode saying there's a diff between protocols? Or was it at runtime while trying to replicate it? Instinctively I would have liked to see this ideally at connection time or at least at scene switch time (like a soft sync error).
3) PR improvement: we could just not allow release builds connecting to debug builds and require both to be the same version. It makes sense, since we wouldn't have the same protocols in both cases.
OR we surround this in an if check (Debug.isDebugBuild) and make sure this var isn't used in releases.
OR we just leave it, some games will have cheats available even in release builds.

LPLafontaineB: 1. Definitely. I think this page would be a good candidate for that: https://docs-multiplayer.unity3d.com/docs/develop/basics/networkobject
2. It happened at runtime, giving errors when trying to spawn the characters because of differing `GlobalObjectIdHash`, and failing to obtain the correct `NetworkAvatarGuid`. Maybe it could be caught at connection time by comparing the registered `NetworkObjects`' `GlobalObjectIdHash`?
3. The first option makes sense, we could handle that during connection approval. As for the second option, I suppose I could add that to the getter, so that it always returns false on regular builds. Anyways, there are still preprocessor directives everywhere this cheat is used, so it shouldn't be an issue, we would simply have a useless NetworkVariable that isn't called from anywhere. The third option could work too, but I would make the cheat window initially hidden on release builds.

PR #455, opened 2022-01-27
fix: cheat window toggling on mobile

LPLafontaineB:
### Description (*)

This PR adds a verification that one of the four touches is in the TouchPhase.Began phase before toggling the cheat window on mobile, so that it only toggles once.
### Issue Number(s) (*)

Fixes issue(s): [MTT-2279](https://jira.unity3d.com/browse/MTT-2279)
### Manual testing scenarios

1. Build the game on a mobile platform (tested on Android), with a dev build
2. Progress to the BossRoom scene
3. Use four fingers simultaneously touching the screen to open/close the window
4. See that it only does so once
### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Scripts/DebugCheats/DebugCheatsManager.cs
+++ b/Assets/BossRoom/Scripts/DebugCheats/DebugCheatsManager.cs
@@ -31,11 +31,23 @@ namespace Unity.Multiplayer.Samples.BossRoom.Debug
 
         void Update()
         {
-            if (Input.touchCount == k_NbTouchesToOpenWindow ||
+            if (Input.touchCount == k_NbTouchesToOpenWindow && AnyTouchDown() ||
                 m_OpenWindowKeyCode != KeyCode.None && Input.GetKeyDown(m_OpenWindowKeyCode))
             {
                 m_DebugCheatsPanel.SetActive(!m_DebugCheatsPanel.activeSelf);
             }
+        }
+
+        static bool AnyTouchDown()
+        {
+            foreach (var touch in Input.touches)
+            {
+                if (touch.phase == TouchPhase.Began)
+                {
+                    return true;
+                }
+            }
+            return false;
         }
 
         public void SpawnEnemy()


SamuelBellomo (approved):

fernando-cortez (commented):
So while this is definitely an improvement, I'm noticing some unwanted behaviour.

Basically, if you're targeting a PC/NPC and press do the 4 finger press, it almost always tries to attack the target (or move and then attack in the case of a melee).

If there isn't a target highlighted, your player is prompted to move to the location of where one of your fingers touched the screen. Not the best outcome when you want to do a cheat in isolation.

So I'm curious if this touchscreen-only action should be reworked so that it can't ever invoke one of the game-side actions. Thoughts? @SamuelBellomo @LPLafontaineB

LPLafontaineB:
> So while this is definitely an improvement, I'm noticing some unwanted behaviour.
> 
> Basically, if you're targeting a PC/NPC and press do the 4 finger press, it almost always tries to attack the target (or move and then attack in the case of a melee).
> 
> If there isn't a target highlighted, your player is prompted to move to the location of where one of your fingers touched the screen. Not the best outcome when you want to do a cheat in isolation.
> 
> So I'm curious if this touchscreen-only action should be reworked so that it can't ever invoke one of the game-side actions. Thoughts? @SamuelBellomo @LPLafontaineB

I saw this too. By default, Input.SimulateMouseWithTouches is true, which interprets single touches into left clicks and double touches into right clicks. To prevent this issue from happening, I think it would require us to turn this off and to manually handle touch inputs in our input manager. Or I suppose we could also simply cancel all actions when opening the cheat window

SamuelBellomo:
> So while this is definitely an improvement, I'm noticing some unwanted behaviour.
> 
> Basically, if you're targeting a PC/NPC and press do the 4 finger press, it almost always tries to attack the target (or move and then attack in the case of a melee).
> 
> If there isn't a target highlighted, your player is prompted to move to the location of where one of your fingers touched the screen. Not the best outcome when you want to do a cheat in isolation.
> 
> So I'm curious if this touchscreen-only action should be reworked so that it can't ever invoke one of the game-side actions. Thoughts? @SamuelBellomo @LPLafontaineB

@fernando-cortez @LPLafontaineB agreed, however I wouldn't block this PR for this. Do you mind creating a bug for this? In past projects, we'd use something like this https://answers.unity.com/questions/1482927/how-to-make-ui-block-clicks-from-triggering-things.html
This would block calls to ScreenPointToRay. Dunno how that would work with other interactions though.

fernando-cortez (approved):
Created [MTT-2389](https://jira.unity3d.com/browse/MTT-2389) to track the improvements we have mentioned.

PR #457, opened 2022-02-01
fix: cannot host a game after having previously played as a client

LPLafontaineB:
### Description (*)

This PR makes it so only the server adds entries to the m_ClientSceneMap by registering the callback handling this only on the server. This would previously cause issues since the OnClientSceneChanged is also called on clients. Adding entries on clients doesn't make sense, and causes issues when trying to host a new game afterwards. This new entry prevents spawning the characters in the BossRoom scene, which is triggered when all clients are in the scene, based on this map. So if this map has another entry, it will always be in the wrong scene and the spawn will never be triggered.
This PR also prevents clients from updating the static list of active PlayerServerCharacters, which used to result in a null reference exception when starting a new game in this scenario.
### Issue Number(s) (*)

Fixes issue(s): [MTT-2373](https://jira.unity3d.com/browse/MTT-2373)
### Manual testing scenarios

1. Start two instances of the game
2. Host on instance 1 and join with instance 2
3. End the game (either by quitting as instance 1 or by going to the post-game scene and exiting to main menu)
4. Host a new game on instance 2
5. Join with instance 1
6. Proceed through character selection to the BossRoom scene
7. See the player characters properly spawning and the game starting normally

### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ServerGameNetPortal.cs
+++ b/Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ServerGameNetPortal.cs
@@ -93,7 +93,10 @@ namespace Unity.Multiplayer.Samples.BossRoom.Server
 
         public void OnClientSceneChanged(ulong clientId, int sceneIndex)
         {
-            m_ClientSceneMap[clientId] = sceneIndex;
+            if (m_Portal.NetManager.IsServer)
+            {
+                m_ClientSceneMap[clientId] = sceneIndex;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ServerGameNetPortal.cs
```diff
@@ -93,7 +93,10 @@ private void OnClientDisconnect(ulong clientId)
 
         public void OnClientSceneChanged(ulong clientId, int sceneIndex)
         {
-            m_ClientSceneMap[clientId] = sceneIndex;
+            if (m_Portal.NetManager.IsServer)
```
SamuelBellomo, line 96: This feels like the wrong fix, we should fix the root cause. We're in ServerGameNetPortal, how come this callback is registered/called at all on clients??

LPLafontaineB: You're right, I corrected this, preventing the callback from being registered on clients and making sure it was un-registered when exiting a game. I also found another bug caused by the same scenario and included a fix for that as well. I updated the PR description to reflect those changes

SamuelBellomo (approved):

fernando-cortez (approved):

PR #460, opened 2022-02-02
Sam/fix/fixing issue where default ip wouldnt appear

SamuelBellomo:
### Description (*)
Fixing issue where default IP wouldn't appear when coming back to main menu after a playthrough
Fixing wrong error message


LPLafontaineB:
With this change, when opening the popup panel, we no longer see the instructions label, unless we move to a different connection mode and move back to direct IP.

A simpler fix would instead be to only add `OnlineMode = OnlineMode.IpHost;` to the `SetupEnterGameDisplay` method. This, if the OnlineMode is set to `Unset` in the `Reset` method, triggers `OnOnlineModeChanged()` and correctly sets up the display.

LPLafontaineB (commented):
With this change, when opening the popup panel, we no longer see the instructions label, unless we move to a different connection mode and move back to direct IP.

A simpler fix would instead be to only add `OnlineMode = OnlineMode.IpHost;` to the `SetupEnterGameDisplay` method. This, if the OnlineMode is set to `Unset` in the `Reset` method, triggers `OnOnlineModeChanged()` and correctly sets up the display.

LPLafontaineB:
Tested the fix I suggested and it seems to resolve [MTT-2465](https://jira.unity3d.com/browse/MTT-2465)

PR #477, opened 2022-02-10
feat: tool to enable monobehaviours only on server or client

LPLafontaineB:
### Description (*)

This PR adds a NetworkBehaviour that can be placed on a NetworkObject to make sure certain MonoBehaviours can only be enabled on active servers or active clients by using specific interfaces that those MonoBehaviours would inherit. This would remove that responsibility from those MonoBehaviours. It would also allow us to make some of them not be NetworkBehaviours anymore since it was mostly used for the OnNetworkSpawn to enable/disable themselves.

As an example, I have included in this draft PR an implementation of this feature with the interactive door.

Some pros of this solution:

- Provides a unified solution for something we currently need to do in a lot of different places
- Allows us to change some NetworkBehaviours back to just MonoBehaviours
- Provides a single point on which to configure which MonoBehaviours are executed on clients and which are executred on servers

Cons:

- Adds weight to the actual prefabs and in-scene GameObjects
- Needs us to make sure that all MonoBehaviours that should only be enabled on a client/server implement the correct interface and that this NetworkBehaviour is added to every prefab and in-scene GameObject that uses it

### Issue Number(s) (*)

Fixes issue(s): When implemented, should fix [MTT-1683](https://jira.unity3d.com/browse/MTT-1683) & [MTT-2433](https://jira.unity3d.com/browse/MTT-2433)
### Manual testing scenarios

1. ...
2. ...
### Contribution checklist


Assets/BossRoom/Scripts/Shared/OnSpawnBehaviourEnabler.cs
```diff
@@ -0,0 +1,69 @@
+using System;
+using System.Collections.Generic;
+using Unity.Netcode;
+using UnityEngine;
+
+namespace Unity.Multiplayer.Samples.Utilities
```
LPLafontaineB, line 6: This file will be moved to the Utilities package when the PR will leave the draft status. I kept it here to make testing and iterating easier since ParrelSync doesn't sync the packages folder automatically

LPLafontaineB:
The issue this PR fixes could also be the cause of [MTT-2475](https://jira.unity3d.com/browse/MTT-2475), but hard to verify since we don't have consistent reproduction steps for that bug yet

Assets/BossRoom/Scripts/Shared/OnSpawnBehaviourEnabler.cs
```diff
@@ -0,0 +1,69 @@
...
+{
+    public interface IClientOnlyMonoBehaviour
```
SamuelBellomo, line 8: So if those interfaces are tightly coupled to monobehaviour and netcode, any reason we wouldn't just have an abstract class that inherits from NetworkBehaviour and has SetEnabled as abstract method?

Assets/BossRoom/Scripts/Shared/OnSpawnBehaviourEnabler.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            // Disable everything here to prevent those MonoBehaviours to be updated before this one is spawned.
+            DisableAll();
+        }
+
+        public override void OnNetworkSpawn()
+        {
+            if (IsClient)
+            {
+                foreach (var behaviour in gameObject.GetComponents<IClientOnlyMonoBehaviour>())
```
SamuelBellomo, line 35: OnNetworkSpawn would be called with network pooling correct? This has the potential to be a very costly spawn, GetComponent isn't performant

SamuelBellomo: especially if there's multiple multiple spawns with pooling

Assets/BossRoom/Scripts/Shared/OnSpawnBehaviourEnabler.cs
LPLafontaineB: An earlier version of this used inheritance, with a single class inheriting NetworkBehavior with a protected field telling if it was meant for a server or a client. The disabling/enabling was handled by that abstract class in its Awake, OnSpawn and OnDespwn events.
The advantage was that we did not need a second script to enable/disable it, but it came with more restrictions that I wasn't really happy with. First, we would need to make sure to call the base methods inside those three methods if we wanted to override it in daughter classes. Also, it would mean that if a specific class needs that behaviour, all its parents would have to have it too, which might not make sense. i.e. class A is a client-only netbehaviour inheriting from class B that doesn't need to know if it is on a client or on a server. In this case, if we want to restrict A to only clients, we would have to have B inherit this abstract class, even though B doesn't require it.

Assets/BossRoom/Scripts/Shared/OnSpawnBehaviourEnabler.cs
LPLafontaineB: Oh you're right, that would be super costly. Maybe I should cache all of this in its Awake method? Another solution could be to expose a list to the editor, but that would require us to manually enter all netbehaviours, but that would be even more work than what this tool is trying to simplify

SamuelBellomo: yeah caching would be good

Assets/BossRoom/Scripts/Shared/OnSpawnBehaviourEnabler.cs
SamuelBellomo: I see. Yeah makes sense. I wonder if there could be a way to dynamically add the enabler without requiring us to remember to add it. It's too bad "RequireComponent" doesn't work on interfaces.
Or we could have some editor script magic with some reflection...?

SamuelBellomo: @LukeStampfli @pdeschain @fernando-cortez what do you guys think?

LPLafontaineB: We could add "RequireComponent" to the classes implementing these interfaces themselves I suppose, but it's still not as plug-and-play as I hoped it could be

LPLafontaineB: Thinking back on it, would adding this recommendation as remarks in the interface descriptions be sufficient?

LukeStampfli: I think using an interface here is better than using inheritance. With the updated xml docs on the interfaces I think it is easy enough to understand how to use this.

LukeStampfli (approved):

SamuelBellomo:
on hold until we have true solution

SamuelBellomo:

PR #481, opened 2022-02-14
fix: connecting while starting game

LPLafontaineB:
### Description (*)

This PR makes it so the closing of character selection and starting of a new game is cancelled when a new player joins or when a player quits. Players now have to re-confirm their choice when it happens so as to be properly informed of it.
This PR also cleans up ServerCharSelectState to align with our coding standards.
### Issue Number(s) (*)

Fixes issue(s): [MTT-2358](https://jira.unity3d.com/browse/MTT-2358)
### Manual testing scenarios

1. Start a game and go to character selection with one or more instances of the game
2. On a second instance, join the game just after the last player hit ready
3. See that the closing of character selection is properly cancelled and every player has to confirm their character again
4. Confirm on all of them
5. Just after the last player confirms, quit the session with one of the clients
6. See that the closing of character selection is properly cancelled again
### Contribution checklist


Assets/BossRoom/Scripts/Server/Game/State/ServerCharSelectState.cs
```diff
@@ -240,6 +271,12 @@ private void OnClientDisconnectCallback(ulong clientId)
                     break;
                 }
             }
+
+            // If lobby is closing and waiting to start the game, cancel so that other players can react to it
+            if (CharSelectData.IsLobbyClosed.Value)
```
fernando-cortez, line 276: One last condition to consider here is if:
- Player A (host) is ready
- Player B is ready
- Player C disconnects

Outcome is Player A and B are stuck and one of them has to unready and ready back up.

Either all are set to unready when a disconnect happens, or the server should just send the readied players into the game, since the condition is met. I favour the latter scenario.

fernando-cortez: Also if you want to tackle that in a follow-up PR, that works too.

LPLafontaineB: Good catch! I didn't consider this use-case. 

For the rest of this PR, I chose the former, with setting everyone to unready when the lobby closing is cancelled. I remember most games with lobbys I've recently played having a similar mechanic, which allows users to react to that situation, like. waiting for their friend to come back before hitting ready again. But if the unready animation and the disappearance of a player are not enough of a visual cue to make sure users understand what is happening this might just create more confusion.

On the other hand, simply performing the check as a player leaves and directly moving to the next scene could be simpler, but it feels weird to not give them the choice to cancel to wait for the other player to come back. Although since we support late joining it might not really be an issue

fernando-cortez: Makes sense to let the players inside the lobby to wait for the disconnected player!

fernando-cortez: It'd be nice if we had messages that consumed themselves (ie. just faded out on their own). I feel like adding this for every player disconnect might lead to a lot of stacking pop-up messages and/or just busy UI since users would have to clear the pop up message. 

I'd say let's leave it without and see if we need to add it in, in another PR.

fernando-cortez (dismissed):

Assets/BossRoom/Scripts/Server/Game/State/ServerCharSelectState.cs
fernando-cortez: Want to create a ticket for these transient, self-consumable messages?

SamuelBellomo (approved):

PR #482, opened 2022-02-15
feat: flag to prevent dev and release builds to connect

LPLafontaineB:
### Description (*)

This PR adds a flag to the connection payload telling if the game instance is a dev build or not. The connection approval then uses it to deny connections with different values for that flag. Then, they get an error message telling them they are using incompatible build types.
This PR also simplifies and clarifies the connection approval flow.
### Issue Number(s) (*)

Fixes issue(s): [MTT-2521](https://jira.unity3d.com/browse/MTT-2521)
### Manual testing scenarios

1. Start a host in editor or in a dev build
2. Try to join as a client in a release build
3. See the error
4. The reverse scenario should also produce the same error
### Contribution checklist


Assets/BossRoom/Scripts/Client/UI/MainMenuUI.cs
```diff
@@ -178,6 +178,9 @@ private void ConnectStatusToMessage(ConnectStatus status, bool connecting)
                     var text = connecting ? "Something went wrong" : "The connection to the host was lost";
                     m_ResponsePopup.SetupNotifierDisplay(title, text, false, true);
                     break;
+                case ConnectStatus.IncompatibleBuildType:
+                    m_ResponsePopup.SetupNotifierDisplay("Connection Failed", "Server and client builds are not compatible. You cannot connect a release build to a development build or an in-editor session", false, true);
```
fernando-cortez, line 182: Just a comment to add a trailing period to this and to the other messages for uniformity.

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ServerGameNetPortal.cs
```diff
@@ -174,6 +174,15 @@ private void ApprovalCheck(byte[] connectionData, ulong clientId, NetworkManager
             string payload = System.Text.Encoding.UTF8.GetString(connectionData);
             var connectionPayload = JsonUtility.FromJson<ConnectionPayload>(payload); // https://docs.unity3d.com/2020.2/Documentation/Manual/JSONSerialization.html
 
+            if (connectionPayload.isDebug != Debug.isDebugBuild)
+            {
+                gameReturnStatus = ConnectStatus.IncompatibleBuildType;
```
fernando-cortez, line 179: I'm curious if we have to still invoke the connectionApprovalCallback in the cases where we're denying the client's connection. This is me just thinking that perhaps these are the cases where the client can't host afterwards? If there's an explicit reason why, I suppose I'd like to know why (and users would too).

LPLafontaineB: I asked around and it turns out the WaitToDisconnect coroutine is a workaround we had to do to be able to send a disconnect reason message to the client, as if they don't receive messages after they are disconnected/their connection is denied. A ticket has been created for that on the sdk side.

So I will use the same workaround, but add comments to make it clearer.

As for the bug preventing clients to host afterwards, I tested this scenario to be sure and did not encounter issues. The bug is really a UI issue, caused by the same thing that PR #460 is looking to solve

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ServerGameNetPortal.cs
```diff
@@ -153,75 +154,75 @@ private void ApprovalCheck(byte[] connectionData, ulong clientId, NetworkManager
                 return;
             }
 
-            ConnectStatus gameReturnStatus;
+            var payload = System.Text.Encoding.UTF8.GetString(connectionData);
+            var connectionPayload = JsonUtility.FromJson<ConnectionPayload>(payload); // https://docs.unity3d.com/2020.2/Documentation/Manual/JSONSerialization.html
```
LPLafontaineB, line 158: A small change introduced in this PR is that the connection data is deserialized earlier, which means it now happens in the case where the game is at maximum capacity while it did not before.
While profiling, I measured the time it takes for that to be less than 0.3 ms in the editor on my machine with deep profile activated, which is small but not negligible.
I chose to leave it that way anyways because I feel it makes the code simpler and easier to follow.

Assets/BossRoom/Scripts/Client/UI/MainMenuUI.cs
fernando-cortez: I think the case for `ConnectStatus.GenericDisconnect` needs a punctuation pass. Approval once done.

LPLafontaineB: Done

fernando-cortez (dismissed):

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ServerGameNetPortal.cs
```diff
@@ -178,67 +180,79 @@ void ApprovalCheck(byte[] connectionData, ulong clientId, NetworkManager.Connect
...
+        ConnectStatus GetConnectStatus(ConnectionPayload connectionPayload)
+        {
+            ConnectStatus gameReturnStatus;
+
+            // Test for over-capacity connection. This needs to be done asap, to make sure we refuse connections asap
+            // and don't spend useless time server side on invalid users trying to connect
+            // todo this is currently still spending too much time server side.
+            if (m_Portal.NetManager.ConnectedClientsIds.Count >= CharSelectData.k_MaxLobbyPlayers)
+            {
+                gameReturnStatus = ConnectStatus.ServerFull;
```
SamuelBellomo, line 224: having an early return here would reduce the amount of nested ifs we have

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ServerGameNetPortal.cs
SamuelBellomo: is this comment still up to date? looking at current implementation, looks like the server full check is still pretty early

SamuelBellomo (approved):

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ServerGameNetPortal.cs
LPLafontaineB: yes. The main difference here is that before that, we would do the IsServerFull check before deserializing the connecting data. Although, since we now also verify the player capacity in Lobby before starting the connection, it should not really matter.

LPLafontaineB: If you are in agreement with it, I will remove the comment in the code about it at the same time

LPLafontaineB: Tested again to be sure and it still works fine

SamuelBellomo (approved):

fernando-cortez (approved):

PR #483, opened 2022-02-16
fix: SceneBootstrapper detects and allows TestRunner launches

fernando-cortez:
### Description (*)
This PR unblocks TestRunner Playmode tests from executing, since `SceneBootstrapper` immediately launched a Playmode session into `Startup` scene.

`SceneBootstrapper`, on a playmode state change, detects if the scene loaded is a TestRunner scene and if so, does not perform any loading logic.

Remarks: `EditorSceneManager.playModeStartScene` was investigated as an avenue to curb this issue, as well as to simplify `SceneBootstrapper`. However, once `EditorSceneManager.playModeStartScene` is set, there is no way through editor callbacks to prevent the editor to not launch said scene in the cases of hitting the play button or when TestRunner launches a scene. Therefore, it makes sense for now to let `SceneBootstrapper` detect the scene launched and determine what logic to execute. A [ticket](https://jira.unity3d.com/browse/DSTP-1082) has been generated to DSTP to improve this in the future.

### Issue Number(s) (*)

Fixes issue(s): [MTT-2370](https://jira.unity3d.com/browse/MTT-2370)
### Manual testing scenarios

1. Open TestRunner from the editor and launch any of the example Playmode tests.
2. Tests will execute and return a fail or success result.
### Contribution checklist


Assets/BossRoom/Scripts/Editor/SceneBootstrapper.cs
```diff
@@ -34,6 +34,8 @@ public class SceneBootstrapper
         const string k_LoadBootstrapSceneOnPlay = "Boss Room/Load Bootstrap Scene On Play";
         const string k_DoNotLoadBootstrapSceneOnPlay = "Boss Room/Don't Load Bootstrap Scene On Play";
 
+        const string k_TestRunnerObjectName = "Code-based tests runner";
```
SamuelBellomo, line 37: That seems like the kind of thing that would break super easily. In 6 months, someone will remove the "s" from "tests" and we'll be wondering why our tests are breaking all of a sudden.
Couldn't you set some static "isTesting" value in your test fixture setup?
Or detect command line args typical for testing?
Or

fernando-cortez: So coming back to this, the reason this check exists is that TestRunner API is currently all internal. I'll point you to this [ticket](https://jira.unity3d.com/browse/DSTP-1082) I've raised with the Test Platform team: it will be available in a future release.

For now, this check should suffice IMO. Back to the conversation Phil initially raised, methods with `InitializeOnLoadMethod` attribute are invoked after the TestRunner scene is created & launched. My thought was also to set some `IsTesting` flag inside those or even on methods with [Setup] or [UnitySetUp], but those two are invoked per test, again much after the test scene has been launched.

SamuelBellomo: ah I see
how about OneTimeSetup, does that work or is that too late as well?
https://docs.unity3d.com/Packages/com.unity.test-framework@1.1/manual/reference-actions-outside-tests.html

And you can't read command line args like System.Environment.CommandLineArgs? (no need to go through test's API)

fernando-cortez: After testing most of the attributes on that page, they're all far too late in the process to be able to catch inside of SceneBootstrapper.

Order is:
* Launching a test through the TestRunner window creates "InitTestScene" with a timestamp. Inside that scene is a test GameObject (I was using this object as the check previously).
* ExitEditMode (this is when we launch Startup scene), EnterPlayMode
* Tests are executed (so here any class/methods with attributes would be invoked).

And no, command line args unfortunately do not reveal anything about TestRunner when launched through the TestRunner window.

SamuelBellomo (approved):

LPLafontaineB (approved):

PR #488, opened 2022-02-24
feat: replace guid with auth playerid for session management

LPLafontaineB:
### Description (*)

This PR replaces our custom GUID it with the PlayerId obtained from UGS's AuthenticationService when signed in. When not signed in, if using direct IP connection, it will use our custom GUID. It also removes the workaround we had to allow clients with the same GUID to connect on debug builds, since we are using profiles for that. Now the arguments used to set profiles are also used to update our custom GUID when connecting to a host to differentiate each client if using multiple builds or parrelsync clones.
### Issue Number(s) (*)

Fixes issue(s): [MTT-2640](https://jira.unity3d.com/browse/MTT-2640)
### Manual testing scenarios

1. Start a game with two or more clients
2. Disconnect each client
3. Reconnect to the game with each client
4. See that each client is reconnected with the right character and the right data
5. Do it again through direct IP while signed in and not signed in.
### Contribution checklist


LukeStampfli (approved):

Assets/BossRoom/Scripts/Shared/ProfileManager.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    static bool s_IsProfileSet = false;
+
+    static string s_Profile = "";
+
+    public static string Profile
+    {
+        get
+        {
+            if (!s_IsProfileSet)
```
SamuelBellomo, line 19: doing that bool check might be more performant, but do we get Profile that often?
maintainability side, having a != "" might be better? Or even != null

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ClientGameNetPortal.cs
```diff
@@ -18,6 +17,8 @@ public class ClientGameNetPortal : MonoBehaviour
         public static ClientGameNetPortal Instance;
         private GameNetPortal m_Portal;
 
+        string m_JoinCode;
```
SamuelBellomo, line 20: I thought we already had this cached somewhere? (multiple caches of the same value is just asking for desyncs later)

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ClientGameNetPortal.cs
```diff
@@ -177,40 +178,46 @@ public void StartClient(string ipaddress, int port)
...
-                await clientRelayUtilityTask;
-                var (ipv4Address, port, allocationIdBytes, connectionData, hostConnectionData, key) = clientRelayUtilityTask.Result;
+            ConnectClient(onFailure);
+        }
 
-                m_LobbyServiceFacade.UpdatePlayerRelayInfoAsync(allocationIdBytes.ToString(), joinCode, null, null);
-                utp.SetClientRelayData(ipv4Address, port, allocationIdBytes, key, connectionData, hostConnectionData, isSecure: true);
-            }
-            catch (Exception e)
+        async void ConnectClient(Action<string> onFailure)
```
SamuelBellomo, line 196: onFailure isn't used?

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/GameNetPortal.cs
```diff
@@ -87,6 +88,8 @@ public class GameNetPortal : MonoBehaviour
         private LocalLobby m_LocalLobby;
         private LobbyServiceFacade m_LobbyServiceFacade;
 
+        public OnlineMode OnlineMode { get; set; }
```
SamuelBellomo, line 91: Same here, we should only have one source of truth

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ClientGameNetPortal.cs
LPLafontaineB: It's used when catching an exception while starting with relay mode, but this is not needed for this PR, I'll remove it to make things simpler.

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/GameNetPortal.cs
LPLafontaineB: I removed this field entirely, I don't really think it is required here, the GameNetPortal can simply return auth's playerId or the guid based on if we are connected or not.

SamuelBellomo (dismissed):

Assets/BossRoom/Scripts/Shared/ProfileManager.cs
```diff
@@ -0,0 +1,44 @@
+using UnityEngine;
+
+#if UNITY_EDITOR
+using ParrelSync;
+#endif
+
+public static class ProfileManager
```
fernando-cortez, line 7: Namespace 👀

LPLafontaineB: Oops, I really should develop a reflex for those

fernando-cortez (approved):

SamuelBellomo (approved):

PR #490, opened 2022-02-25
fix: same player number for multiple players in second playthrough

LPLafontaineB:
### Description (*)

This PR adds verification that player number is available before re-assigning it to a reconnecting player, preventing the occasional scenario where we would have two players with the same player number.
### Issue Number(s) (*)

Fixes issue(s): [MTT-2645](https://jira.unity3d.com/browse/MTT-2645)
### Manual testing scenarios

Here are some repro steps for the bug this PR is fixing:
1. Start hosting a game and join with a client using a parrelsync clone
2. After going through character selection and moving to the BossRoom scene, join with another parrelsync clone
3. Use the GoToPostGame cheat
4. Pause the first clone to connect, then, with the host, hit Play Again
5. After the second clone loads the Character Selection scene, unpause the first clone
6. Now both clones are P2 (with this fix, now we have a P2 and a P3 instead)

### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Scripts/Server/Game/State/ServerCharSelectState.cs
+++ b/Assets/BossRoom/Scripts/Server/Game/State/ServerCharSelectState.cs
@@ -189,16 +189,7 @@ namespace Unity.Multiplayer.Samples.BossRoom.Server
         {
             for (int possiblePlayerNum = 0; possiblePlayerNum < CharSelectData.k_MaxLobbyPlayers; ++possiblePlayerNum)
             {
-                bool found = false;
-                foreach (CharSelectData.LobbyPlayerState playerState in CharSelectData.LobbyPlayers)
-                {
-                    if (playerState.PlayerNum == possiblePlayerNum)
-                    {
-                        found = true;
-                        break;
-                    }
-                }
-                if (!found)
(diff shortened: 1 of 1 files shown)


SamuelBellomo (dismissed):

Assets/BossRoom/Scripts/Server/Game/State/ServerCharSelectState.cs
```diff
@@ -207,15 +198,30 @@ private int GetAvailablePlayerNum()
             return -1;
         }
 
+        bool IsPlayerNumAvailable(int playerNum)
```
fernando-cortez, line 201: Just a nitpick on Num -> Number.

fernando-cortez (commented):
Abbreviation nitpick aside, this looks good to me.

Assets/BossRoom/Scripts/Server/Game/State/ServerCharSelectState.cs
fernando-cortez: Didn't anticipate that many external modifications!

fernando-cortez (approved):

SamuelBellomo (approved):

PR #494, opened 2022-03-01
feat: popup panel for connection status messages

LPLafontaineB:
### Description (*)

This PR adds a simple reusable popup panel that lives in the SettingsPanelCanvas inside the DontDestroyOnLoadScene and uses it to display messages to players related to connection status. Connection status messages are now published through a new message channel instead of being handled by LobbyUIMediator. Finally, a manager is added, which subscribes to this message channel and calls the PopupPanel when needed.
### Issue Number(s) (*)

Fixes issue(s): [MTT-2677](https://jira.unity3d.com/browse/MTT-2677)
### Manual testing scenarios

1. Start a game without linking the project id to see a message about anonymous sign-in failing
2. Start a game with one or more clients and disconnect or pause the host to disconnect clients or make them timeout and see the corresponding messages
3. Try to join a full lobby to see another message
### Contribution checklist


Assets/BossRoom/Scripts/Client/UI/PopupPanel.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        /// <summary>
+        /// Sets the panel to match the given specifications to notify the player.  If display image is set to true, it will display
+        /// </summary>
+        /// <param name="titleText">The title text at the top of the panel</param>
+        /// <param name="mainText"> The text just under the title- the main body of text</param>
+        /// <param name="confirmFunction"> The function to call when the confirm button is pressed.</param>
+        public static void ShowPopupPanel(string titleText, string mainText, Action confirmFunction = null)
+        {
+            if (s_Instance != null)
```
SamuelBellomo, line 57: One of the classic issues with popup panel is popup stacking, where you have multiple events triggering popups one after another. We might not need this right now, but I'd still have some assert that checks if the popup was already shown, to have some alarm in case this ever happens.

Assets/BossRoom/Scripts/Client/UI/PopupPanel.cs
```diff
@@ -0,0 +1,77 @@
...
+        [SerializeField]
+        TextMeshProUGUI m_TitleText;
+        [SerializeField]
+        TextMeshProUGUI m_MainText;
+
+        static PopupPanel s_Instance;
+
+        void Awake()
+        {
+            s_Instance = this;
```
SamuelBellomo, line 25: if you want a singleton, you might want to check if s_Instance is null here.

SamuelBellomo (dismissed):
some improvements for popup, but looks good to me

Assets/BossRoom/Scripts/Client/Game/State/ClientMainMenuState.cs
```diff
@@ -100,7 +100,7 @@ void OnAuthSignIn()
 
             void OnSignInFailed()
             {
-                Debug.LogError("For some reason we can't authenticate the user anonymously - that typically means that project is not properly set up with Unity services.");
+                PopupPanel.ShowPopupPanel("Authentication Error", "For some reason we can't authenticate the user anonymously - that typically means that project is not properly set up with Unity services.");
```
fernando-cortez, line 103: I'm thinking this, and `LobbyUIMediator.OnRelayJoinFailed`'s invocation of PopupPanel could serve instead as places where a "Relay" message is published, instead of Popup panel being invoked in these two places. Granted, there's only 2 instances of Relay-based connection messages, but it would be a neat place to wrap up these two message strings that shouldn't live inside `ClientMainMenuState` or `LobbyUIMediator`.

LPLafontaineB: There currently exists a UnityServiceErrorMessage that is used to publish messages linked with Lobby. I could use this same type to publish messages for Authentication and Relay as well, and add a subscriber to this channel to call the PopupPanel, which would also add popups for those other Lobby messages. However, some of the Lobby messages happen at the same time as connection status messages (i.e.: receive a GenericDisconnect from the host shutting down, then a "Host has left the lobby" message), so to avoid overwriting the original popup we would need to add popup stacking. Unless we think it should be in a later PR, I'll refactor this to allow stacking

Assets/BossRoom/Scripts/Client/UI/PopupPanel.cs
LPLafontaineB: Good point. I think it would still be useful now, it would allow us to use the same popup for the UnityServiceErrorMessage that are being published right now but would overwrite some ConnectionStatusMessages if they were also sent to the PopupPanel

SamuelBellomo: dunno if we want to complicate this for GDC though. New features == new bugs. If a popup is already open, we could just log the next popup for example. And fix this on develop

LPLafontaineB: Makes sense, I'll create a new ticket to add stacking and include UnityServiceErrorMessages in it

SamuelBellomo (approved):

Assets/BossRoom/Scripts/Client/UI/PopupPanel.cs
pdeschain: Yes, but why use singleton when we have DI?

SamuelBellomo: Is DI the best use for this here? Every single class that'll want to show a popup will need to have an "Inject" method. while here, those classes just call popup.show. What would be the pros of using DI?
cc @LPLafontaineB

SamuelBellomo: Meh, I see you already did the change, why not. I would just be careful about blindly using DI everywhere. Let's go with what makes sense.

SamuelBellomo (dismissed):

Assets/BossRoom/Scripts/Client/UI/PopupPanel.cs
LPLafontaineB: Yes I just saw your comment, I can still revert it if we find using a singleton is better. Thinking back on it, since this might be used in a lot of different places, we might prefer to use a singleton instead, to avoid having to inject it in a lot of places when it will be used for a single method call.

SamuelBellomo: DI could be better than singletons in cases we want to unit test something. But here, this isn't really a true singleton, you could make sure you can destroy and recreate that monobehaviour and it'd work fine.
I could see a use for DI if we want to have different popups between main menu et boss room scene? Where we'd have different instances of popup, but only one at a time? But that seems like pre-optimization. If we reach that point, there's nothing preventing us from mixing singleton and DI too, where singleton would just return the injected instance.

LPLafontaineB: Wouldn't we want on the contrary to have popups persist from one scene to the other?

SamuelBellomo: We could have a different instance, with different visuals? That's a lot of "could" and "maybe", current situation is we have one single popup across multiple scenes.

LPLafontaineB: in that case, should this be reverted back? Since both work similarly, which version would be easier and cleaner for us to use, and for users?

SamuelBellomo (dismissed):

Assets/BossRoom/Scripts/Client/Game/State/ClientMainMenuState.cs
fernando-cortez: I see. So if stacking not going in for this PR, this is okay. Is there a ticket already for stacking popup messages?

LPLafontaineB: yes, and another one to redirect those messages to the popup panel

Assets/BossRoom/Scripts/Client/UI/ConnectionStatusMessageUIManager.cs
```diff
@@ -0,0 +1,53 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using Unity.Multiplayer.Samples.BossRoom;
+using Unity.Multiplayer.Samples.BossRoom.Shared.Infrastructure;
+using Unity.Multiplayer.Samples.BossRoom.Visual;
+using UnityEngine;
+
+public class ConnectionStatusMessageUIManager : MonoBehaviour
```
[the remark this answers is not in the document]
fernando-cortez: A small summary here and it's gtg.

Assets/BossRoom/Scripts/Client/Game/State/ClientMainMenuState.cs
fernando-cortez: Beautiful 👍

Assets/BossRoom/Scripts/Client/UI/ConnectionStatusMessageUIManager.cs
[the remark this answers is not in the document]
LPLafontaineB: my bad, I forgot the summary. I updated the one for the popup panel as well to better match what it is now

SamuelBellomo (approved):

fernando-cortez (approved):

PR #497, opened 2022-03-02
fix: imps spawning issues when late joining

LPLafontaineB:
### Description (*)

This PR disables the CharacterVisualization script and the renderer components in its children before spawning. This is a temporary fix for GDC. Even though the visuals are not activated, the GameObjects still exist, and the clients receive warnings that they haven't been spawned.
### Issue Number(s) (*)

Fixes issue(s): temporary fix for [MTT-2433](https://jira.unity3d.com/browse/MTT-2433) & [MTT-1558](https://jira.unity3d.com/browse/MTT-1558)
### Manual testing scenarios

1. Start a game as a host
2. Kill a few imps and progress to the room with the door to load the second additive scene
3. Late join as a client
4. See the dead imps are no longer visible and the other imps that are still alive in the first additive scene are, as well as those in the second additive scene.
### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Scripts/Client/Game/Character/ClientCharacterVisualization.cs
+++ b/Assets/BossRoom/Scripts/Client/Game/Character/ClientCharacterVisualization.cs
@@ -64,13 +64,29 @@ namespace Unity.Multiplayer.Samples.BossRoom.Visual
 
         Quaternion m_LerpedRotation;
 
+        void Awake()
+        {
+            foreach (var rendererComponent in GetComponentsInChildren<Renderer>())
+            {
+                rendererComponent.enabled = false;
+            }
+
+            enabled = false;
+        }
+
         public override void OnNetworkSpawn()
         {
             if (!IsClient || transform.parent == null)
             {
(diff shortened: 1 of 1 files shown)


Assets/BossRoom/Scripts/Client/Game/Character/ClientCharacterVisualization.cs
```diff
@@ -64,14 +64,30 @@ public class ClientCharacterVisualization : NetworkBehaviour
 
         Quaternion m_LerpedRotation;
 
+        void Awake()
+        {
+            foreach (var rendererComponent in GetComponentsInChildren<Renderer>())
```
SamuelBellomo, line 69: Why not SetActive(false) on the whole GameObject? Why disable visual components only?

LPLafontaineB: Because if the GameObject is not active, it will never be spawned

fernando-cortez (approved):

PR #505, opened 2022-03-04
lobby fix: Removing quit when lobby detects host left.

SamuelBellomo:
### Description (*)

Removing quit when lobby detects host left.
This should already be handled by NGO side callback (on disconnect)
Adding in-editor quit, to have the same behaviour as builds

### Issue Number(s) (*)

Fixes issue(s):
[MTT-2744](https://jira.unity3d.com/browse/MTT-2744)
### Manual testing scenarios

1. Tested in char select scene, host quits.
2. Tested in game scene, host quits.

### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Scripts/Shared/ApplicationController.cs
+++ b/Assets/BossRoom/Scripts/Shared/ApplicationController.cs
@@ -109,7 +109,11 @@ namespace Unity.Multiplayer.Samples.BossRoom.Shared
             }
             else
             {
+#if UNITY_EDITOR
+                UnityEditor.EditorApplication.isPlaying = false;
+#else
                 Application.Quit();
+#endif
             }
         }
     }
--- a/Assets/BossRoom/Scripts/Shared/ApplicationController.cs
+++ b/Assets/BossRoom/Scripts/Shared/ApplicationController.cs
@@ -93,28 +93,26 @@ namespace Unity.Multiplayer.Samples.BossRoom.Shared
             return canQuit;
         }
 
+        public void LeaveSession()
+        {
+            m_LobbyServiceFacade.ForceLeaveLobbyAttempt();
+
+            // first disconnect then return to menu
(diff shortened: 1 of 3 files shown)


Assets/BossRoom/Scripts/Shared/ApplicationController.cs
```diff
@@ -109,7 +109,11 @@ public void QuitGame()
             }
             else
```
LPLafontaineB, line 110: Not exactly related to this specific issue, but it might help prevent some other bugs if we replaced this if-else with two methods instead, since the only place where this is now called (UIQuitPanel) knows about that condition and knows what outcome it wants based on it.

SamuelBellomo: good catch, fixed

LPLafontaineB (approved):

Assets/BossRoom/Scripts/Shared/Net/UnityServices/Lobbies/LobbyServiceFacade.cs
```diff
@@ -126,7 +127,7 @@ void OnSuccess(Lobby lobby)
                     }
                     m_UnityServiceErrorMessagePub.Publish(new UnityServiceErrorMessage("Host left the lobby","Disconnecting."));
                     ForceLeaveLobbyAttempt();
-                    m_ApplicationController.QuitGame();
+                    // no need to disconnect NGO, it should already be handled by NGO callback to disconnect
```
fernando-cortez, line 130: Just nitpick of NGO->Netcode or Netcode for GameObjects. Good to go otherwise!

SamuelBellomo: done

fernando-cortez (approved):

LPLafontaineB (approved):

PR #507, opened 2022-03-04
Simplified Floor Tiles

jilfranco-unity:
### Description (*)
This PR is part of the art improvements for GDC epic. The ~750 tiny floor tiles have been replaced with a mix of ~150 larger and small floor tiles. Replacing the hundreds of small tiles with larger tiles reduces scene complexity (making it easier to manage and modify the level), reduces vert/tri count (aiding performance), and makes lighting bakes cleaner/faster (by reducing the amount of objects that need to bake, and reducing the amount of places there can be seams).

Prefabs for the new pieces have been created

Prefabs for Entrance, Transition Area, and Boss Room have been changed to use these new tiles

Lighting has been rebaked

### Issue Number(s) (*)
[Jira ticket here](https://jira.unity3d.com/browse/MTT-2758)


SamuelBellomo:
@jilfranco-unity 
Is this just for GDC or both GDC and develop?
Also watch out, this points to develop at the moment.

jilfranco-unity:
> 

This would be for both! I think it's good to have this change overall. At which branch should it point if I want the change for both?

SamuelBellomo:
> > 
> 
> This would be for both! I think it's good to have this change overall. At which branch should it point if I want the change for both?

@jilfranco-unity this is fine, but we'll need to cherrypick your squashed commit once its merged to develop. Hopefully that's something you can do (in case there are conflicts). We can do it together if you need help.

Assets/BossRoom/Prefabs/Dungeon/Dungeon Pieces/env_floor3.prefab
```diff
@@ -92,7 +92,7 @@ BoxCollider:
   m_GameObject: {fileID: 4371491365999748502}
   m_Material: {fileID: 0}
   m_IsTrigger: 0
-  m_Enabled: 1
```
fernando-cortez, line 95: So from what I remember, the BoxCollider component on these floor tile prefabs should be enabled (and in-scene as well) to bake the NavMesh. A script on awake turns these components off, and the big level plane is used instead for the raycasts from click/touch.

Assets/BossRoom/Prefabs/Dungeon/Dungeon Pieces/env_floor7.prefab
```diff
@@ -0,0 +1,112 @@
+%YAML 1.1
```
fernando-cortez, line 1: So I'm noticing a considerable amount of smaller floor pieces on top of the newer big floor pieces.

These are just accent pieces, right? I'm curious how it'd look like without them and just rely on the bigger pieces, to double down on the optimization we're gaining.

jilfranco-unity: They are just accent pieces, yeah! I think just having the same texture repeated everywhere won't look very good, but let me try modifying the UVs of the big pieces so they have more texture variation!

jilfranco-unity:
Removed accent floor 2 & floor 3 pieces to further simplify the floor down to ~60 pieces. Edited the UVs of the bigger floor tiles to break up repeating texture with the texture that was on floor 2

SamuelBellomo (dismissed):

fernando-cortez (commented):
Can you push the changes to the floor Colliders to the prefabs themselves as well?

SamuelBellomo (dismissed):

SamuelBellomo (approved):

fernando-cortez (approved):

PR #510, opened 2022-03-04
feat: popup panel stacking

LPLafontaineB:
### Description (*)

This PR simplifies PopupPanel so its only responsibility is to display a single message. It moves the responsibility of creating those messages to the PopupManager. The PopupManager will instantiate PopupPanel gameObjects and display them when it is called. It will reuse those instantiated objects when possible (which is most of the time since we likely won't display more than one or two at once in a regular case). 
### Issue Number(s) (*)

Fixes issue(s): [MTT2733](https://jira.unity3d.com/browse/MTT-2733)
### Manual testing scenarios

The simplest way to see this PR's effects is to duplicate calls to ShowPopupPanel() to have multiple popups at once. Or add a coroutine that perdically calls this, so that you can experiment with closing some of them before new ones are created.
### Contribution checklist


LPLafontaineB:
Put on hold as it is not needed for GDC

LPLafontaineB:
tested on my phone with a 20:9 ratio and 720x1600 resolution (both in portrait and landscape orientations) with enough popups stacked to reach the maximum popup offset and they all were still visible in the screen.

Assets/BossRoom/Scripts/Client/UI/PopupPanel.cs
```diff
@@ -13,72 +13,38 @@ public class PopupPanel : MonoBehaviour
         TextMeshProUGUI m_TitleText;
         [SerializeField]
         TextMeshProUGUI m_MainText;
+        [SerializeField]
+        CanvasGroup m_CanvasGroup;
 
-        bool m_IsPopupShown;
-
-        static PopupPanel s_Instance;
+        public bool IsDisplaying => m_CanvasGroup.alpha > 0;
```
SamuelBellomo, line 19: not a fan of implicit meaning like this. (canvasGroup alpha == "IsDisplaying --> what happens when we change our structure and canvasGroup isn't representative of the whole popup anymore?
Having an explicit IsDisplaying = false in Hide() would make this clearer

SamuelBellomo (approved):

Assets/BossRoom/Scripts/Client/UI/PopupManager.cs
```diff
@@ -0,0 +1,90 @@
...
+    /// messages in succession.
+    /// </summary>
+    public class PopupManager : MonoBehaviour
+    {
+        [SerializeField]
+        GameObject m_PopupPanelPrefab;
+
+        List<PopupPanel> m_PopupPanels = new List<PopupPanel>();
+
+        static PopupManager s_Instance;
```
pdeschain, line 19: This is a good candidate for being turned into singleton with our DI.

pdeschain:
Apart from the DI comment - this PR is 👍

Assets/BossRoom/Scripts/Client/UI/PopupManager.cs
LPLafontaineB: Since this is defined in the Client assembly, I cannot bind it inside ApplicationController, however, so it should either be moved to the Shared assembly or it should be bound in all the ClientGameStates. If the latter makes more sense, it would need to wait until PR #579 is merged to avoid conflicts

fernando-cortez (commented):
Will approve once the conversation regarding DI is resolved.

SamuelBellomo (approved):

SamuelBellomo (approved):

pdeschain (approved):

PR #513, opened 2022-03-07
feat: auto reconnect

LPLafontaineB:
### Description (*)

This PR adds automatic reconnect attempts when a players loses connection. It will try to reconnect three times before giving up and sending the player back to the main menu (unless the player uses the quit button earlier).  As a workaround until a fix for [MTT-2684](https://jira.unity3d.com/browse/MTT-2684) is available in NGO pre.7, before trying to reconnect we load an empty scene. We will wait until that fix is available before cherrypicking this to develop.
### Issue Number(s) (*)

Fixes issue(s): [MTT-2617](https://jira.unity3d.com/browse/MTT-2617)
### Manual testing scenarios

1. Start a game with at least one client
2. On the host, pause the game until the client disconnects
3. See the client moved to the new scene with a popup message
4. Before the end of the third attempt, unpause the host
5. See the player correctly rejoins the host
### Questions or comments

There currently is an issue when testing with debug builds or in the editor. Since the host is paused, it doesn't have the time to time out the clients on its side, so when it is unpaused and the clients reconnect to it, it considers them as duplicate connections. However, in debug builds or in the editor, duplicate connections are allowed and it will be treated as a new player instead. This issue will be fixed when #488 is merged.
### Contribution checklist


SamuelBellomo:
I see two labels, cherrypick and workaround, which is it? They are mutually exclusive. Knowing this relies on a workaround and that there's an MTT bug we're waiting to fix, this should be just workaround and not cherrypicked to develop correct?

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ClientGameNetPortal.cs
```diff
@@ -127,26 +138,61 @@ private void OnDisconnectOrTimeout(ulong clientID)
...
         }
 
+        private IEnumerator TryToReconnect()
+        {
+            Debug.Log("Lost connection to host, trying to reconnect...");
+            int nbTries = 0;
+            while (nbTries < 3)
+            {
+                NetworkManager.Singleton.Shutdown();
+                yield return new WaitForSeconds(1); // wait a short while to leave time for NetworkManager to complete shutting down
```
SamuelBellomo, line 174: What happens if the StopCoroutine happens during that yield? I'd do a IsConnectedClient check before the shutdown.

Assets/BossRoom/Scripts/Client/UI/ConnectionStatusMessageUIManager.cs
```diff
@@ -38,13 +38,17 @@ void OnConnectStatus(ConnectStatus status)
                     PopupPanel.ShowPopupPanel("Connection Failed", "The Host is full and cannot accept any additional connections.");
                     break;
                 case ConnectStatus.Success:
+                    PopupPanel.ShowPopupPanel("Success!", "Joining Now");
```
SamuelBellomo, line 41: don't all our popup panels have an "ok" button to click? We shouldn't have to click "ok" on a connection success.

Assets/BossRoom/Scripts/Client/UI/ConnectionStatusMessageUIManager.cs
```diff
@@ -38,13 +38,17 @@ void OnConnectStatus(ConnectStatus status)
...
                     break;
                 case ConnectStatus.LoggedInAgain:
                     PopupPanel.ShowPopupPanel("Connection Failed", "You have logged in elsewhere using the same account.");
                     break;
                 case ConnectStatus.GenericDisconnect:
                     PopupPanel.ShowPopupPanel("Disconnected From Host", "The connection to the host was lost");
                     break;
+                case ConnectStatus.Reconnecting:
+                    PopupPanel.ShowPopupPanel("Attempting reconnection", "Lost connection to the Host, attempting to reconnect...");
```
SamuelBellomo, line 50: same here, this should be a temporary popup, not something that requires an interraction.

LPLafontaineB:
> I see two labels, cherrypick and workaround, which is it? They are mutually exclusive. Knowing this relies on a workaround and that there's an MTT bug we're waiting to fix, this should be just workaround and not cherrypicked to develop correct?

Yes my bad, it should only have had the workaround label

Assets/BossRoom/Scripts/Client/UI/ConnectionStatusMessageUIManager.cs
LPLafontaineB: good point, I'll add something that auto-closes after x seconds, or maybe wait until a scene load. I'll look into which one feels better for this

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ClientGameNetPortal.cs
LPLafontaineB: Normally it shouldn't, since while connecting, we wait until k_TimeoutDuration until failing connection, so the StopCoroutine should only happen during the yield on line 177. There is maybe a small probability of the timing being too tight though, so maybe we should increase the duration of that yield to 1.1*k_timeoutDuration or something similar?

Assets/BossRoom/Scripts/Client/UI/ConnectionStatusMessageUIManager.cs
LPLafontaineB: Went with a manual close instead, I think it makes more sense. For this one, I call it when a new scene is loaded, and for the reconnect popup, I call it when a new connection satus message is received

Assets/BossRoom/Scripts/Client/UI/ConnectionStatusMessageUIManager.cs
```diff
@@ -38,17 +42,28 @@ void OnConnectStatus(ConnectStatus status)
                     PopupPanel.ShowPopupPanel("Connection Failed", "The Host is full and cannot accept any additional connections.");
                     break;
                 case ConnectStatus.Success:
+                    m_PopupIdToClose = PopupPanel.ShowPopupPanel("Success!", "Joining Now", false);
```
SamuelBellomo, line 45:
```suggestion
                    m_PopupIdToClose = PopupPanel.ShowPopupPanel("Success!", "Joining Now", isCloseableByUser: false);
```
Just to make this clearer and not have a single bool hanging around

Assets/BossRoom/Scripts/Client/UI/ConnectionStatusMessageUIManager.cs
```diff
@@ -38,17 +42,28 @@ void OnConnectStatus(ConnectStatus status)
...
+                    SceneManager.sceneLoaded += ClosePopupOnsceneLoaded;
                     break;
                 case ConnectStatus.LoggedInAgain:
                     PopupPanel.ShowPopupPanel("Connection Failed", "You have logged in elsewhere using the same account.");
                     break;
                 case ConnectStatus.GenericDisconnect:
                     PopupPanel.ShowPopupPanel("Disconnected From Host", "The connection to the host was lost");
                     break;
+                case ConnectStatus.Reconnecting:
+                    m_PopupIdToClose = PopupPanel.ShowPopupPanel("Attempting reconnection", "Lost connection to the Host, attempting to reconnect...", false);
```
SamuelBellomo, line 55:
```suggestion
                    m_PopupIdToClose = PopupPanel.ShowPopupPanel("Attempting reconnection", "Lost connection to the Host, attempting to reconnect...", isCloseableByUser: false);
```

Assets/BossRoom/Scripts/Client/UI/PopupPanel.cs
```diff
@@ -13,9 +13,16 @@ public class PopupPanel : MonoBehaviour
...
         [SerializeField]
         TextMeshProUGUI m_MainText;
+        [SerializeField]
+        GameObject m_ConfirmButton;
+        [SerializeField]
+        GameObject m_LoadingImage;
 
         bool m_IsPopupShown;
 
+        long m_DisplayedPopupId = -1;
```
SamuelBellomo, line 23: shouldn't resetState reset this as well?
(that's why I hate "reset" functions, that's really against OOP. You usually just want to destroy your objects and create new ones)

Assets/BossRoom/Scripts/Client/UI/PopupPanel.cs
```diff
@@ -51,34 +60,61 @@ void ResetState()
...
-                m_IsPopupShown = true;
+            gameObject.SetActive(true);
+            m_IsPopupShown = true;
+            m_DisplayedPopupId = s_NextPopupId++;
+            return m_DisplayedPopupId;
+        }
+
+        void ClosePopupPanel(long popupId)
+        {
+            if (m_DisplayedPopupId == popupId)
```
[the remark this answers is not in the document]
LPLafontaineB: This is not only in case of stacking, this is simply to make sure that we are closing the right popup. There are other cases than popup stacking where the current popup might be different than the one we want to close. I'll add the summary of the public static method to this one too to clarify

Assets/BossRoom/Scripts/Client/UI/PopupPanel.cs
LPLafontaineB: I agree, I'm not a fan of those as well. For this PR I'll simply add that to ResetState, but I'll definitely rework this in my PR implementing stacking

LPLafontaineB: In its current state, it already solves some of the issues with this, but it can be better I think

SamuelBellomo (dismissed):

LukeStampfli (approved):

SamuelBellomo (dismissed):

LukeStampfli (approved):

SamuelBellomo (approved):

fernando-cortez (approved):

PR #515, opened 2022-03-07
fix: leaving lobby when disconnecting

LPLafontaineB:
### Description (*)

With this PR, the host now removes disconnecting clients from their lobby and deletes it when they themselves leave. Meanwhile, the clients also leave the lobby when disconnecting, and after each failed reconnection attempt. This PR also fixes an infinite loop issue that appeared in certain circumstances with RateLimitCooldown by using two operation queues. This fix allow us to put the number of reconnection attempts back at 2.
### Issue Number(s) (*)

Fixes issue(s): [MTT-2701](https://jira.unity3d.com/browse/MTT-2701) and works around the issues caused by [MTT-2667](https://jira.unity3d.com/browse/MTT-2667). Also fixes [MTT-2794](https://jira.unity3d.com/browse/MTT-2794)
### Manual testing scenarios

1. Start a game as a host
2. Join with a client
3. Pause the client until the host times it out
4. Unpause the client and see that it is properly reconnected

This also works when pausing the host to cause a timeout, or when quitting the game as the client or host, or disconnecting via the in-game quit button.

Also works when connection gets denied during approval check (i.e. full game or duplicate connection).
### Contribution checklist


SamuelBellomo:
same here. cherrypick vs workaround labels

fernando-cortez (approved):

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ClientGameNetPortal.cs
```diff
@@ -123,6 +123,9 @@ private void OnDisconnectOrTimeout(ulong clientID)
             // Only handle client disconnect
             if (!NetworkManager.Singleton.IsHost)
             {
+                // Leave the current lobby.
+                //TODO: When the SessionManager will use the PlayerId from AuthenticationService instead of our current custom GUID, this could be replaced by the host kicking the disconnecting client from the lobby
+                m_LobbyServiceFacade.ForceLeaveLobbyAttempt();
```
SamuelBellomo, line 128: check in lobby data if we have player IDs so the host can do it

LukeStampfli (approved):

SamuelBellomo (approved):

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ClientGameNetPortal.cs
```diff
@@ -31,7 +31,7 @@ public class ClientGameNetPortal : MonoBehaviour
         /// Time in seconds before the client considers a lack of server response a timeout
         /// </summary>
         private const int k_TimeoutDuration = 10;
-        const int k_NbReconnectAttempts = 1;
+        const int k_NbReconnectAttempts = 2;
```
SamuelBellomo, line 34: From my current experiments, one reconnect attempt seems enough. Each will have a 10 sec timeout. Waiting 20 seconds to then be able to join again seems excessive.

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ClientGameNetPortal.cs
```diff
@@ -192,13 +199,33 @@ private IEnumerator TryToReconnect()
...
+                    // todo should use yield return new WaitUntil(() => task.IsCompleted);
+                    var joiningLobby = true;
+                    m_LobbyServiceFacade.JoinLobbyAsync("", lobbyCode, onSuccess: lobby =>
+                        {
+                            m_LobbyServiceFacade.BeginTracking(lobby);
+                            ConnectClient(null);
+                            joiningLobby = false;
+                        }
+                        , onFailure: () => joiningLobby = false);
+                    yield return new WaitWhile(() => joiningLobby); // This is not a clean way of doing this, ideally we would want to replace that logic with a proper await.
```
pdeschain, line 216: If you need the task object - then I suggest modifying the `LobbyAPIInterface` in such a way so that instead of returning `void` all over the place it would return the Task object that it creates. This way you'd be able to avoid this whole pattern with `joiningLobby = true` and the two functions that switch it to false.

LPLafontaineB: I was thinking of leaving it this way for the moment since #501 already aims to do that change and I didn't want to step on any toes, but anyways this specific part won't be going to develop immediately since auto-reconnect won't be there until pre.7 at least. So in that case it shouldn't create conflicts

pdeschain (approved):

PR #523, opened 2022-03-09
fix: postgame menu button

LukeStampfli:
### Description (*)

Fixes a bug where pressing the "return to menu" button in the post game screen did not do anything when pressing the button. A repeated (2nd) press of the button would correctly load the menu scene.

When returning to the menu from the post game scene the active lobby was not quit so the lobby list still showed 2/6 players and did not allow the client to join the lobby again.

### Issue Number(s) (*)

Fixes issue(s): [MTT-2791](https://jira.unity3d.com/browse/MTT-2791)
### Manual testing scenarios
- Build and run a host (create a lobby session)
- Run a client in editor and join the host.
- Select characters and press ready
- Use the cheat window to go to the post game lobby
- Press the "return to menu" button on the client

### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Scripts/Client/UI/PostGameUI.cs
+++ b/Assets/BossRoom/Scripts/Client/UI/PostGameUI.cs
@@ -15,7 +15,7 @@ namespace Unity.Multiplayer.Samples.BossRoom.Visual
     /// </summary>
     public class PostGameUI : MonoBehaviour
     {
-        private ApplicationController m_ApplicationController;
+        ApplicationController m_ApplicationController;
 
         [SerializeField]
         private Light m_SceneLight;
@@ -42,7 +42,7 @@ namespace Unity.Multiplayer.Samples.BossRoom.Visual
         private Color m_LoseLightColor;
 
         [Inject]
-        private void InjectDependencies(ApplicationController applicationController)
+        void InjectDependencies(ApplicationController applicationController)
         {
             m_ApplicationController = applicationController;
         }
--- a/Assets/BossRoom/Scripts/Client/UI/PostGameUI.cs
+++ b/Assets/BossRoom/Scripts/Client/UI/PostGameUI.cs
@@ -2,8 +2,11 @@ using UnityEngine;
 using UnityEngine.SceneManagement;
 using UnityEngine.UI;
 using TMPro;
+using Unity.Multiplayer.Samples.BossRoom.Shared;
+using Unity.Multiplayer.Samples.BossRoom.Shared.Infrastructure;
 using Unity.Multiplayer.Samples.Utilities;
 using Unity.Netcode;
+using UnityEngine.Assertions;
 
 namespace Unity.Multiplayer.Samples.BossRoom.Visual
 {
@@ -12,6 +15,8 @@ namespace Unity.Multiplayer.Samples.BossRoom.Visual
     /// </summary>
     public class PostGameUI : MonoBehaviour
     {
+        private ApplicationController m_ApplicationController;
+
         [SerializeField]
         private Light m_SceneLight;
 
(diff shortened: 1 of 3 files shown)


Assets/BossRoom/Scripts/Client/UI/PostGameUI.cs
```diff
@@ -77,18 +94,15 @@ void SetPostGameUI(WinState winState)
...
         }
 
         public void OnMainMenuClicked()
         {
-            // Player is leaving this group - leave current network connection first
-            var gameNetPortal = GameObject.FindGameObjectWithTag("GameNetPortal").GetComponent<GameNetPortal>();
-            gameNetPortal.RequestDisconnect();
-
-            SceneLoaderWrapper.Instance.LoadScene("MainMenu");
+            m_ApplicationController.LeaveSession();
```
LukeStampfli, line 105: This function was already doing the same as we did in the PostGameUI function but it also properly leaves the lobby.

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoaderWrapper.cs
```diff
@@ -69,7 +69,7 @@ public void AddOnSceneEventCallback()
         /// <param name="loadSceneMode">If LoadSceneMode.Single then all current Scenes will be unloaded before loading.</param>
         public void LoadScene(string sceneName, LoadSceneMode loadSceneMode = LoadSceneMode.Single)
         {
-            if (m_NetworkManager != null && m_NetworkManager.IsListening && m_NetworkManager.NetworkConfig.EnableSceneManagement)
+            if (m_NetworkManager != null && m_NetworkManager.IsListening && !m_NetworkManager.ShutdownInProgress && m_NetworkManager.NetworkConfig.EnableSceneManagement)
```
LukeStampfli, line 72: This was needed because else when calling Shutdown on NetworkManager and after call LoadScene nothing would happen. Because Shutdown is asynchronous. Changing this did not seem to break anything but it feels a bit sketchy

Assets/BossRoom/Scripts/Client/UI/PostGameUI.cs
```diff
@@ -77,18 +94,15 @@ void SetPostGameUI(WinState winState)
         public void OnPlayAgainClicked()
         {
             // this should only ever be called by the Host - so just go ahead and switch scenes
+            Assert.IsTrue(NetworkManager.Singleton.IsServer);
```
LukeStampfli, line 97: Not related to the PR but I saw a "should" so we have an assert now.

LukeStampfli:
#515 Seems to address similar issues in regards to quitting lobby @LPLafontaineB 👀

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoaderWrapper.cs
LPLafontaineB: Should IsListening return false when ShutdownInProgress is true?

LPLafontaineB (approved):
I don't think this needs to be cherrypicked to develop though, since the bug is introduced by the loading screen which won't be merged to develop until the next NGO release

fernando-cortez (commented):
Nitpick aside, lgtm.

LukeStampfli:
Switched to workaround instead of cherry pick since this seems to be introduced by our latest loading screen work as mentioned above by @LPLafontaineB

Assets/BossRoom/Scripts/Client/UI/PostGameUI.cs
```diff
@@ -12,6 +15,8 @@ namespace Unity.Multiplayer.Samples.BossRoom.Visual
     /// </summary>
     public class PostGameUI : MonoBehaviour
     {
+        private ApplicationController m_ApplicationController;
```
LukeStampfli, line 18:
```suggestion
        ApplicationController m_ApplicationController;
```

Assets/BossRoom/Scripts/Client/UI/PostGameUI.cs
```diff
@@ -36,6 +41,18 @@ public class PostGameUI : MonoBehaviour
         [SerializeField]
         private Color m_LoseLightColor;
 
+        [Inject]
+        private void InjectDependencies(ApplicationController applicationController)
```
LukeStampfli, line 45:
```suggestion
        void InjectDependencies(ApplicationController applicationController)
```

fernando-cortez (approved):

LPLafontaineB (approved):

SamuelBellomo (approved):

PR #525, opened 2022-03-09
feat: display popup when failing to connect to full lobby

LukeStampfli:
### Description (*)
Added a feature to display a popup to the user when they fail to join a lobby because it is full.
### Issue Number(s) (*)

Fixes issue(s): [MTT-2646](https://jira.unity3d.com/browse/MTT-2646)
### Manual testing scenarios
1. Modify LobbyCreationUI.cs line 36 to use a maxplayer count of 1 instead of 8.
2. Start to instances of Boss Room.
3. In one of them create a lobby.
4. In the other connect to the lobby using the connect key of the lobby.
5. Popup should appear telling you that the lobby is full.

### Contribution checklist


Assets/BossRoom/Scripts/Client/UI/UnityServicesUIHandler.cs
```diff
@@ -27,21 +27,24 @@ void ServiceErrorHandler(UnityServiceErrorMessage error)
...
             if (error.AffectedService == UnityServiceErrorMessage.Service.Lobby)
             {
-                if ((error.OriginalException as LobbyServiceException).Reason == LobbyExceptionReason.LobbyConflict)
+                switch ((error.OriginalException as LobbyServiceException).Reason)
                 {
-                    // LobbyConflict can have multiple causes. Let's add other solutions here if there's other situations that arise for this.
-                    errorMessage += "\nSee logs for possible causes and solution.";
-                    Debug.LogError($"Got service error {error.Message} with LobbyConflict. Possible conflict cause: Trying to play with two builds on the " +
-                                   $"same machine. Please use command line arg '{ClientMainMenuState.AuthProfileCommandLineArg} someName' to set a different auth profile.\n");
+                    case LobbyExceptionReason.LobbyConflict:
```
LukeStampfli, line 32: @fernando-cortez Since you just recently created this, do you think extending this here like I did is the right way to go or would you suggest we do something else?

fernando-cortez: I believe this was @SamuelBellomo's work. However, this looks good.

So if this class is meant to catch all Unity Services errors, I think we'd want to use this class as well to display Authentication errors, no?

`ClientMainMenu` creates a popup message when authentication fails. However, it seems that an authentication message is published inside of `AuthenticationServiceFacade`, but there's no listener for that to create a popup message. Could that popup message creation be moved here instead?

Assets/BossRoom/Scripts/Client/UI/UnityServicesUIHandler.cs
```diff
@@ -27,21 +27,24 @@ void ServiceErrorHandler(UnityServiceErrorMessage error)
             var errorMessage = error.Message;
```
fernando-cortez, line 27: This is cached, but never used below.

Assets/BossRoom/Scripts/Client/UI/UnityServicesUIHandler.cs
```diff
@@ -27,21 +27,24 @@ void ServiceErrorHandler(UnityServiceErrorMessage error)
...
+                            $"same machine. Please use command line arg '{ClientMainMenuState.AuthProfileCommandLineArg} someName' to set a different auth profile.\n");
+                        PopupPanel.ShowPopupPanel("Service error", error.Message + "\nSee logs for possible causes and solution.");
+                        break;
+                    case LobbyExceptionReason.LobbyFull:
+                        PopupPanel.ShowPopupPanel("Failed to join lobby", "Lobby is full and can't accept more players");
+                        break;
                 }
             }
-
-            PopupPanel.ShowPopupPanel("Service error", errorMessage);
```
SamuelBellomo, line 39: Wait, wouldn't this hide other error messages? This was here to at least show "something" when we had an unhandled error

Assets/BossRoom/Scripts/Client/UI/UnityServicesUIHandler.cs
SamuelBellomo: I'm thinking we should have a lobby specific message handler and then a catch all for anything unhandled.

LukeStampfli: By lobby specific message handler you'd still mean something in the same class right (just trying to make sure I'd understand this correctly).

So if we were to add a `default` case to the switch here which would show a popup like it did before I manipulated the code here we would be fine. That was an oversight from my side my goal was not to change the initial behaviour so I'll add that back. I think splitting it into a separate class would be quite messy because we'd still need to track here which messages don't need the generic popup.

SamuelBellomo: my worry is we end up with a god class here.
My thinking was we'd have one handler per service plus one "catch everything else that falls through the filter" class.
This could be preoptimization though, I'll let you decide.
And yes to the default :)

fernando-cortez: Just a heads up @LukeStampfli that this seems to end up what @pdeschain did [here](https://github.com/Unity-Technologies/com.unity.multiplayer.samples.coop/pull/547).

I think we'll just need to add the lobby full message to the lobby case, correct?

LukeStampfli: Yes I just refactored this and integrated it into @pdeschain's changes.

Assets/BossRoom/Scripts/Client/UI/UnityServicesUIHandler.cs
LukeStampfli: Fixed this, we're no longer hiding other messages.

LukeStampfli:
Merged with Phil's changes. I moved the lobby related stuff out into a separate function to keep things a bit more readable (wanted to avoid nested switches).

fernando-cortez (approved):

Assets/BossRoom/Scripts/Client/UI/UnityServicesUIHandler.cs
fernando-cortez: My concerns here were resolved. @SamuelBellomo I'll let you resolve conversation.

SamuelBellomo (approved):

PR #532, opened 2022-03-09
fix: not returning to mainmenu when reconnect fails

LPLafontaineB:
### Description (*)

This PR does three things: first it adds null checks to prevent null reference exceptions, then it adds a mechanism for the host to tell the clients when it voluntarily quits a game, so that if they quit using the in-game quit button, they will not attempt to reconnect and simply display a message instead. Finally, it reduces the number of reconnection attempts to 1 to avoid an issue that seems linked to Relay, until further investigation allows us to better understand that bug (see [MTT-2794](https://jira.unity3d.com/browse/MTT-2794) for details)
### Issue Number(s) (*)

Fixes issue(s): [MTT-2786](https://jira.unity3d.com/browse/MTT-2786) and [MTT-2795](https://jira.unity3d.com/browse/MTT-2795)
### Manual testing scenarios

1. Start a game as a host
2. Join with a client
3. With the host, leave the session via the in-game quit button
4. See the client is disconnected and returned to the main menu and doesn't try to reconnect
5. Start another game with a host and client
6. End the play mode if in editor or quit the application if using a build (not via the in-game quit button)
7. See the client attempting to reconnect, then failing and going back to the main menu
### Questions or comments

On hold because will have to wait until #523 is merged first, as this is depending on a fix introduced by this other PR.
### Contribution checklist


Assets/BossRoom/Scripts/Client/UI/UnityServicesUIHandler.cs
```diff
@@ -27,7 +27,7 @@ void ServiceErrorHandler(UnityServiceErrorMessage error)
             var errorMessage = error.Message;
             if (error.AffectedService == UnityServiceErrorMessage.Service.Lobby)
             {
-                if ((error.OriginalException as LobbyServiceException).Reason == LobbyExceptionReason.LobbyConflict)
+                if (error.OriginalException is LobbyServiceException {Reason: LobbyExceptionReason.LobbyConflict})
```
SamuelBellomo, line 30: wait, how come we can get errors coming from lobby, but with not lobby exception?

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/GameNetPortal.cs
```diff
@@ -245,8 +247,27 @@ public void RequestDisconnect()
...
+
+        void ShutDown()
+        {
             SessionManager<SessionPlayerData>.Instance.OnUserDisconnectRequest();
             NetManager.Shutdown();
         }
+
+        IEnumerator WaitToShutdown()
+        {
+            yield return new WaitForSeconds(0.5f);
```
SamuelBellomo, line 269: ouch, why?

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ServerGameNetPortal.cs
```diff
@@ -230,6 +234,16 @@ public void SendServerToClientSetDisconnectReason(ulong clientID, ConnectStatus
             writer.WriteValueSafe(status);
             NetworkManager.Singleton.CustomMessagingManager.SendNamedMessage(nameof(ClientGameNetPortal.ReceiveServerToClientSetDisconnectReason_CustomMessage), clientID, writer);
         }
+        /// <summary>
+        /// Sends a DisconnectReason to all connected clients. This should only be done on the server, prior to disconnecting the clients.
+        /// </summary>
+        /// <param name="status"> The reason for the upcoming disconnect.</param>
+        public void SendServerToAllClientsSetDisconnectReason(ConnectStatus status)
```
SamuelBellomo, line 241: I thought we already had this?

SamuelBellomo:
there's a few things that shouldn't be workarounds though, no? Only the 1 reconnect attempt limit should be a workaround?

Assets/BossRoom/Scripts/Client/UI/UnityServicesUIHandler.cs
LPLafontaineB: The default value for OriginalException is null. In LobbyServiceFacade, line 128 there is a case where a UnityServiceErrorMessage is instantiated and published without an exception.

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/GameNetPortal.cs
LPLafontaineB: Same reason as when we do it before disconnecting a client when approval check fails. If we don't, Netcode will disconnect the clients before we can send them a message telling them why.

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ServerGameNetPortal.cs
LPLafontaineB: We had this for sending it to a specific client. Adding this is simpler than looping through ConnectedClientIds to call the method we already had

LPLafontaineB:
> there's a few things that shouldn't be workarounds though, no? Only the 1 reconnect attempt limit should be a workaround?

Auto-reconnect as a whole was flagged as a workaround though and will not be in develop until pre.7 which should have the fix needed to remove the workaround auto-reconnect used

Assets/BossRoom/Scripts/Client/UI/UnityServicesUIHandler.cs
SamuelBellomo: ah crap true

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/GameNetPortal.cs
SamuelBellomo: thing is, we don't send here, we send elsewhere. The wait should be before the send of the custom message, not here then.

SamuelBellomo (dismissed):

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/GameNetPortal.cs
LPLafontaineB: I added it here since this is where the Shutdown was, but I could move it to the ClientGameNetPortal and ServerGameNetPortal if it makes it clearer

SamuelBellomo: if you plan to reuse any of this code for develop, then yeah, let's make this clearer.

pdeschain (approved):

PR #537, opened 2022-03-10
fix: simpler quitting flow

LPLafontaineB:
### Description (*)

This PR simplifies the condition in QuitPanel so that now it quits/sends to menu depending on the current active scene.
### Issue Number(s) (*)

Fixes issue(s): [MTT-2800](https://jira.unity3d.com/browse/MTT-2800)
### Manual testing scenarios

1. Try quitting from the main menu and from any other scene
2. See that in main menu the application quits while in other scenes you return to main menu
### Contribution checklist


pdeschain (dismissed):

Assets/BossRoom/Scripts/Client/UI/UIQuitPanel.cs
```diff
@@ -23,8 +23,8 @@ private void InjectDependencies(ApplicationController applicationController)
 
         void OnEnable()
         {
-            m_QuitMode = NetworkManager.Singleton == null || !NetworkManager.Singleton.IsListening;
-            m_QuitButtonText.text = m_QuitMode ? "Exit Game?" : "Leave session?";
+            m_QuitMode = SceneManager.GetActiveScene().name == "MainMenu";
```
fernando-cortez, line 26: I'd mark this as a workaround. I don't advocate for checking scene name strings anywhere in code, if we can avoid it. Perhaps this can be moved to a unique button inside MainMenu scene that just quits the game, to simplify things. That would require some UI re-work, but it's definitely cleaner/safer.

fernando-cortez (approved):

Assets/BossRoom/Scripts/Client/UI/UIQuitPanel.cs
LPLafontaineB: Good point, I'll create a cloned ticket for that

LPLafontaineB: Unless I could simply check the GameStateBehavior each time a new scene is loaded. I think it could be a cherrypickable solution

LPLafontaineB: @fernando-cortez In the end it was quite simple to implement, and I think it should be clean enough now to be cherry-picked back to develop as well (just without adding the game state for the loading scene that doesn't exist on develop)

fernando-cortez: I'm in favour of keeping the code you had originally checking for scene name. We're now adding new classes, and potentially adding test scenarios since we're adding a lot of responsibility to this quitting class. 

To get back to my original point, the goal is to have the scenes and their UI components drive the (simple) behaviour of the UI buttons. Here, I think of `GameState`s as just hardcoded scene identifiers as well, so I don't believe we're necessarily better off.

LPLafontaineB: I see. So if I understand correctly, ideally when doing a proper fix, we would like to add the quit panel to each scene instead of inside the DDOL scene, and have a serialized bool or something like it to tell it whether it should quit or return to main menu, right?

fernando-cortez: Yeah we could even separate them into two classes. Main menu should be non-networked, so we could have a simple Quit script, and a script on networked scenes that shutdown + redirect to Main Menu.

SamuelBellomo (approved):

fernando-cortez (approved):

PR #551, opened 2022-03-14
fix: leaving lobby when connection denied

LPLafontaineB:
### Description (*)

With this PR, the host removes clients from the lobby when their connection gets denied. Clients also leave on their side.
### Issue Number(s) (*)

Fixes issue(s): [MTT-2796](https://jira.unity3d.com/browse/MTT-2796)
### Manual testing scenarios

1. Start a host
2. Join with clients normally
3. Try to join with a client that gets their approval denied (currently not possible under normal circumstances, but can happen if you manually change CharSelectData.k_MaxLobbyPlayers to a value lower than the maximum number of players in the actual Lobby)
4. See that when you try to join again, you are allowed to and do not receive the "Player is already a part of the lobby" message
### Questions or comments

Something to note with this and with #515 is that if the host tries to remove a player at the same time as they are leaving, or vice-versa, it will generate an exception inside the Lobby service saying that it could not find the player. We may want to either catch and ignore that specific exception (not sure if it's a good idea) or maybe see with Lobby if a RemovePlayerAsync alternative could be added that doesn't generate exceptions in that case?
### Contribution checklist


Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ServerGameNetPortal.cs
```diff
@@ -201,53 +204,42 @@ void ApprovalCheck(byte[] connectionData, ulong clientId, NetworkManager.Connect
...
-                : ConnectStatus.LoggedInAgain;
-
-            //Test for Duplicate Login.
-            if (gameReturnStatus == ConnectStatus.LoggedInAgain)
+            else
             {
-                SessionPlayerData? sessionPlayerData =
-                    SessionManager<SessionPlayerData>.Instance.GetPlayerData(connectionPayload.playerId);
-
-                ulong oldClientId = sessionPlayerData?.ClientID ?? 0;
```
SamuelBellomo, line 232: What happened to this logic?

LPLafontaineB: We discussed it a while ago while I was working on #482 . We want to deny the new connection, not kick the first one. Anyways, this code did not work, it sent a disconnect reason to the old client, but disconnected the new one, so the old client didn't do anything with the disconnect reason message and the new denied one did not know why they were disconnected

SamuelBellomo: oh ok then, I thought this was addressed already. ok!

pdeschain (approved):

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ServerGameNetPortal.cs
LPLafontaineB: It is, but 482 is still in review since it won't be needed for GDC

PR #563, opened 2022-03-16
fix: null refs preventing users from quitting

LPLafontaineB:
### Description (*)

This PR simply adds a try/catch to LeaveBeforeQuit to make sure that we quit even if exceptions happen.

### Issue Number(s) (*)

Fixes issue(s): [MTT-2656](https://jira.unity3d.com/browse/MTT-2656)
### Manual testing scenarios

Sadly, we haven't found repro steps to MTT-2656
### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Scripts/Shared/ApplicationController.cs
+++ b/Assets/BossRoom/Scripts/Shared/ApplicationController.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections;
 using BossRoom.Scripts.Shared.Net.UnityServices.Auth;
 using Unity.Multiplayer.Samples.BossRoom.Client;
@@ -6,7 +7,6 @@ using Unity.Multiplayer.Samples.BossRoom.Shared.Infrastructure;
 using Unity.Multiplayer.Samples.BossRoom.Shared.Net.UnityServices.Infrastructure;
 using Unity.Multiplayer.Samples.BossRoom.Shared.Net.UnityServices.Lobbies;
 using Unity.Multiplayer.Samples.Utilities;
-using Unity.Netcode;
 using UnityEngine;
 using UnityEngine.SceneManagement;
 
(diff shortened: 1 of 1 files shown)


LPLafontaineB:
Not sure if this should be marked as a workaround or if this should be later cherrypicked to develop. What do you think?

SamuelBellomo (approved):

SamuelBellomo:
added the cherrypick label

pdeschain (approved):
discussed during standup!

Assets/BossRoom/Scripts/Shared/ApplicationController.cs
```diff
@@ -88,7 +88,15 @@ private void OnDestroy()
         /// </summary>
         private IEnumerator LeaveBeforeQuit()
         {
-            m_LobbyServiceFacade.ForceLeaveLobbyAttempt();
+            // We want to quit anyways, so if anything happens while trying to leave the Lobby, log the exception then carry on
+            try
+            {
+                m_LobbyServiceFacade.ForceLeaveLobbyAttempt();
```
fernando-cortez, line 94: Doesn't this method just end up calling `LobbyServiceFacade.EndTracking`? 

So I was looking into this this [bug](https://jira.unity3d.com/browse/MTT-2799) here, and I found success when hooking into the `onFailure` Action on some of these Lobby API calls, even when in an offline state. And looking at `EndTracking()`, it could only possibly return an `AsyncException` when trying to leave the lobby and failing. I have a suspicion that is the error we're seeing here. Wouldn't it make sense to just call `EndTracking()` here, modify `EndTracking` to allow passing in a `onFailure` action, and here quit the application when failing? @SamuelBellomo @pdeschain

LPLafontaineB: This is likely another bug. From the screenshots we had regarding this bug, the users had `NullReferenceException`s. If returning an AsyncException when failing to leave the lobby, wouldn't that not impact the flow here since it is done asynchronously?

fernando-cortez: I see both on the referenced Jira ticket. It's a shame we didn't capture the `NullReferenceException`s.

Since it's hard to reproduce, why not include a try-catch inside of `OnWantToQuit()`. As well, why are we starting the `LeaveBeforeQuit()` coroutine even when `canQuit` is `true`?

fernando-cortez: And my first point is that there's a redundancy in the two methods. `ForceLeaveLobbyAttempt()` just calls `EndTracking()`. Why not call `EndTracking()` directly?

PR #573, opened 2022-03-17
fix: hard coding frame rate

SamuelBellomo:
### Description (*)
setting to 120 so it doesn't influence RTT too much, but still shaves off about 1/3 of CPU usage when running 2 builds on the same machine

We'll need to test this on Android, make sure this doesn't mess things up. @fernando-cortez @jilfranco-unity if one of you guys could test it on your devices, that'd be great :)


fernando-cortez (commented):
Tested on a pixel 3, which has a refresh rate of 60 Hz, and this was playing nicely, mostly since it gets capped at 60. 

Per the docs on `targetFrameRate`, it is usually defaulted to 30 Hz on mobile devices. Pixel 6 will have a 90 Hz refresh rate, but I feel like you'll see diminishing returns with that so high. Capping this to 60 on mobile devices I think should be enough.

PR #399, opened 2021-11-02
fix: dead imps spawning on late join

LPLafontaineB:
### Description (*)

This PR fixes the issue where clients connecting to an ongoing game would see imps that have already been killed and despawned by destroying the associating gameobjects on start if they are not spawned.
### Issue Number(s) (*)

Fixes issue(s): MTT-1558
### Manual testing scenarios

1. ...
2. ...
### Contribution checklist
 - [ ] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [ ] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds are green)

PR #574, opened 2022-03-18
fix: imps self-destroying on clients when late-joining with multiple additive scenes loaded

LPLafontaineB:
### Description (*)

This PR makes sure ClientCharacterVisualization is only enabled on clients when it is spawned. This prevents Update from being called at the wrong time, causing it to self-destruct. This PR also removes an unneeded property and moves the OnDestroy code to OnNetworkDespawn to better mirror its initialization in OnNetworkSpawn.
### Issue Number(s) (*)

Fixes issue(s): [MTT-2433](https://jira.unity3d.com/browse/MTT-2433)
### Manual testing scenarios

1. Start a game as a host
2. Start the game and move to the second door to load the TransitionArea additive scene
3. As a client, join the game instance
4. See that the imps in the first room are visible
### Contribution checklist


Assets/BossRoom/Scripts/Client/Game/Action/FXProjectileTargetedActionFX.cs
```diff
@@ -120,7 +120,7 @@ private NetworkObject GetTarget()
         /// </summary>
         private bool IsParentAnNPC()
         {
-            var targetable = m_Parent.Parent.GetComponent<ITargetable>();
```
LukeStampfli, line 123: Why are we changing this from the clientvis Parent to the transform.Parent. From my understanding the Parent property of the client visualization is not the actual parent of the object? XML doc says:
```
        /// <summary>
        /// Returns our pseudo-Parent, the object that owns the visualization.
        /// (We don't have an actual transform parent because we're on a top-level GameObject.)
        /// </summary>
        public Transform Parent { get; private set; }
```

LPLafontaineB: It got changed some time ago with a refactor of the characters. Now the clientVis is a child of the actual character, so this property is no longer necessary.

fernando-cortez: I believe you can retrieve that from `m_Parent.NetState.IsNpc`, right?

Assets/BossRoom/Scripts/Client/Game/Action/TargetActionFX.cs
```diff
@@ -22,7 +22,7 @@ public TargetActionFX(ref ActionRequestData data, ClientCharacterVisualization p
         public override bool Start()
         {
             base.Start();
-            m_ParentState = m_Parent.Parent.GetComponent<NetworkCharacterState>();
+            m_ParentState = m_Parent.transform.parent.GetComponent<NetworkCharacterState>();
```
fernando-cortez, line 25: We could cache a reference to the parent's `NetworkCharacterState`, and remove parent entirely, since it seems like it's deprecated (removes two GetComponents).

Assets/BossRoom/Scripts/Client/Game/Action/FXProjectileTargetedActionFX.cs
LPLafontaineB: I just noticed the removal of that property was not included in my commit. Fixed it now.

Assets/BossRoom/Scripts/Client/Game/Character/ClientCharacterVisualization.cs
```diff
@@ -105,7 +99,7 @@ public override void OnNetworkSpawn()
             {
                 name = "AvatarGraphics" + m_NetState.OwnerClientId;
 
-                if (Parent.TryGetComponent(out ClientAvatarGuidHandler clientAvatarGuidHandler))
+                if (transform.parent.TryGetComponent(out ClientAvatarGuidHandler clientAvatarGuidHandler))
```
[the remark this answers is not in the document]
LPLafontaineB: done

Assets/BossRoom/Scripts/Client/Game/Action/FXProjectileTargetedActionFX.cs
fernando-cortez: Looks good to me. @LukeStampfli I'll let you resolve.

SamuelBellomo:
Noel is looking at this bug here https://jira.unity3d.com/browse/MTT-2924
Should we put this PR on hold?

LPLafontaineB:
> Noel is looking at this bug here https://jira.unity3d.com/browse/MTT-2924 Should we put this PR on hold?

This PR is not related to that bug (I think you have #399 in mind). This is for an issue on our side, where in the case of multiple scenes being loaded additively, when synchronizing, there are a few frames on the client where Update and FixedUpdate are called before OnNetworkSpawn happens.

SamuelBellomo (approved):
approving once other conversations are solved

LukeStampfli (approved):

PR #576, opened 2022-03-21
feat: Environment layer pathing [MTT-2412], [MTT-2830], [MTT-1990]

fernando-cortez:
### Description (*)
Borrowing description from [543](https://github.com/Unity-Technologies/com.unity.multiplayer.samples.coop/pull/543), which will be closed.

This is a pathfinding solution that ties player navigation more rigidly with the navigation mesh generated.

A significant change here is the re-layering of Dungeon prefabs. A `Environment` layer has been introduced to distinguish between Ground vs Environment (non-interactive) level prefabs. To prevent these items from being added to the nav mesh bake, Environment prefabs have been identified with `NavMeshModifier` components. 

Navigation is only successful when any "Ground"-tagged object is intersected via raycast, as well as when the intersection point is a valid point on the generated nav mesh. This eliminates any teleports to the tops of any ground colliders, since the point of interest will always be on the navmesh floor.

While testing, it was also discovered that the rogue could dash outside the level. If there was any part of the navmesh that bled outside the bounds of the level (very few cases), the dash could travel to that point.

To combat this, a minor tweak was added to `DynamicNavPath` which eliminates appending the last point to the nav path (since this point could actually be out of bounds).

An added bonus from this PR is that the rogue no longer goes through the door, since the visibility check inside `ActionUtils` now uses the Environment layer as a blocking layer (as well as for the mage shot). The collision detection for arrows has also been accounted for.

A last benefit from this PR is that the pillars in the boss room area have been modified with a NavMeshObstacle that, when destroyed, fills in the navmesh surface underneath. This means players will path around these pillars when not broken, and can path through these spots when destroyed.

### Related Pull Requests
This is work on top of PR [542](https://github.com/Unity-Technologies/com.unity.multiplayer.samples.coop/pull/542).

### Issue Number(s) (*)

Fixes issue(s): [MTT-2412](https://jira.unity3d.com/browse/MTT-2412), [MTT-2830](https://jira.unity3d.com/browse/MTT-2830), [MTT-1990](https://jira.unity3d.com/browse/MTT-1990)
### Manual testing scenarios

1. Select the rogue.
2. Pick a point on top of a pillar or out of bounds.
3. Press dash action.
4. Rogue will travel to the last point inside the level before the spot chosen.
### Questions or comments
Just a comment that `ClientInputSender` is beginning to become too much of a god class. This needs to be broken down to only process input and publish results to listeners.

### Contribution checklist


LPLafontaineB (dismissed):

Assets/BossRoom/Scripts/Client/Game/Character/ClientInputSender.cs
```diff
@@ -169,12 +176,30 @@ void FixedUpdate()
                 {
                     m_LastSentMove = Time.time;
                     var ray = m_MainCamera.ScreenPointToRay(Input.mousePosition);
-                    if (Physics.RaycastNonAlloc(ray, k_CachedHit, k_MouseInputRaycastDistance, k_GroundLayerMask) > 0)
+
+                    // clear previous results
+                    for (int i = 0; i < k_CachedHit.Length; i++)
```
SamuelBellomo, line 181: This for loop shouldn't be necessary if you just iterate over the first items until you reach "groundHits". with non-alloc physics calls, you really don't need to clear anything from one run to another. As long as you iterate from 0 to the returned number of items, you should be fine

SamuelBellomo (approved):

Assets/BossRoom/Scripts/Client/Game/Character/ClientInputSender.cs
fernando-cortez: This is correct. Old `RaycastHit`s don't get cleared, but the sorting I did below accounted for this anyways.

LPLafontaineB (approved):

SamuelBellomo (approved):

Assets/BossRoom/Scripts/Client/Game/Character/ClientInputSender.cs
SamuelBellomo: oh wait true, there's a sort. How does it take it into account? It just orders by distance. So if an old collision is closer, wouldn't get still be included in this?

fernando-cortez: You can pass an `index` and a `length` to the sort method. So long as I only sort until `groundHits`, any hits from 0 to `groundHits` inside `k_CachedHit` will be valid, regardless of what was searched the previous occurrence.

SamuelBellomo: And that's what happens when you read too quickly.
thanks

fernando-cortez: No prob!

PR #577, opened 2022-03-22
fix: Generate unique Profile ID on clone project

bendoyon:
### Description (*)
Running BossRoom on two instances locally, either when cloning the project manually, or when using Virtual Projects. When attempting to host in one Editor and then join a game in the other Editor, we get a conflict with the message "You have logged in elsewhere using the same account." since they are both trying to join using the same user ID.

When running in the Editor and not a ParrelSync clone, make a unique ID from the Application.dataPath. This will work for cloning projects manually, or with Virtual Projects. Since only a single instance of the Editor can be open for a specific dataPath, uniqueness is ensured.

### Issue Number(s) (*)

Fixes issue(s):
### Contribution checklist
 - [x] All automated tests passed successfully (all builds are green)

### DIFF
--- a/Assets/BossRoom/Scripts/Shared/ProfileManager.cs
+++ b/Assets/BossRoom/Scripts/Shared/ProfileManager.cs
@@ -1,6 +1,9 @@
 using UnityEngine;
 
 #if UNITY_EDITOR
+using System;
+using System.Security.Cryptography;
+using System.Text;
 using ParrelSync;
 #endif
 
@@ -31,17 +34,28 @@ namespace Unity.Multiplayer.Samples.BossRoom.Shared
 
                 return hardcodedProfileID;
             }
-#else
-        var arguments = System.Environment.GetCommandLineArgs();
-        for (int i = 0; i < arguments.Length; i++)
-        {
-            if (arguments[i] == AuthProfileCommandLineArg)
-            {
-                var profileId = arguments[i + 1];
-                return profileId;
-            }
-        }
+
+            // When running in the Editor and not a ParrelSync clone, make a unique ID
+            // from the Application.dataPath. This will work for cloning projects
+            // manually, or with Virtual Projects. Since only a single instance of
+            // the Editor can be open for a specific dataPath, uniqueness is ensured.
+            var hashedBytes = new MD5CryptoServiceProvider()
+                .ComputeHash(Encoding.UTF8.GetBytes(Application.dataPath));
+            Array.Resize(ref hashedBytes, 16);
+            return new Guid(hashedBytes).ToString("N");
+
 #endif
+
+            var arguments = System.Environment.GetCommandLineArgs();
+            for (int i = 0; i < arguments.Length; i++)
+            {
+                if (arguments[i] == AuthProfileCommandLineArg)
+                {
+                    var profileId = arguments[i + 1];
+                    return profileId;
+                }
+            }
+
             return "";
         }
     }
--- a/Assets/BossRoom/Scripts/Shared/ProfileManager.cs
+++ b/Assets/BossRoom/Scripts/Shared/ProfileManager.cs
@@ -30,10 +30,10 @@ namespace Unity.Multiplayer.Samples.BossRoom.Shared
 
 #if UNITY_EDITOR
 
-            // When running in the Editor and not a ParrelSync clone, make a unique ID
-            // from the Application.dataPath. This will work for cloning projects
-            // manually, or with Virtual Projects. Since only a single instance of
-            // the Editor can be open for a specific dataPath, uniqueness is ensured.
+            // When running in the Editor make a unique ID from the Application.dataPath.
+            // This will work for cloning projects manually, or with Virtual Projects.
+            // Since only a single instance of the Editor can be open for a specific
+            // dataPath, uniqueness is ensured.
             var hashedBytes = new MD5CryptoServiceProvider()
                 .ComputeHash(Encoding.UTF8.GetBytes(Application.dataPath));
             Array.Resize(ref hashedBytes, 16);
--- a/Assets/BossRoom/Scripts/Shared/ProfileManager.cs
+++ b/Assets/BossRoom/Scripts/Shared/ProfileManager.cs
@@ -19,21 +19,17 @@ namespace Unity.Multiplayer.Samples.BossRoom.Shared
 
         static string GetProfile()
         {
-#if UNITY_EDITOR
-
-            //The code below makes it possible for the clone instance to log in as a different user profile in Authentication service.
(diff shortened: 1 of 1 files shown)


LPLafontaineB (commented):
Nice! I suppose this would work when using ParrelSync too and could replace the arguments we are using? Unless we want to keep it as an option to be able to easily test using the same profile on multiple instances?

bendoyon:
> Nice! I suppose this would work when using ParrelSync too and could replace the arguments we are using? Unless we want to keep it as an option to be able to easily test using the same profile on multiple instances?

It would! I wasn't sure how you wanted to manage this and did not want to change the behavior, but we could simply remove the if (IsClone) block if that's what you want.

LPLafontaineB:
> > Nice! I suppose this would work when using ParrelSync too and could replace the arguments we are using? Unless we want to keep it as an option to be able to easily test using the same profile on multiple instances?
> 
> It would! I wasn't sure how you wanted to manage this and did not want to change the behavior, but we could simply remove the if (IsClone) block if that's what you want.

In that case I think we should use this version with ParrelSync as well. We can still use builds if we want to test having the same profile in different instances, and I don't think that is something that will need to be tested often anyways. What do you guys think? @pdeschain @fernando-cortez @SamuelBellomo

SamuelBellomo:
> > > Nice! I suppose this would work when using ParrelSync too and could replace the arguments we are using? Unless we want to keep it as an option to be able to easily test using the same profile on multiple instances?
> > 
> > 
> > It would! I wasn't sure how you wanted to manage this and did not want to change the behavior, but we could simply remove the if (IsClone) block if that's what you want.
> 
> In that case I think we should use this version with ParrelSync as well. We can still use builds if we want to test having the same profile in different instances, and I don't think that is something that will need to be tested often anyways. What do you guys think? @pdeschain @fernando-cortez @SamuelBellomo

Agreed, the less code paths we need to maintain the better. If we can have one single solution to manage multiple auth profiles, that'd be great :) @bendoyon have fun with the delete key :)

bendoyon:
> > > > Nice! I suppose this would work when using ParrelSync too and could replace the arguments we are using? Unless we want to keep it as an option to be able to easily test using the same profile on multiple instances?
> > > 
> > > 
> > > It would! I wasn't sure how you wanted to manage this and did not want to change the behavior, but we could simply remove the if (IsClone) block if that's what you want.
> > 
> > 
> > In that case I think we should use this version with ParrelSync as well. We can still use builds if we want to test having the same profile in different instances, and I don't think that is something that will need to be tested often anyways. What do you guys think? @pdeschain @fernando-cortez @SamuelBellomo
> 
> Agreed, the less code paths we need to maintain the better. If we can have one single solution to manage multiple auth profiles, that'd be great :) @bendoyon have fun with the delete key :)

Done! I moved things around a little bit. I figured we'd probably want to use the command line argument if it's there, even when using a clone. Custom command line parameters are already supported by virtual projects, even if it's via the API and not he UI yet.

Tested the new code with both VP and ParrelSync and it works for  both.

Assets/BossRoom/Scripts/Shared/ProfileManager.cs
```diff
@@ -16,32 +18,29 @@ public static class ProfileManager
...
 
-                //second argument is our custom ID, but if it's not set we would just use some default.
+#if UNITY_EDITOR
 
-                var hardcodedProfileID = customArguments.Length > 1 ? customArguments[1] : "defaultCloneID";
+            // When running in the Editor and not a ParrelSync clone, make a unique ID
+            // from the Application.dataPath. This will work for cloning projects
+            // manually, or with Virtual Projects. Since only a single instance of
+            // the Editor can be open for a specific dataPath, uniqueness is ensured.
+            var hashedBytes = new MD5CryptoServiceProvider()
```
SamuelBellomo, line 37: wait, does this work for builds as well? Say I have two players running at once, won't they use the same path?

bendoyon: My guess is that you build will be a specific platform and fall in the first case and not UNITY_EDITOR

SamuelBellomo: And so running two players, without the -AuthProfile arg would result in the same issue correct? Wouldn't it be best if we could have some local file where we save whether there's already a build running? I guess this could be done in another PR, I don't want to add scope to your work

bendoyon: Yep, same thing would happen, they would both try to initialize with string.Empty.

It really depends what you want the behavior to be. From just reading quickly I assumed you'd want something deterministic.

But if that's not the case, and you always want different builds to use a new ID unless specified by the command line parameter, you can simply return new Guids every time.

That would be a simple addition, but it really depends on what you folks want the behavior to be :D

Assets/BossRoom/Scripts/Shared/ProfileManager.cs
```diff
@@ -16,32 +18,29 @@ public static class ProfileManager
...
+                    var profileId = arguments[i + 1];
+                    return profileId;
+                }
+            }
 
-                //second argument is our custom ID, but if it's not set we would just use some default.
+#if UNITY_EDITOR
 
-                var hardcodedProfileID = customArguments.Length > 1 ? customArguments[1] : "defaultCloneID";
+            // When running in the Editor and not a ParrelSync clone, make a unique ID
```
LPLafontaineB, line 33: Since this now works with ParrelSync clones too, we might want to change this comment

Assets/BossRoom/Scripts/Shared/ProfileManager.cs
SamuelBellomo: mmm I see, so we'd have new users every single time we run the game?
I can see this breaking if/when we integrate UGS services like Cloud Save.
And even if our own session management, we're basing ourselves on auth id, which means a game client crashing wouldn't be able to continue its progress when connecting back.
And thinking of it, managing this with local files wouldn't be trivial either. If a client crashes and we reopen a new one, how do we know which of the locally saved profiles it was using?
I guess we could heartbeat in use GUIDs and then when reconnecting just reuse the oldest one. again not trivial and I don't want to explode the scope of this PR

bendoyon: In this case I would suggest merging this and addressing the scenario in the future in the following ways:

1. If you don't care about creating new users every time, you can generate random IDs.
2. If you want to reuse specific IDs, you need a way to manage them and assign them to game instances, whether they're Editor instances or player builds.

LPLafontaineB (approved):

PR #578, opened 2022-03-23
Fix/UI blocking pass

jilfranco-unity:
### Description (*)
This PR is for a fix for the HUD UI buttons letting movement input through on mobile. After doing a bunch of looking through our settings, building, and testing, the one thing that works is a simple code fix. In fixed update, right before our movement input logic, I use the Event System to check if the pointer is over a game object or if there is currently a selected game object. If there is, we just return. If not, we do the movement logic.

I've also done a bit of clean up on our lobby UI, applying any overrides in the scene to the prefabs, unchecking the raycast target option on some of our graphics, and deleting some unneeded vars in the Lobby and IP UI Mediators leftover from the old lobby UI

Fixes issue(s):
[Here's the Jira ticket](https://jira.unity3d.com/browse/MTT-2890)

### DIFF
--- a/Assets/BossRoom/Scripts/Client/Game/Character/ClientInputSender.cs
+++ b/Assets/BossRoom/Scripts/Client/Game/Character/ClientInputSender.cs
@@ -176,10 +176,10 @@ namespace Unity.Multiplayer.Samples.BossRoom.Client
                     var ray = m_MainCamera.ScreenPointToRay(Input.mousePosition);
                     if (Physics.RaycastNonAlloc(ray, k_CachedHit, k_MouseInputRaycastDistance, k_GroundLayerMask) > 0)
                     {
-                            m_NetworkCharacter.SendCharacterInputServerRpc(k_CachedHit[0].point);
+                        m_NetworkCharacter.SendCharacterInputServerRpc(k_CachedHit[0].point);
 
-                            //Send our client only click request
-                            ClientMoveEvent?.Invoke(k_CachedHit[0].point);
+                        //Send our client only click request
+                        ClientMoveEvent?.Invoke(k_CachedHit[0].point);
                     }
                 }
             }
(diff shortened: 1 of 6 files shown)


jilfranco-unity:
Made a tiny change per some feedback from LP, instead of checking the event system to see the pointer is over a game object and if something is currently selected, now it only checks for if something is currently selected. The pointer check was a bit redundant. 

A question tho, should this logic happen on all platforms? Or just mobile since the problem only seems to happen on mobile?

LPLafontaineB (dismissed):

Assets/BossRoom/Scripts/Client/Game/Character/ClientInputSender.cs
```diff
@@ -161,6 +161,11 @@ void FixedUpdate()
             }
 
             m_ActionRequestCount = 0;
+            
+            if (EventSystem.current.currentSelectedGameObject != null)
```
fernando-cortez, line 165: So this works great for preventing movement. Though there's some actions (namely Archer base attack, mage bolt, mage heal, and rogue's base attack) where it seems like a position under the action UI object is set as the look vector. It would likely be a per action change, as well so not a blocking issue for this PR.

fernando-cortez (commented):
The last thing I'd add to this PR is modifying the popup window (I believe it is PopupPrefab). I can hit through the black background and hit the "Start IP" button. 

Also, I don't think I was able to test, but also look into the two buttons on the top right (the exit button and the volume/quality button). There could be a scenario where you could hit through the black background and hit a lobby listed for instance.

jilfranco-unity:
> This ended up being a really simple fix, just needed to make the background of the pop up panel a raycast target so that it blocks input to other ui elements! This fixes the issue on both desktop and android--I think because it's not a physics raycast.

LPLafontaineB (approved):

fernando-cortez (approved):

PR #579, opened 2022-03-23
feat: simpler quitting flow

LPLafontaineB:
### Description (*)

This PR implements the same feature as #537 did, but without checking scene names in code. To do this, the SettingsPanelCanvas is removed from the DDOL scene, and added to every scene. An enum field in QuitPanel decides if we want to return to the menu, or quit the application, so each scene can specify the expected behavior. In this case, in every scene except the MainMenu, we return to the menu when using the quit button, and in the MainMenu scene, we quit the application.
### Issue Number(s) (*)

Fixes issue(s): [MTT-2919](https://jira.unity3d.com/browse/MTT-2919)
### Manual testing scenarios

1. Launch the game and use the quit button in the different scenes
2. See that it brings you back to the main menu in all scenes except the MainMenu scene, where it quits the application
### Contribution checklist


Assets/BossRoom/Scripts/Client/UI/UIQuitPanel.cs
```diff
@@ -1,41 +1,48 @@
...
             m_ApplicationController = applicationController;
         }
 
-        void OnEnable()
+        void Awake()
         {
-            m_QuitMode = NetworkManager.Singleton == null || !NetworkManager.Singleton.IsListening;
-            m_QuitButtonText.text = m_QuitMode ? "Exit Game?" : "Leave session?";
+            var scope = DIScope.RootScope;
+            scope.InjectIn(gameObject);
```
pdeschain, line 31: I would much rather see this injection in ApplicationController - this way we are spreading initialization around. How do you know for sure that DIScope.RootScope is fully constructed at the point of Awake? Can't have that guarantee unless you have it in one spot.

pdeschain (changes requested):

Assets/BossRoom/Scripts/Client/UI/UIQuitPanel.cs
LPLafontaineB: Makes sense. I'm not sure how to go about doing it though, since this is now moved to different scenes. It used to be done through ApplicationController's `m_GameObjectsThatWillBeInjectedAutomatically`. What would be the best way to do this without `m_GameObjectsThatWillBeInjectedAutomatically`?

pdeschain: If we know for a fact that DIScope.RootScope is loaded and set up in a different scene - then we're good I guess.

pdeschain (approved):

Assets/BossRoom/Scripts/Client/UI/UIQuitPanel.cs
LPLafontaineB: In that case, yes RootScope is set up before that because it is done in ApplicationController's Awake, and UIQuitPanel's Awake is called when its game object gets activated (which is when we first click on the Quit button, or when loading the MainMenu scene if we don't deactivate the Panel in the prefab. However, if we set the panel active in the prefab, and add the prefab to the Startup scene, there may be a possibility that this Awake is called before ApplicationController's. Should I move this to Start just to be sure?

Assets/BossRoom/Prefabs/UI/SettingsPanelCanvas.prefab
```diff
@@ -2571,136 +2665,168 @@ PrefabInstance:
...
+    - target: {fileID: 7721113380319664588, guid: addf3685293c35d48a673397c748f116,
+        type: 3}
       propertyPath: m_OnClick.m_PersistentCalls.m_Calls.Array.data[0].m_MethodName
       value: SetQualitySettings
       objectReference: {fileID: 0}
-    - target: {fileID: 7721113380319664588, guid: addf3685293c35d48a673397c748f116, type: 3}
+    - target: {fileID: 7721113380319664588, guid: addf3685293c35d48a673397c748f116,
+        type: 3}
       propertyPath: m_OnClick.m_PersistentCalls.m_Calls.Array.data[0].m_TargetAssemblyTypeName
       value: QualityButton, Assembly-CSharp
```
fernando-cortez, line 2706: I believe the text component here got dereferenced. There's a null ref when you try to change quality settings.

fernando-cortez (changes requested):

fernando-cortez (dismissed):

SamuelBellomo (approved):
good to go once we've solved the other conversation

fernando-cortez (dismissed):

Assets/BossRoom/Scripts/Client/UI/UIQuitPanel.cs
pdeschain: Start is safer in this case, sure!

pdeschain: My thought process regarding overall maintainability of the codebase is that as much as possible should be explicit and in code (it's version-controlled, it's readable by humans). `Awake` and `Start` functions are called for us by Unity (thus allowing us to have arbitrary entry points to our logic), but we do have both ApplicationController and various state controllers - these are the "designated entry points". Designated entry point versus non-designated entry point boils down to difference in cognitive complexity - if everything in the app follows the convention of being set up in app/state controllers, then we can rely on that fact. Breaking this assumption would bring us back to a point where we have multiple places where logic starts. << that's me being a bit of a zealot

I understand that Start would work in this case (and it probably would even be safe), however you're relying on a side-effect that for some reason we chose to do initialization in Awake, not Start. To be specific -I don't think that this assumption will be broken, I think that your approach would work.

LPLafontaineB: Updated this PR based on the discussion we had. Updated the PR description as well to reflect this change

Assets/BossRoom/Scripts/Shared/Game/State/GameStateBehaviour.cs
```diff
@@ -53,6 +55,22 @@ public virtual bool Persists
...
+
+        protected virtual void Awake()
+        {
+            InitializeScope();
+            m_Scope.FinalizeScopeConstruction();
+            foreach (var autoInjectedGameObject in m_GameObjectsThatWillBeInjectedAutomatically)
+            {
+                m_Scope.InjectIn(autoInjectedGameObject);
+            }
+            DIScope.RootScope.InjectIn(this);
```
pdeschain, line 71: That's redundant - scopes work as a tree, if the child doesn't have something - it looks in a parent scope.

Assets/BossRoom/Scripts/Shared/Game/State/GameStateBehaviour.cs
```diff
@@ -87,8 +105,14 @@ protected virtual void Start()
             }
         }
 
+        protected virtual void InitializeScope()
+        {
+            m_Scope = new DIScope(DIScope.RootScope);
```
pdeschain, line 110: I would move this line to line 65, before we call this method - this will be where people need to actually bind the dependencies. Otherwise someone might forget to create the m_Scope ?

Assets/BossRoom/Scripts/Shared/Game/State/GameStateBehaviour.cs
pdeschain, line 71: I think that this line should be the first one in the method, no? Because the InitializeScope could be referring to wahtever dependencies that are in the child class while it's running InitializeScope function

Assets/BossRoom/Scripts/Shared/Game/State/GameStateBehaviour.cs
```diff
@@ -53,6 +55,22 @@ public virtual bool Persists
         /// </summary>
         private static GameObject s_ActiveStateGO;
 
+        protected DIScope m_Scope;
```
pdeschain, line 58: we probably should make this public {get; protected set} parameter - Fernando will need access to this reference for test runner stuff

Assets/BossRoom/Scripts/Shared/Game/State/GameStateBehaviour.cs
LPLafontaineB: I expect that in this case we would get errors when running if we don't properly override this while calling base.InitializeScope. But it could be moved to Awake instead and make this method empty for the parent class. That would be one less thing for users to worry about when creating new states. We would still need to make sure that base.awake and base.start are called, but that would be it.

Assets/BossRoom/Scripts/Shared/Game/State/GameStateBehaviour.cs
LPLafontaineB: made the set private since at the moment it is only set here in Awake. We can still change that later if we find that a child class needs to set it itself

pdeschain (approved):
👍

fernando-cortez (approved):

PR #581, opened 2022-03-25
fix: decoupling SessionManager from NetworkManager [MTT-2603]

LPLafontaineB:
### Description

This PR removes dependencies that SessionManager had on NetworkManager, by removing it as a field and replacing the callbacks it registered to with public methods called from GameNetPortal. This solves the issues of the SessionManager not handling the destruction of the NetworkManager and the creation of a new one. In such a case, the SessionManager would have no idea it happened and would not unregister its callbacks and register them again on the new NetworkManager instance.

This PR also simplifies the handling of clearing data from disconnected players and reinitializing the rest when dealing with successive game sessions. We now simply do it when a session ends. Furthermore, a new field tells the SessionManager if a session is currently started, so that when a client disconnects it can decide if it needs to keep their data or not.

### Issue Number(s)

[MTT-2603](https://jira.unity3d.com/browse/MTT-2603)

### Contribution checklist


fernando-cortez (approved):

Packages/com.unity.multiplayer.samples.coop/Utilities/Net/SessionManager.cs
```diff
@@ -232,36 +207,31 @@ public void OnSessionStarted()
         public void OnSessionEnded()
```
SamuelBellomo, line 207: OnSessionEnded is called in the OnNetworkSpawn for the post game state. What happens if we ever get a disconnect host side (I would guess some day we'll get that disconnect from relay if we lose our connection). Host would in theory brought back to main menu, and would skip the post game state. Which means the session would never get "ended" correct?

SamuelBellomo: and that point, why not have the session live inside the main scene as a gameobject? (it can still be accessible statically, but as soon as you start the session, sessionManager could spawn itself in the scene to follow the scene's lifecycle

LPLafontaineB: You're right, I did not think of that possibility.

I thought about your suggestions, and I think that the cleanest solution would be to simply make sure that Clear is called when the server ends (whether by requested disconnect or if disconnected by relay) and that this sets m_HasSessionStarted to false.

I still think it's better to not confine SessionManager to a gameobject so that it can have a lifetime longer than a scene. We need this to allow us to save player session data while they are connecting, and this makes it easier to reuse for projects where a session could last more than a single main scene.

Assets/BossRoom/Scripts/Shared/Net/ConnectionManagement/ServerGameNetPortal.cs
```diff
@@ -112,6 +113,7 @@ void OnClientDisconnect(ulong clientId)
                     {
                         m_LobbyServiceFacade.RemovePlayerFromLobbyAsync(playerId, m_LobbyServiceFacade.CurrentUnityLobby.Id, null, null);
                     }
+                    SessionManager<SessionPlayerData>.Instance.DisconnectClient(clientId);
```
pdeschain, line 116: In all those instances where we're using singleton access pattern to get the SessionManager I'm thinking about using DI for this (even with keeping sessionmanager a singleton on it's own - just binding the .Instance in DI scope would do the trick). Not 100% on this, would appreciate your thoughts.

fernando-cortez (commented):
Can approve once other conversations are resolved 👍.

SamuelBellomo (approved):

pdeschain (approved):

PR #583, opened 2022-03-25
fix: dead player not dead when reconnecting [MTT-2965]

LPLafontaineB:
### Description

With this PR, NetworkCharacterState now sets its LifeState to fainted when reconnecting a fainted player character.
### Issue Number(s)

[MTT-2965](https://jira.unity3d.com/browse/MTT-2965)
### Contribution checklist


Assets/BossRoom/Scripts/Shared/Game/Entity/NetworkCharacterState.cs
```diff
@@ -113,21 +114,32 @@ public LifeState LifeState
...
+                    StartCoroutine(SetReconnectingHitPointsCoroutine(sessionPlayerData.Value.CurrentHitPoints));
                 }
             }
+        }
 
-            return CharacterClass.BaseHP.Value;
+        IEnumerator SetReconnectingHitPointsCoroutine(int hitPoints)
+        {
+            // Wait until end of frame to make sure other NetworkBehaviors for this character had time to spawn.
+            yield return new WaitForEndOfFrame();
```
fernando-cortez, line 137: Is this necessary? If other NetworkBehaviours were subscribed to the OnValueChanged callback, that should fire regardless, right? What other considerations are there?

LPLafontaineB: From my tests, it doesn't fire. ServerCharacter and ServerAnimationHandler both register to LifeState's OnValueChanged inside their OnNetworkSpawned, but if I change its value in NetworkCharacterState's OnNetworkSpawn, it is not caught by them. Concretely, this means the animation trigger for fainting is not sent and the reconnecting character is simply immobile, but still standing and doing the anticipation animations when trying to move

fernando-cortez: Hold on. After rethinking it, this should actually be done inside `ServerAnimationHandler`. You can call `OnLifeStateChanged` directly as you're subscribing to the `OnValueChanged` callback for `LifeState`. It's a common pattern in the codebase for initializing based on a `NetworkVariable`'svalue on `OnNetworkSpawn()`.

Assets/BossRoom/Scripts/Shared/Game/Entity/NetworkCharacterState.cs
```diff
@@ -113,21 +113,25 @@ public LifeState LifeState
...
             {
                 SessionPlayerData? sessionPlayerData = SessionManager<SessionPlayerData>.Instance.GetPlayerData(OwnerClientId);
                 if (sessionPlayerData is { HasCharacterSpawned: true })
                 {
-                    return sessionPlayerData.Value.CurrentHitPoints;
+                    HitPoints = sessionPlayerData.Value.CurrentHitPoints;
+                    if (HitPoints <= 0)
+                    {
+                        LifeState = IsNpc ? LifeState.Dead : LifeState.Fainted;
+                    }
```
fernando-cortez, line 132: Feels to me like initializing `HitPoints` should be moved to `ServerCharacter` completely. You could just initialize the `LifeState` there on `OnNetworkSpawn()` based on `HitPoints` and isolate the setting of `LifeState` to just that class, since both should be server-driven.

Assets/BossRoom/Scripts/Server/Game/Character/ServerCharacter.cs
```diff
@@ -72,6 +72,7 @@ public override void OnNetworkSpawn()
                 NetState.ReceivedClientInput += OnClientMoveRequest;
                 NetState.OnStopChargingUpServer += OnStoppedChargingUp;
                 NetState.NetworkLifeState.LifeState.OnValueChanged += OnLifeStateChanged;
+                OnLifeStateChanged(LifeState.Alive, NetState.NetworkLifeState.LifeState.Value);
```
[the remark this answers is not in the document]
fernando-cortez: Calling this here directly is pretty clean, IMO. Can we guarantee that `OnValueChanged` on a `NetworkBehaviour` won't be fired before `OnNetworkSpawn()`? And if it did, can you guarantee that the `IsServer` is valid as well? (since you'd have to check that per `OnValueChanged` invocation as the callback will be fired on clients + server). I personally like gating the callback to just the server.

SamuelBellomo (approved):

Assets/BossRoom/Scripts/Server/Game/Character/ServerCharacter.cs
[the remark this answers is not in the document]
LPLafontaineB: On an unrelated note, however, I found an issue with my solution (happens if the hp and lifestate is initialized on the netcharstate and if it is done in serverchar). Basically, when an unconscious player reconnects, the host will notice it and set its lifestate correctly, then sends a trigger to the NetworkAnimator, but somehow the trigger is not handled on the host, only on the client.
If instead, I initialize the hp and lifestate one frame later (i.e. via a coroutine with a yield return null), it fixes it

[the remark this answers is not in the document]
LPLafontaineB: I pushed the fix for it, but I'm not convinced by it. We may want to investigate this further to see if this is a bug from NetworkAnimator or something similar.

Assets/BossRoom/Scripts/Server/Game/Character/ServerCharacter.cs
```diff
@@ -105,6 +106,31 @@ public override void OnNetworkDespawn()
...
+            {
+                SessionPlayerData? sessionPlayerData = SessionManager<SessionPlayerData>.Instance.GetPlayerData(OwnerClientId);
+                if (sessionPlayerData is {HasCharacterSpawned: true})
+                {
+                    NetState.HitPoints = sessionPlayerData.Value.CurrentHitPoints;
+                    if (NetState.HitPoints <= 0)
+                    {
+                        // Update LifeState on next frame to make sure everything listening on it has time to initialize
+                        StartCoroutine(WaitToUpdateLifeState(LifeState.Fainted));
+                    }
```
fernando-cortez, line 123: Couldn't this coroutine be removed by applying the same pattern to `ServerAnimationHandler`? It seems to register itself to the `OnValueChanged` callback for `LifeState`, but doesn't initialize itself inside `OnNetworkSpawn()`.

pdeschain (dismissed):

pdeschain:
approved pending the conversation above

Assets/BossRoom/Scripts/Server/Game/Character/ServerCharacter.cs
LPLafontaineB: Sorry, what I meant is that I tried to directly call `OnValueChanged` inside `ServerAnimationHandler`'s `OnNetworkSpawn()`. That is what led to the bug where the animation would only play on clients, and not on the host. From what I understand, it seems to be because the `NetworkAnimator`'s `OnNetworkSpawn()` is called after `ServerAnimationHandler`'s `OnNetworkSpawn()`.

fernando-cortez: If that's the case, what do you think about moving the coroutine to that script instead? Let this class set the `LifeState` here and isolate that as a `NetworkAnimator` issue. We can create a Jira to look into that order of execution.

LPLafontaineB: Makes sense!

fernando-cortez: For visibility: [ticket](https://jira.unity3d.com/browse/MTT-3152).

fernando-cortez (approved):

Assets/BossRoom/Scripts/Server/Game/Character/ServerCharacter.cs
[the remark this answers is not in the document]
SamuelBellomo: yeah that sounds like a bug the way you're describing it, I'd expect NA to always trigger on host.
wanna create a jira and resolve this conversation after?

SamuelBellomo (approved):

Assets/BossRoom/Scripts/Server/Game/Character/ServerCharacter.cs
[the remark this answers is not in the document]
fernando-cortez: Was already done [below](https://github.com/Unity-Technologies/com.unity.multiplayer.samples.coop/pull/583#discussion_r843194452). 😃

PR #568, opened 2022-03-17
feat: better lobby exceptions handling

LPLafontaineB:
### Description (*)

This PR changes how lobby exceptions are handled. Only some of them will be shown to players through popups.
### Issue Number(s) (*)

Fixes issue(s):
### Manual testing scenarios

1. ...
2. ...
### Contribution checklist
 - [ ] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [ ] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds are green)

PR #585, opened 2022-03-29
feat: better lobby exception handling [MTT-2649]

LPLafontaineB:
### Description

This PR is mostly a cherrypick of #568, which restricted popup messages to only some lobby exceptions and removed an unnecessary call to check if a player was part of a lobby before trying to kick them.

On top of the cherrypick, I replaced the handling of 400 bad request which from my earlier tests happened when we tried to input a badly formatted user-code with ValidationError which is the exception reason we receive now.

### Issue Number(s)

[MTT-2649](https://jira.unity3d.com/browse/MTT-2649) and [MTT-2840](https://jira.unity3d.com/browse/MTT-2840)
### Contribution checklist


pdeschain (dismissed):
👍

Assets/BossRoom/Scripts/Client/UI/Lobby/LobbyJoiningUI.cs
```diff
@@ -57,9 +61,22 @@ void OnDestroy()
...
+        /// Added to the InputField component's OnValueChanged callback for the join code text.
+        /// </summary>
+        public void SanitizeJoinCodeInputText()
+        {
+            m_JoinCodeField.text = SanitizeJoinCode(m_JoinCodeField.text);
+        }
+
+        string SanitizeJoinCode(string dirtyString)
+        {
+            return Regex.Replace(dirtyString.ToUpper(), $"[^{k_LobbyJoinCodeAlphabet}]", "");
```
fernando-cortez, line 74: Why not just cache the Regex pattern?

LPLafontaineB: done

fernando-cortez (approved):

Assets/BossRoom/Scripts/Client/UI/Lobby/LobbyJoiningUI.cs
```diff
@@ -23,6 +24,10 @@ public class LobbyJoiningUI : MonoBehaviour
         IDisposable m_Subscriptions;
         List<LobbyListItemUI> m_LobbyListItems = new List<LobbyListItemUI>();
 
+        // Adding this here until it is available in the Lobby package itself
```
SamuelBellomo, line 27: what's the jira ticket for this lobby side?
Since they are going to fix this, any reason we're doing this workaround for it here at all?

Assets/BossRoom/Scripts/Client/UI/UnityServicesUIHandler.cs
```diff
@@ -49,25 +49,38 @@ void ServiceErrorHandler(UnityServiceErrorMessage error)
...
+                    case LobbyExceptionReason.Unauthorized:
+                        PopupPanel.ShowPopupPanel("Lobby error", "Received HTTP error 401 Unauthorized from Lobby Service.");
+                        break;
+                    case LobbyExceptionReason.RequestTimeOut:
+                        PopupPanel.ShowPopupPanel("Lobby error", "Received HTTP error 408 Request timed out from Lobby Service.");
+                        break;
                 }
             }
-
-            PopupPanel.ShowPopupPanel("Service error: " + error.Title, errorMessage);
```
SamuelBellomo, line 70: Should we still log the unhandled ones? Or will they be handled somewhere else?

SamuelBellomo (dismissed):

Assets/BossRoom/Scripts/Client/UI/UnityServicesUIHandler.cs
LPLafontaineB: They are already logged with the stack trace from before the async task before creating the UnityServiceErrorMessage that is published and caught here

Assets/BossRoom/Scripts/Client/UI/Lobby/LobbyJoiningUI.cs
LPLafontaineB: It was still in discussion whether it was something they wanted to expose in the SDK or not, so no Jira ticket yet.
I added this here to simply prevent us from getting the ValidationError, figuring that if we know it will fail, there is no point of actually calling the service.
However, I got some new feedback from the Lobby team about this, I'll wait for the conclusion of that conversation before merging this PR.

LPLafontaineB:
Putting on hold temporarily, just waiting for some feedback from the lobby team

Assets/BossRoom/Scripts/Client/UI/Lobby/LobbyJoiningUI.cs
LPLafontaineB: I ended up removing this and just doing sanitization with the full alphabet instead for this specific PR. If we want to do more and prevent the ValidationError, I think it should be a new ticket, following the conclusion of the aforementioned conversation

Assets/BossRoom/Scripts/Client/UI/UnityServicesUIHandler.cs
```diff
@@ -49,25 +49,38 @@ void ServiceErrorHandler(UnityServiceErrorMessage error)
 
         void HandleLobbyError(UnityServiceErrorMessage error)
         {
-            var errorMessage = error.Message;
-            switch (((LobbyServiceException)error.OriginalException).Reason)
+            var exception = (LobbyServiceException) error.OriginalException;
```
pdeschain, line 52: I think that you'd get a miscast exception if OriginalException is something else than LobbyServiceException.

pdeschain, line 52:
```suggestion
            var exception = error.OriginalException as LobbyServiceException;
```

pdeschain (changes requested):

Assets/BossRoom/Scripts/Client/UI/UnityServicesUIHandler.cs
LPLafontaineB: Normally it should not happen since we only call HandleLobbyError when error.AffectedService equals UnityServiceErrorMessage.Service.Lobby. If it does, it would mean that the UnityServiceErrorMessage was not properly constructed, right?

fernando-cortez (commented):
Can approve once comments are addressed 👍.

SamuelBellomo (dismissed):

Assets/BossRoom/Scripts/Client/UI/UnityServicesUIHandler.cs
pdeschain: Well, HandleLobbyError method is presuming that no other UnityServices will ever use UnityServiceErrorMessage or something like that That's a wrong assumption. Exception is the type of OriginalException. Everything else is just a possibility.

SamuelBellomo (approved):

pdeschain (dismissed):

SamuelBellomo (approved):

pdeschain (approved):

PR #593, opened 2022-04-06
fix: arrow despawned rpc [MTT-3134]

fernando-cortez:
### Description
Without repro steps for the attached ticket, this PR should address the issue for RPCs sent from an arrow on the host. Basically, OnNetworkDespawn() never cleared a flag that prevented any FixedUpdate() logic, thus potentially sending an RPC on a despawned object. This has now been taken care of and will be extensively tested in next test session.

### Issue Number(s)
[[MTT](https://jira.unity3d.com/browse/MTT-3134)](https://jira.unity3d.com/browse/MTT-3134)

### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Scripts/Server/Game/Entity/ServerProjectileLogic.cs
+++ b/Assets/BossRoom/Scripts/Server/Game/Entity/ServerProjectileLogic.cs
@@ -82,6 +82,11 @@ namespace Unity.Multiplayer.Samples.BossRoom.Server
             m_NpcLayer = LayerMask.NameToLayer("NPCs");
         }
 
+        public override void OnNetworkDespawn()
+        {
+            m_Started = false;
+        }
+
         void FixedUpdate()
         {
             if (!m_Started)


Assets/BossRoom/Scripts/Server/Game/Entity/ServerProjectileLogic.cs
```diff
@@ -82,6 +82,11 @@ public override void OnNetworkSpawn()
             m_NpcLayer = LayerMask.NameToLayer("NPCs");
         }
 
+        public override void OnNetworkDespawn()
+        {
+            m_Started = false;
```
SamuelBellomo, line 87: Why not just use "IsSpawned"?

fernando-cortez: Performance. These types of Netcode API calls on FixedUpdates really accumulate on lower-end devices.

SamuelBellomo: oh because of NetworkBehaviour's get on NetworkObject which is doing a bunch of things? right. This is dumb, we should really have a way to do something like Get NetworkObjectUnchecked, something like that. Anyway, ok approving this PR

SamuelBellomo (approved):

PR #597, opened 2022-04-07
fix: silencing errors that are not errors [MTT-2930] [MTT-3249]

LPLafontaineB:
### Description

This is a quick PR to prevent RemovePlayerFromLobbyAsync from throwing a PlayerNotFound exception, which is something that is already handled. It also changes the error logs for when SessionManager does not find a player's session data to regular logs, since this is also already handled game-side through the null return value.
### Issue Number(s)

[MTT-3249](https://jira.unity3d.com/browse/MTT-3249) & [MTT-2930](https://jira.unity3d.com/browse/MTT-2930)

### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Scripts/Shared/Net/UnityServices/Lobbies/LobbyAPIInterface.cs
+++ b/Assets/BossRoom/Scripts/Shared/Net/UnityServices/Lobbies/LobbyAPIInterface.cs
@@ -119,8 +119,8 @@ namespace Unity.Multiplayer.Samples.BossRoom.Shared.Net.UnityServices.Lobbies
             {
                 await ExceptionHandling(Lobbies.Instance.RemovePlayerAsync(lobbyId, requesterUasId));
             }
-            catch (Exception e)
-                when (e is LobbyServiceException {Reason: LobbyExceptionReason.PlayerNotFound})
+            catch (LobbyServiceException e)
+                when (e is {Reason: LobbyExceptionReason.PlayerNotFound})
             {
                 // If Player is not found, they have already left the lobby or have been kicked out. No need to throw here
             }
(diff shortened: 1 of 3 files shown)


LPLafontaineB:
Thank you to @bendoyon for suggesting this solution for RemovePlayerFromLobbyAsync

Assets/BossRoom/Scripts/Shared/Net/UnityServices/Lobbies/LobbyAPIInterface.cs
```diff
@@ -115,7 +115,15 @@ public async Task<Lobby> QuickJoinLobby(string requesterUasId, Dictionary<string
 
         public async Task RemovePlayerFromLobby(string requesterUasId, string lobbyId)
         {
-           await ExceptionHandling(Lobbies.Instance.RemovePlayerAsync(lobbyId, requesterUasId));
+            try
+            {
+                await ExceptionHandling(Lobbies.Instance.RemovePlayerAsync(lobbyId, requesterUasId));
+            }
+            catch (Exception e)
```
bendoyon, line 122: You can probably just catch LobbyServiceException specifically here, instead of generic Exception, then do the reason check in the filter

LPLafontaineB: Currently, this message is only subscribed to by the utility that creates popup messages, and it already ignores PlayerNotFound, so it is not necessary, but we might want later to have something else subscribe to it and do something with this specific exception

pdeschain (approved):

fernando-cortez (approved):

PR #599, opened 2022-04-07
fix: Alpha4 input press doesn't open emote drawer on UI [MTT-3236]

fernando-cortez:
### Description
Input was never detected for opening emote drawer. An input check for Alpha4 has been added to `HeroActionBar` to open/close emote drawer.

### Issue Number(s)
[MTT-3236](https://jira.unity3d.com/browse/MTT-3236)

### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Scripts/Client/UI/HeroActionBar.cs
+++ b/Assets/BossRoom/Scripts/Client/UI/HeroActionBar.cs
@@ -197,6 +197,11 @@ namespace Unity.Multiplayer.Samples.BossRoom.Visual
             // If we have another player selected, see if their aliveness state has changed,
             // and if so, update the interactiveness of the basic-action button
 
+            if (Input.GetKeyUp(KeyCode.Alpha4))
+            {
+                m_ButtonInfo[ActionButtonType.EmoteBar].Button.OnPointerUpEvent.Invoke();
+            }
+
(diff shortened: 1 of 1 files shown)


LPLafontaineB (approved):

Assets/BossRoom/Scripts/Client/UI/HeroActionBar.cs
```diff
@@ -197,6 +197,11 @@ void Update()
             // If we have another player selected, see if their aliveness state has changed,
             // and if so, update the interactiveness of the basic-action button
 
+            if (Input.GetKeyUp(KeyCode.Alpha4))
```
SamuelBellomo, line 200: any reason we're not managing this the same way we're managing other shortcuts? Why the special case here?

fernando-cortez: You mean why is this not included inside `ClientInputSender`? The other actions (1-3) are sent as request RPCs from `ClientInputSender`, and on top of that we've also added `OnButtonDown` callbacks here for UI button clicks. We've simply just not added a response from input on Alpha4, and I certainly think it should be kept here instead of `ClientInputSender`, since it's just enabling/disabling a GameObject.

SamuelBellomo (approved):

PR #601, opened 2022-04-08
feat: in-game message feed [MTT-2678] [MTT-2318]

LPLafontaineB:
### Description

This PR adds a simple in-game message feed and some message channels for in-game events. A message feed is added inside the CharSelect, BossRoom and PostGame scenes that subscribes to those messages to display them. No history is being kept.

The Feed itself is placed on the left side of the screen in the CharSelect scene, and on the right side in the BossRoom scene and the PostGame scene.

Each message is displayed individually and disappears after a specific amount of time.

Messages are published when players connect/disconnect from a game, when their lifestate changes, when the door opens/closes, when the boss is defeated and when a player uses a cheat.

This PR also uses those new message channels to simplify ServerBossRoomState, which no longer needs to register to every player character's NetworkedLifeState's OnValueChanged and no longer needs to use the GameEvent scriptable object to detect when the boss is defeated.

### Issue Number(s)

[MTT-2678](https://jira.unity3d.com/browse/MTT-2678) & [MTT-2318](https://jira.unity3d.com/browse/MTT-2318)
### Contribution checklist


Assets/BossRoom/Scripts/Shared/ApplicationController.cs
```diff
@@ -52,6 +53,9 @@ private void Awake()
             //this message channel is essential and persists for the lifetime of the lobby and relay services
             scope.BindMessageChannelInstance<ConnectStatus>();
 
+            //this message channel is essential and persists for the lifetime of the lobby and relay services
+            scope.BindMessageChannelInstance<MessageFeedHandler.MessageFeed>();
```
SamuelBellomo, line 57: So this is a trap of message feeds we shouldn't fall into.
There should be a message queue for in game events, there should be one for connection events. And the feed UI should hook into those different message queues (just like we could have some graphic updates from those same events). We shouldn't have one big message queue with a bunch of different message types in it.

SamuelBellomo: This way those queues can be reused by other consumers that are not the feed UI

LPLafontaineB: I see. I went that way mainly to solve two constraints. First, a lot of these messages are published locally on the different clients and the server, but here we want to replicate all these messages to all clients. Second, I wanted to keep a history of messages that would live throughout the application that each instance of UIMessageFeed could listen to. So here, this message channel is used to publish the feed history that is constructed client-side from messages created on the server and sent through rpcs. The actual messages are sent through MessageFeedHandler as the singleton pattern.

I'll rework this and look into replacing my singleton use with DI and create a new message channel for in-game messages. I'll also try to think of a different solution to solve those two constraints without a specific message channel to publish the feed history

SamuelBellomo: I see.
We might need a new message channel type then. This would be a great use for a networked message channel, that'd distribute your messages not only locally but to clients as well, this could be a great use case for this.
cc @LukeStampfli @pdeschain we already discussed this

SamuelBellomo: Why do you want to keep a history, not sure I understood that part.

LPLafontaineB: I figured that we could want to look at older messages, like for examples in some games you can open the game feed and look at older chat messages or events like the kill log in first person shooters

LPLafontaineB: If we don't want this though, I can make this much simpler and only keep the history in UIMessageFeed which would be cleared after each scene switch

LPLafontaineB: I'm not sure what the point of having a buffered message channel would be here. What makes the latest message interesting, but not those before it?

LPLafontaineB: I see, but if the point is to not send the whole chat history and instead only publish messages as they are created, what would be the state that we want to get?

SamuelBellomo: sorry, when I mentioned buffered above, that just generally speaking. I wasn't suggesting we use it in this case here.

LPLafontaineB: oh my bad I misunderstood your point

Assets/BossRoom/Scripts/Server/Game/Entity/PublishMessageOnLifeChange.cs
```diff
@@ -0,0 +1,57 @@
...
+using Unity.Netcode;
+using UnityEngine;
+
+namespace Unity.Multiplayer.Samples.BossRoom.Server
+{
+    /// <summary>
+    /// Server-only component which publishes a message once the LifeState changes.
+    /// </summary>
+    [RequireComponent(typeof(NetworkLifeState))]
+    public class PublishMessageOnLifeChange : NetworkBehaviour
```
SamuelBellomo, line 11: We need a whole component just to publish a message?

Assets/BossRoom/Scripts/Shared/Game/GameEventMessage.cs
```diff
@@ -0,0 +1,26 @@
+namespace Unity.Multiplayer.Samples.BossRoom
+{
+    public struct LifeStateChangedEventMessage
+    {
+        public LifeState NewLifeState;
+        public FixedPlayerName CharacterName;
+        public CharacterTypeEnum CharacterType;
+
+        public static int Size => sizeof(LifeState) + FixedPlayerName.Size + sizeof(CharacterTypeEnum);
```
SamuelBellomo, line 9: you don't need to manually calculate this, you can just do sizeof(MyStruct). Since there's no polymorphism (that I can see?), everything in here should be known at compile time

Assets/BossRoom/Scripts/Server/Game/Entity/PublishMessageOnLifeChange.cs
LPLafontaineB: the goal here was to use composition to decide where to publish on, instead of making ServerCharacter even bigger and adding if-elses or switch statements to decide if we want to publish it or not (since we only want to publish it on player characters and the boss). I took inspiration from RaiseEventOnLifeChange, which we used for the boss.

Assets/BossRoom/Scripts/Shared/Game/GameEventMessage.cs
LPLafontaineB: sizeof requires an unsafe context for everything except for primitive types (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/sizeof), so from my understanding, that's why I cannot apply it directly on the struct and  get compile errors if I try.

SamuelBellomo: You can look at what FastBufferWriter does with its GetWriteSize method
        public static unsafe int GetWriteSize<T>() where T : unmanaged
This is what we use for our own internal messages (for example ServerRpcMessage)

SamuelBellomo: Also note you can enable unsafe code at the assembly level (doesn't have to be for the whole project). It's a checkbox in the assembly's options

LPLafontaineB: as discussed, a cleaner solution was integrated in #605

SamuelBellomo (dismissed):

jilfranco-unity (commented):
I was looking at these on my end and the animations look great in game! But the font is very large, so much so that the message gets cut off. I think the overall scale of the message box is too big as well. Is this happening on your end? If not, I think you have a 4k screen, right? :0 That could be causing some weirdness for the font scale

image
image

pdeschain (dismissed):

Assets/BossRoom/Scripts/Shared/ApplicationController.cs
```diff
@@ -47,11 +47,18 @@ private void Awake()
...
+            //these message channels are essential and persist for the lifetime of the lobby and relay services
             scope.BindMessageChannelInstance<UnityServiceErrorMessage>();
-
-            //this message channel is essential and persists for the lifetime of the lobby and relay services
             scope.BindMessageChannelInstance<ConnectStatus>();
+            scope.BindMessageChannelInstance<DoorStateChangedEventMessage>();
+
+            //these message channels are essential and persist for the lifetime of the lobby and relay services
+            //they are networked so that the clients can subscribe to those messages that are published by the server
+            scope.BindNetworkedMessageChannelInstance<LifeStateChangedEventMessage>();
```
SamuelBellomo, line 57: binding should be done at the state scope? let's investigate, this way no need for disconnect complexity?

SamuelBellomo (approved):

fernando-cortez (approved):

PR #602, opened 2022-04-08
fix: only server finds navigation object, asserts refactored [MTT-3248]

fernando-cortez:
### Description
This PR resolves an issue encountered when late-joining a lobby that was already mid-game. Granted, the scenario described was never reproduced, but it likely came from a scene init race condition on spawned characters (including imps/boss) and the navigation object inside the Boss Room scene. This moves that critical `GameObject.Find` to the NetworkObject's `OnNetworkSpawn()`, when both are guaranteed to have invoked `Awake()`. It also removes an unnecessary `GameObject.Find` calculation performed on clients.

### Issue Number(s)
[MTT-3248](https://jira.unity3d.com/browse/MTT-3248)

### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Scripts/Client/UI/UIStateDisplayHandler.cs
+++ b/Assets/BossRoom/Scripts/Client/UI/UIStateDisplayHandler.cs
@@ -79,11 +79,14 @@ namespace Unity.Multiplayer.Samples.BossRoom.Client
                 return;
             }
 
-            m_Camera = Camera.main;
+            m_Camera = GameObject.FindWithTag("MainCamera").GetComponent<Camera>();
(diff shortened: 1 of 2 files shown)


SamuelBellomo (approved):

Assets/BossRoom/Scripts/Client/UI/UIStateDisplayHandler.cs
```diff
@@ -79,19 +79,25 @@ public override void OnNetworkSpawn()
                 return;
             }
 
-            m_Camera = Camera.main;
```
LPLafontaineB, line 82: What is this change doing?

LPLafontaineB (approved):

PR #604, opened 2022-04-11
fix: relogin when auth token is stale [MTT-3253]

pdeschain:
### Description

Added a relogin function to the AuthenticationServiceFacade - it checks if we're no longer authenticated to do api calls and then re-signs in.

This is to prevent HTTP 401 when our token expires.

In order to prevent any potential issues when the UI was left hanging for a long time - we do a relogin attempt (which immediately bounces, if it's not needed) before any of the lobby calls from the UI. 

UI now also handles the failure to reconnect - it would unlock the UI's but would NOT do the lobby calls that require us to be authenticated. UI resets itself to try again, essentially. 

### Issue Number(s)

MTT-3253
### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Scripts/Client/UI/Lobby/LobbyUIMediator.cs
+++ b/Assets/BossRoom/Scripts/Client/UI/Lobby/LobbyUIMediator.cs
@@ -76,7 +76,14 @@ namespace Unity.Multiplayer.Samples.BossRoom.Visual
 
             BlockUIWhileLoadingIsInProgress();
 
-            await m_AuthenticationServiceFacade.ReloginIfNecessary();
+            bool playerIsAuthorized = await m_AuthenticationServiceFacade.EnsurePlayerIsAuthorized();
+
+            if (!playerIsAuthorized)
+            {
+                UnblockUIAfterLoadingIsComplete();
+                return;
+            }
+
             var lobbyCreationAttempt = await m_LobbyServiceFacade.TryCreateLobbyAsync(lobbyName, maxPlayers, isPrivate);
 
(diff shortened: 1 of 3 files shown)


Assets/BossRoom/Scripts/Client/UI/Lobby/LobbyUIMediator.cs
```diff
@@ -72,6 +76,7 @@ public async void CreateLobbyRequest(string lobbyName, bool isPrivate, int maxPl
 
             BlockUIWhileLoadingIsInProgress();
 
+            await m_AuthenticationServiceFacade.ReloginIfNecessary();
```
LPLafontaineB, line 79: should we do something here to handle when it fails to relogin? like returning to main menu? Or is it something that is already handled by the callers of these methods?

pdeschain: So what would happen currently - if such failure occurs, it'll be an unhandled exception here, and it will also show up as a popup message. You're thinking that we should not do the calls after this one, if it fails?

LPLafontaineB: If I'm understanding it correctly, if it fails, the other calls we will try to do in the same method would also fail, wouldn't they? If that's the case, then yeah maybe we shouldn't do them to avoid getting spammed with exceptions. Unless this is something that should not happen normally anyways and should be considered an invalid state from which we should not try to recover?

fernando-cortez (approved):

PR #607, opened 2022-04-12
feat: Added popup that tells users to set up project id if it's not set up [MTT-3265][MTT-3237]

pdeschain:
### Description

I've added the popup that tells users to set up their project id. Also adjusted the flow slightly - if there's no project id, we block away the lobby UI and prevent attempts to authenticate (they'd cause errors and nuisance popups)

### Issue Number(s)

MTT-3265

### Contribution checklist

### DIFF
--- a/Assets/BossRoom/Scripts/Client/Game/State/ClientMainMenuState.cs
+++ b/Assets/BossRoom/Scripts/Client/Game/State/ClientMainMenuState.cs
@@ -7,7 +7,7 @@ using Unity.Multiplayer.Samples.BossRoom.Visual;
 using Unity.Services.Authentication;
 using Unity.Services.Core;
 using UnityEngine;
-
+using UnityEngine.UI;
 
 namespace Unity.Multiplayer.Samples.BossRoom.Client
 {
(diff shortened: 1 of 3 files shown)


LPLafontaineB (commented):
If we try to host a game through direct IP without the project id setup, it now gives us an exception and prevents us from doing so. Is that intentional or a bug? Should playing through direct IP still be allowed?

pdeschain:
SessionManager actually relies on having access to Player ID through authentication service. Seems like the best course of action is to also block the IP connection option.

LPLafontaineB (approved):

fernando-cortez (approved):
This resolves [MTT-3237](https://jira.unity3d.com/browse/MTT-3237), so I'll assign that one to you @pdeschain.

Assets/BossRoom/Scripts/Client/Game/State/ClientMainMenuState.cs
```diff
@@ -46,6 +45,15 @@ protected override void InitializeScope()
         [Inject]
         async void InjectDependenciesAndInitialize(AuthenticationServiceFacade authServiceFacade, LocalLobbyUser localUser, LocalLobby localLobby)
         {
+            #if UNITY_EDITOR
+            if (string.IsNullOrEmpty(UnityEditor.CloudProjectSettings.projectId))
```
SamuelBellomo, line 49: There's no way to check with Application.cloudProjectId? And this way it would work in builds too?

PR #613, opened 2022-04-13
feat: connection feedback + IP connection window [MTT-2315] [MTT-3234]

fernando-cortez:
### Description
This PR provides feedback on connection errors on IP flow. 

* A popup window is shown when unable to host.
* A connection window is shown when connecting to a Host IP, with the function to cancel. If on UTP, the connection attempts are displayed to the user.

Connect attempts have been reduced on UTP (attempt every second for 60s to just 15s).

### Issue Number(s)
[MTT-2315](https://jira.unity3d.com/browse/MTT-2315), [MTT-3234](https://jira.unity3d.com/browse/MTT-3234).

### Contribution checklist


Assets/BossRoom/Scripts/Client/UI/IPUIMediator.cs
```diff
@@ -87,9 +75,14 @@ public void HostIPRequest(string ip, string port)
...
-            m_GameNetPortal.StartHost(ip, portNum);
-
-            m_SignInSpinner.SetActive(true);
+            if (m_GameNetPortal.StartHost(ip, portNum))
+            {
+                m_SignInSpinner.SetActive(true);
+            }
+            else
+            {
+                PopupManager.ShowPopupPanel("Connection Failed", "Server failed to bind.");
```
LPLafontaineB, line 84: can a different issue cause StartHost to return false?

LPLafontaineB, line 84: can a different issue cause StartHost to return false?

Assets/BossRoom/Scripts/Client/UI/IPUIMediator.cs
```diff
@@ -104,9 +97,30 @@ public void JoinWithIP(string ip, string port)
 
             m_GameNetPortal.PlayerName = m_PlayerNameLabel.text;
 
+            m_SignInSpinner.SetActive(true);
+
             m_ClientNetPortal.StartClient(ip, portNum);
```
LPLafontaineB, line 102: Shouldn't we do something similar here and check the return value of StartClient?

Assets/BossRoom/Scripts/Client/UI/IPUIMediator.cs
LPLafontaineB: Also, while testing I noticed that if we get here after failing to bind, further StartHost and StartClient calls all fail. It looks like a bug for the sdk though

Assets/BossRoom/Scripts/Client/UI/IPConnectionWindow.cs
```diff
@@ -0,0 +1,109 @@
...
+
+namespace Unity.Multiplayer.Samples.BossRoom.Visual
+{
+    public class IPConnectionWindow : MonoBehaviour
+    {
+        [SerializeField]
+        CanvasGroup m_CanvasGroup;
+
+        [SerializeField]
+        IPUIMediator m_IPUIMediator;
```
pdeschain, line 17: Shouldn't a reference to this object be passed in via DI?

Assets/BossRoom/Scripts/Client/UI/IPConnectionWindow.cs
```diff
@@ -0,0 +1,109 @@
...
+                    StartCoroutine(DisplayUTPReconnectAttempts(maxConnectAttempts, connectTimeoutMS, OnTimeElapsed));
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(chosenTransport));
+            }
+
+            Show();
+        }
+
+        public void Cancel()
```
pdeschain, line 69: Is this function necessary if only wraps `End();`?

Assets/BossRoom/Scripts/Client/UI/IPConnectionWindow.cs
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        IEnumerator WaitUntilUNETDisconnected(Action endAction)
+        {
+            yield return new WaitUntil(() => !NetworkManager.Singleton.IsListening);
+
+            endAction();
+        }
+
+        IEnumerator DisplayUTPReconnectAttempts(int maxReconnectAttempts, int connectTimeoutMS, Action endAction)
```
SamuelBellomo, line 87: As discussed, let's just use time * nbAttempts for this countdown. We want to be SDK authoritative.

Assets/BossRoom/Scripts/Client/UI/IPConnectionWindow.cs
SamuelBellomo: SerializeField is a form of DI by itself, just controlled by the editor

Assets/BossRoom/Scripts/Client/UI/IPUIMediator.cs
SamuelBellomo, line 84: Should this go into the message channel flow for failed connections?

SamuelBellomo: How do we bubble issue details SDK side? error logs?
Also bug that bug!

Assets/BossRoom/Scripts/Client/UI/IPConnectionWindow.cs
pdeschain: The reference would always be fulfilled and could never fail, unless there is no IPUIMediator bound (which in itself is a programmer error).

pdeschain: Since we are using DI for similar purposes elsewhere - I think that it would also be idiomatic and conformant to the general approach we are taking.

fernando-cortez: After discussion: will pass in via DI.

Assets/BossRoom/Scripts/Client/UI/IPUIMediator.cs
fernando-cortez: Regardless of the reason, the error returned by UTP is the same, so I think it makes sense to log this generic message.

Assets/BossRoom/Scripts/Client/UI/IPUIMediator.cs
fernando-cortez: If this returns false, UTP will retry in the background, following the reconnect attempts / connect timeout settings set on the UTP component. So, it's not enough to catch it just here.

Assets/BossRoom/Scripts/Client/UI/IPConnectionWindow.cs
fernando-cortez: Cleaned this up 👍

Assets/BossRoom/Scripts/Client/UI/IPConnectionWindow.cs
```diff
@@ -84,15 +87,15 @@ IEnumerator WaitUntilUNETDisconnected(Action endAction)
...
 
-            while (attempt < maxReconnectAttempts)
+            while (connectionDuration > 0f)
             {
-                attempt++;
-                m_TitleText.text = $"Connecting...{attempt}/{maxReconnectAttempts}.";
-                yield return new WaitForSeconds(connectTimeoutMS / 1000f);
+                m_TitleText.text = $"Connecting...\n{Mathf.CeilToInt(connectionDuration)}";
+                connectionDuration -= Time.deltaTime;
+                yield return null;
```
SamuelBellomo, line 98: waitforseconds(1) as discussed

SamuelBellomo (dismissed):

SamuelBellomo (dismissed):

LPLafontaineB (approved):

SamuelBellomo (approved):

PR #635, opened 2022-04-29
fix: reduced scene load timeout

LPLafontaineB:
### Description

This PR simply reduces the scene load timeout from 120s to 20s. Since scene load times are short in BossRoom, we don't need to have such a long timeout.
### Contribution checklist


Assets/BossRoom/Prefabs/NetworkingManager.prefab
```diff
@@ -230,8 +238,8 @@ MonoBehaviour:
     RecycleNetworkIds: 1
     NetworkIdRecycleDelay: 120
     RpcHashSize: 0
-    LoadSceneTimeOut: 120
-    MessageBufferTimeout: 20
+    LoadSceneTimeOut: 20
```
SamuelBellomo, line 241:
```suggestion
    LoadSceneTimeOut: 10
```
did you try shorter? was it too short?

LPLafontaineB: I did not, but I don't think it should be as low as the disconnect timeout. If a player briefly loses connection to the server while loading a scene but recuperates before being disconnected by the timeout, they could take longer to load the scene, but I don't think we should consider this a scene load timeout if they recover.
i.e: 
1. server sends load scene event
2. 1/2 rtt later, client receives it (say with an rtt of 500 to make it a bit extreme)
3. client loses connection before finishing to load it
4. 8 seconds later, client recovers. Thus, not disconnected by timeout
5. a bit more than one second later, client finishes loading the scene. Sends load scene complete to server
6. server receives load scene comple 1/2 rtt later, which makes it slightly over 10s. Client is considered having timed out for the scene load

SamuelBellomo: what happens on load scene timeout? Is the client disconnected?

LPLafontaineB: The SceneManager creates a LoadEventCompleted event, which it handles locally and sends to all clients. The message contains a field with the list of clientids that timed out. BossRoom doesn't use that information, so concretely what would happen on our side is that the loading screen would end for all clients except the one for which the scene load timed out. The message would still be sent to them, and since they are reliable messages, they would eventually get it and end their loading screen. There is a possibility, though, that they would receive it before they finished, and would end their loading screen early.

SamuelBellomo: what happens to that client? What happens if the host changes scene again? Will we have a client that's always desynced?

LPLafontaineB: we don't do anything with the timeout, so the client should simply finish loading/unloading the scene after having received the LoadEventCompleted message. I'll do some testing to be sure.
But even though it should not create issues on our side, conceptually I think it makes more sense to have the load scene timeout higher than the disconnect timeout to prevent this

SamuelBellomo: @NoelStephensUnity what do you think?

NoelStephensUnity: I think this:
https://github.com/Unity-Technologies/com.unity.netcode.gameobjects/pull/1886
:)

NoelStephensUnity: But honestly that wouldn't solve for the issue with a client that actually is disconnected purposely during a synchronization.
There is still an open jira ticket to resolve this issue, and I think I can easily add some additional checks while it is waiting for players to finish the scene event (loading/unloading/synchronizing) where it determines if any clients are no longer connected and if so exclude them from the check.  
That should at least reduce the wait time down to how long it takes the transport to detect the connection being lost and perhaps 1 frame longer...and then as long as the rest of the clients are synchronized the server should finish the scene event.

SamuelBellomo:

fernando-cortez:
Also just commenting that this is targeting main at the moment. Should be develop.

LPLafontaineB:
> Also just commenting that this is targeting main at the moment. Should be develop.

This was initially intended as a hotfix right after release to reduce the effects of [MTT-3396](https://jira.unity3d.com/browse/MTT-3396) and [MTT-3494](https://jira.unity3d.com/browse/MTT-3494).
However, since [a PR fixing the root cause](https://github.com/Unity-Technologies/com.unity.netcode.gameobjects/pull/1973) has been created on the sdk, we could wait until the next release. What do you think @SamuelBellomo ?

SamuelBellomo:
> > Also just commenting that this is targeting main at the moment. Should be develop.
> 
> This was initially intended as a hotfix right after release to reduce the effects of [MTT-3396](https://jira.unity3d.com/browse/MTT-3396) and [MTT-3494](https://jira.unity3d.com/browse/MTT-3494). However, since [a PR fixing the root cause](https://github.com/Unity-Technologies/com.unity.netcode.gameobjects/pull/1973) has been created on the sdk, we could wait until the next release. What do you think @SamuelBellomo ?

Let's wait @LPLafontaineB

pdeschain (approved):

SamuelBellomo (approved):

PR #660, opened 2022-05-26
feat: replacing discope with vcontainer [MTT-3675]

pdeschain:
### Description

Moving the project from our very basic DI implementation - DIScope to a fully-featured and performant VContainer https://vcontainer.hadashikick.jp

Pulled in some bits of Sam's work that went into adding dedicated server support, specifically the bits that allowed GameStateBehaviour to stop being a NetworkedBehaviour. This further allowed me to inherit from a VContainer primitive for "assembly points" in the application - the LifetimeContainer . 

Every scene (but the Startup and MainMenu) has two states - the client and the server states. All of these objects, including ApplicationController are now inheriting from a LifetimeContainer, which allows us to specify a list of auto-injected objects and allows us to specify the parent container for any scene-related State LifetimeContainer. This is a nice feature that VContainer has, that allows us to visually set things up in Inspector.
We can also simplify our code-base a bit by using the more expressive [Inject] attribute that VContainer provides - it can be placed on other things, but the function calls, hence we can have property injection and other benefits of a more solid DI framework.

[VContainer License (MIT)](https://github.com/hadashiA/VContainer/blob/master/LICENSE)

### Issue Number(s)

MTT-2365

### Contribution checklist


Assets/BossRoom/Scripts/Game/Client/UI/ConnectionStatusMessageUIManager.cs
```diff
@@ -1,6 +1,7 @@
 using System;
 using Unity.Multiplayer.Samples.BossRoom.Shared.Infrastructure;
 using UnityEngine;
+using VContainer;
```
SamuelBellomo, line 4: needed?

Assets/BossRoom/Scripts/Game/Client/UI/Lobby/LobbyUIMediator.cs
```diff
@@ -7,6 +7,7 @@
 using Unity.Services.Core;
 using Unity.Services.Lobbies.Models;
 using UnityEngine;
+using VContainer;
```
SamuelBellomo, line 10: same?

Assets/BossRoom/Scripts/Game/Client/UI/RoomNameBox.cs
```diff
@@ -4,6 +4,7 @@
 using Unity.Multiplayer.Samples.BossRoom.Shared.Infrastructure;
 using Unity.Multiplayer.Samples.BossRoom.Shared.Net.UnityServices.Lobbies;
 using UnityEngine.UI;
+using VContainer;
```
SamuelBellomo, line 7: I'll stop creating comments for all of these, I'd make sure to review the PR for all of these

SamuelBellomo: @fernando-cortez I'm actually surprised CI didn't catch this?

fernando-cortez: I've noticed that too. I believe the formatting doesn't actually check for unnecessary directives, though I think it can. I will look into that.

Assets/BossRoom/Scripts/Game/Client/UI/ConnectionStatusMessageUIManager.cs
pdeschain: Yup, to provide [Inject] attribute, now that we're using VContainer

Assets/BossRoom/Scripts/Game/Client/UI/RoomNameBox.cs
pdeschain: This is not an unnecessary directive.

Assets/BossRoom/Scripts/Game/Client/UI/ConnectionStatusMessageUIManager.cs
SamuelBellomo: But I don't see any [Inject] in this file though...
Is this a codegen thing? Could you add comments to these usings so users understand this please?

LPLafontaineB: There is one for the InjectDependencies() method below, no?

SamuelBellomo: oooh now I understand, since it's using the same "Inject" name, there's no new changes to this file.
Now the question is, why aren't we doing like the others and injecting directly in properties?

pdeschain: Because there's logic that needs to be run after we get the dependencies injected - specifically, there are some subscriptions happening. True, one can do the same by using VContainer interfaces for the IStartable or whatnot, but... why? Same result, tom[ay]to==tom[ah]to

pdeschain:
Merged develop, need to verify that all the important changes in scenes aren't overriden and restore them if needed.

pdeschain:
Recreated the PR here: https://github.com/Unity-Technologies/com.unity.multiplayer.samples.coop/pull/679

PR #666, opened 2022-05-30
refactor: replacing GameNetPortal with state machine [MTT-1742] [MTT-3501] [MTT-3502]

LPLafontaineB:
### Description

This PR aims to simplify and clarify the connection flow in Boss Room. It replaces GameNetPortal (as well as ClientGameNetPortal and ServerGameNetPortal) with a state machine (ConnectionManager). This state machine receives inputs from other parts of our code, and registers to callbacks from NGO. It then redirects these inputs and callbacks to the current state, which then handles it accordingly.

ConnectionManager is responsible for initializing the different states and redirecting inputs and callbacks to them. It also  handles our custom messaging for giving a disconnect reason to clients.
The connection logic itself is the responsibility of the different states.

### Issue Number(s)

[MTT-1742](https://jira.unity3d.com/browse/MTT-1742), [MTT-3502](https://jira.unity3d.com/browse/MTT-3502) & [MTT-3501](https://jira.unity3d.com/browse/MTT-3501)

### Contribution checklist


Assets/Scripts/Gameplay/ConnectionManagement/ConnectionManager.cs
```diff
@@ -0,0 +1,210 @@
...
+            NetworkManager.OnClientConnectedCallback += OnClientConnectedCallback;
+            NetworkManager.OnClientDisconnectCallback += OnClientDisconnectCallback;
+            NetworkManager.ConnectionApprovalCallback += ApprovalCheck;
+        }
+
+        void OnDestroy()
+        {
+            NetworkManager.OnClientConnectedCallback -= OnClientConnectedCallback;
+            NetworkManager.OnClientDisconnectCallback -= OnClientDisconnectCallback;
+
```
fernando-cortez, line 109: Deregister from ConnectionApprovalCallback here?

fernando-cortez (dismissed):
I really like this! A suggestion would be to include a little blurb (either in each state) or somewhere central where you could detail what state each state can transition to. Even without that, it was real simple to navigate. 🚀

pdeschain:
Overall I like this implementation much-much more than what we had before - it cleanly wraps the whole connection logic concept. However there's one aspect of the implementation itself that I have questions about, specifically - should we have a 
```
 public enum ConnectionStateType
    {
        Offline,
        Connecting,
        Connected,
        Reconnecting,
        Hosting
    }
```
that maps 1-1 to the classes that provide the logic for these?

I'm thinking that it's a direct use-case for the "smart enum" pattern in C# - https://docs.microsoft.com/en-us/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/enumeration-classes-over-enum-types

and an open-source implementation of this pattern, if you like it and want to borrow ideas:
https://github.com/ardalis/SmartEnum

Basically, we use static instances of a class instead of an enum. Then we can define the logic that is contained within those implementations to be housed in the smart enums.

LPLafontaineB:
> Overall I like this implementation much-much more than what we had before - it cleanly wraps the whole connection logic concept. However there's one aspect of the implementation itself that I have questions about, specifically - should we have a
> 
> ```
>  public enum ConnectionStateType
>     {
>         Offline,
>         Connecting,
>         Connected,
>         Reconnecting,
>         Hosting
>     }
> ```
> 
> that maps 1-1 to the classes that provide the logic for these?
> 
> I'm thinking that it's a direct use-case for the "smart enum" pattern in C# - https://docs.microsoft.com/en-us/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/enumeration-classes-over-enum-types
> 
> and an open-source implementation of this pattern, if you like it and want to borrow ideas: https://github.com/ardalis/SmartEnum
> 
> Basically, we use static instances of a class instead of an enum. Then we can define the logic that is contained within those implementations to be housed in the smart enums.

I like that idea, it would be much cleaner that way! I don't think that for this specific case we would need something as complex as the SmartEnum example you shared, so I simply went for adding the different states as static members of ConnectionState. However, if we want to use something similar for the state machine we use for the enemy AI, I could implement something more generic and reusable

pdeschain (approved):
👍

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ConnectionState.cs
```diff
@@ -0,0 +1,80 @@
...
+using Unity.Multiplayer.Samples.BossRoom.Shared.Infrastructure;
+using Unity.Netcode;
+using UnityEngine;
+
+namespace Unity.Multiplayer.Samples.BossRoom
+{
+    /// <summary>
+    /// Base class representing a connection state. Contains a static member for each possible state.
+    /// </summary>
+    public abstract class ConnectionState
```
SamuelBellomo, line 13: "BaseConnectionState"?

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ConnectionState.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        public static readonly OfflineConnectionState Offline = new OfflineConnectionState();
```
SamuelBellomo, line 15: This is interesting and I like this. I'd still like to have a quick convo on the other option we discussed, just to be thorough about this design decision.

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ConnectingConnectionState.cs
```diff
@@ -0,0 +1,47 @@
...
+using Unity.Multiplayer.Samples.BossRoom.Shared.Infrastructure;
+using Unity.Multiplayer.Samples.Utilities;
+
+namespace Unity.Multiplayer.Samples.BossRoom
+{
+    /// <summary>
+    /// Connection state corresponding to when a client is attempting to connect to a server. If successful,transitions
+    /// to the Connected state. If not, transitions to the Offline state.
+    /// </summary>
+    public class ConnectingConnectionState : ConnectionState
```
SamuelBellomo, line 12: "ConnectingState"

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ConnectedConnectionState.cs
```diff
@@ -0,0 +1,58 @@
...
+using Unity.Multiplayer.Samples.BossRoom.ApplicationLifecycle.Messages;
+using Unity.Multiplayer.Samples.BossRoom.Shared.Infrastructure;
+
+namespace Unity.Multiplayer.Samples.BossRoom
+{
+    /// <summary>
+    /// Connection state corresponding to a connected client. When being disconnected or timed out, transitions to the
+    /// Offline or Reconnecting state, depending on the cause of the disconnect.
+    /// </summary>
+    public class ConnectedConnectionState : ConnectionState
```
SamuelBellomo, line 11: "ClientConnectedState"? (I'm assuming this is only for clients?

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ConnectedConnectionState.cs
```diff
@@ -0,0 +1,58 @@
...
+        [Inject]
+        void InjectDependencies(IPublisher<QuitGameSessionMessage> quitGameSessionPublisher, ISubscriber<ConnectStatus> connectStatusSubscriber)
+        {
+            m_QuitGameSessionPublisher = quitGameSessionPublisher;
+            m_ConnectStatusSubscriber = connectStatusSubscriber;
+        }
+
+        public override void Enter()
+        {
+            m_Subscription = m_ConnectStatusSubscriber.Subscribe(status => m_ConnectStatus = status);
```
SamuelBellomo, line 28: why is this handled in this state? Shouldn't this be provided in OnClientDisconnect as a parameter?

LPLafontaineB: My intent here was to avoid giving the ConnectionManager this responsibility to hold this data.
I had an alternative solution in mind that could avoid both of those, but I'm not entirely convinced by it. I think it might be easier to discuss it in person, though

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ConnectingConnectionState.cs
LPLafontaineB: The intent here was to keep the base type as a suffix, but if it makes less sense I will remove it

SamuelBellomo: It's really about readability. Right now, I'm reading double connecting, it's weird.

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ConnectionState.cs
SamuelBellomo: @LPLafontaineB I don't see this change?

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ClientReconnectingState.cs
```diff
@@ -0,0 +1,157 @@
...
+        {
+            if (m_ReconnectCoroutine != null)
+            {
+                m_ConnectionManager.StopCoroutine(m_ReconnectCoroutine);
+                m_ReconnectCoroutine = null;
+            }
+            m_ReconnectMessagePublisher.Publish(new ReconnectMessage(k_NbReconnectAttempts, k_NbReconnectAttempts));
+        }
+
+        public override void OnClientConnected(ulong clientId)
```
SamuelBellomo, line 52: I clientID as a parameter here, I'd assume this gets triggered when other clients connect as well. Did you test a case with more than a one client?

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionManager.cs
```diff
@@ -0,0 +1,171 @@
...
+        {
+            NetworkManager.OnClientConnectedCallback -= OnClientConnectedCallback;
+            NetworkManager.OnClientDisconnectCallback -= OnClientDisconnectCallback;
+            NetworkManager.ConnectionApprovalCallback -= ApprovalCheck;
+
+        }
+
+        public void ChangeState(ConnectionState newState)
+        {
+            Debug.Log($"Changed connection sate from {m_CurrentState.GetType().Name} to {newState.GetType().Name}.");
```
SamuelBellomo, line 87:
```suggestion
            Debug.Log($"Changed connection state from {m_CurrentState.GetType().Name} to {newState.GetType().Name}.");
```

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionManager.cs
```diff
@@ -0,0 +1,171 @@
...
+            writer.WriteValueSafe(status);
+            NetworkManager.Singleton.CustomMessagingManager.SendNamedMessageToAll(nameof(ReceiveServerToClientSetDisconnectReason_CustomMessage), writer);
+        }
+
+        /// <summary>
+        /// Sends a DisconnectReason to the indicated client. This should only be done on the server, prior to disconnecting the client.
+        /// </summary>
+        /// <param name="clientID"> id of the client to send to </param>
+        /// <param name="status"> The reason for the upcoming disconnect.</param>
+        public static void SendServerToClientSetDisconnectReason(ulong clientID, ConnectStatus status)
```
SamuelBellomo, line 164: This is the same as above, just with a targeted client right? Why not merge them and have an optional parameter?

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ClientConnectedState.cs
```diff
@@ -0,0 +1,43 @@
+using System;
```
SamuelBellomo, line 1: I don't see any meta files for this connection states, did you forget to commit them?

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ClientConnectingState.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Unity.Multiplayer.Samples.BossRoom
+{
+    /// <summary>
+    /// Connection state corresponding to when a client is attempting to connect to a server. If successful,transitions
+    /// to the Connected state. If not, transitions to the Offline state.
+    /// </summary>
+    public class ClientConnectingState : ConnectionState
+    {
+        IPublisher<ConnectStatus> m_ConnectStatusPublisher;
```
SamuelBellomo, line 13: I see this in a good majority of your states, why not put it in the base class? It'd make sense to have every state be able to publish connection status.

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ClientReconnectingState.cs
```diff
@@ -0,0 +1,157 @@
...
+using UnityEngine;
+
+namespace Unity.Multiplayer.Samples.BossRoom
+{
+    /// <summary>
+    /// Connection state corresponding to a client attempting to reconnect to a server. It will try to reconnect a
+    /// number of times defined by k_NbReconnectAttempts. If it succeeds, it will transition to the Connected state. If
+    /// not, it will transition to the Offline state.
+    /// </summary>
+    public class ClientReconnectingState : OfflineState
```
SamuelBellomo, line 18: (I'm not saying we should do this) any reason why connecting is an offline state but reconnecting isn't?

Assets/Scripts/Gameplay/GameplayObjects/Character/NetworkAvatarGuidState.cs
```diff
@@ -34,6 +34,14 @@ public Avatar RegisteredAvatar
             }
         }
 
+        public void SetRandomAvatar()
+        {
+            if (IsServer)
+            {
+                AvatarGuid.Value = m_AvatarRegistry.GetRandomAvatar().Guid.ToNetworkGuid();
+            }
```
SamuelBellomo, line 42: should this throw if called from client? let's not have silent errors

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/OfflineState.cs
```diff
@@ -0,0 +1,197 @@
...
+            catch (Exception e)
+            {
+                Debug.LogErrorFormat($"{e.Message}");
+                throw;
+            }
+
+            StartHost(GetPlayerId(), playerName);
+        }
+
+        public override void ApprovalCheck(byte[] connectionData, ulong clientId, NetworkManager.ConnectionApprovedDelegate connectionApprovedCallback)
```
SamuelBellomo, line 127: oooh that's interesting that now that we have this pattern, our approval is separated in two calls... mmm it seems a bit weird though that approval happens while you're offline, couldn't this be in a "starting host" state?

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/OfflineState.cs
```diff
@@ -0,0 +1,197 @@
...
+        {
+            SetConnectionPayload(playerId, playerName);
+            var success = m_ConnectionManager.NetworkManager.StartHost();
+            if (!success)
+            {
+                m_ConnectStatusPublisher.Publish(ConnectStatus.StartHostFailed);
+            }
+            else
+            {
+                m_ConnectionManager.ChangeState(Hosting);
```
SamuelBellomo, line 181: I wouldn't hide state changing in a "StartHost". This seems like hidden behaviour. When reading "StartHost" from the callsite, I'd expect it to do some NGO things, nothing else.

LPLafontaineB: Good point! I moved it here to reduce code duplication but I agree that it makes more sense to not hide it

Assets/Scripts/Gameplay/GameplayObjects/Character/NetworkAvatarGuidState.cs
LPLafontaineB: If I remove the IsServer check, we would get the typical NGO exception about clients writing to NetworkVariables. Would that be better?

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ClientReconnectingState.cs
LPLafontaineB: Reconnecting inherits from OfflineState because it handles some connection methods that are present in the OfflineState.

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ClientReconnectingState.cs
LPLafontaineB: The docs mention that "this callback is only ran on the server and on the local client that connects." But even if it was also ran on other clients, if we are here, we are not connected to the server anyways, so this couldn't be triggered when other clients connect to the server

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ClientConnectingState.cs
LPLafontaineB: With our current DI implementation, I would have to inject this in every daughter class anyways, but once VContainer is merged I'll update this

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ConnectionState.cs
```diff
@@ -0,0 +1,83 @@
...
+        public virtual Task StartHostLobbyAsync(string playerName)
+        {
+            return Task.CompletedTask;
+        }
+
+        public virtual void OnUserRequestedShutdown() { }
+
+        public virtual void OnDisconnectReasonReceived(ConnectStatus disconnectReason) { }
+
+        /// <summary>
```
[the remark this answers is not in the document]
LPLafontaineB: I could move the "remarks" section to the implementations, yes, but I'd keep the rest here at least, since there could be multiple implementations. Anyways, this will be changed if pre.10 is released before this PR is merged

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ClientConnectedState.cs
SamuelBellomo: nvm, I think I had them filtered out lol, resolving
silly sam

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionManager.cs
SamuelBellomo: with a "broadcast" parameter for example

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/OfflineState.cs
LPLafontaineB: Done, it's much cleaner now I think

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/OfflineState.cs
LPLafontaineB: The addition of the StartingHost state fixes this as well

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ClientConnectedState.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public override void Exit() { }
+
+        public override void OnClientDisconnect(ulong clientId)
+        {
+            m_ConnectStatusPublisher.Publish(ConnectStatus.Reconnecting);
+            m_ConnectionManager.ChangeState(ClientReconnecting);
+        }
+
+        public override void OnUserRequestedShutdown()
```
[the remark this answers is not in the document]
LPLafontaineB: As for the publishing of UserRequestedDisconnect status, it could be moved to the base class, or handled directly by the ConnectionManager itself. Would that make sense?

Assets/Scripts/Gameplay/GameplayObjects/Character/NetworkAvatarGuidState.cs
SamuelBellomo: if we don't need to handle this ourselves, indeed

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionManager.cs
```diff
@@ -0,0 +1,171 @@
+using System;
```
SamuelBellomo, line 1: Why is ConnectionManager in Gameplay?

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionManager.cs
```diff
@@ -0,0 +1,171 @@
...
+using System.Threading.Tasks;
```
SamuelBellomo, line 2: We should have tests for this. TestHelpers should help here.
cc @fernando-cortez

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionManager.cs
SamuelBellomo: cc @pdeschain

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ConnectionState.cs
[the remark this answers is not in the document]
SamuelBellomo: yeah, I changed the comment in my PR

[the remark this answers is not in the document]
SamuelBellomo: I mean, the logic it's describing is in the implementation, let's have them close to each other. A comment describing a method signature isn't super helpful...

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ClientReconnectingState.cs
SamuelBellomo: yep, that's not a valid reason for having it inherit. Think of inheritance as "class A IS a classB". If we need common utils or things like that, let's extract those in a common util. It needs to make sense to say that class A IS a classB, else you end up with a class that contains behaviour it shouldn't.
To me, from a design perspective, it'd make sense to have both connecting and reconnecting "IS" an offline state (as they are not online yet). They could also be their own thing. But we should be purposeful about it not do it "just because it has some common methods".

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ClientConnectedState.cs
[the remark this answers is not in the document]
SamuelBellomo: my thinking is mostly about making sure everyone has the same "procedure" for shutting things down. This could be in a static method in a utils as well.

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionManager.cs
LPLafontaineB: The goal of this PR was to first refactor it. I intended to move it to a separate assembly in a subsequent PR, but I could do it as part of this one too.

Assets/Scripts/ConnectionManagement/ConnectionState/ClientConnectedState.cs
```diff
@@ -24,20 +24,19 @@ void InjectDependencies(IPublisher<ConnectStatus> connectStatusPublisher)
         public override void OnClientDisconnect(ulong _)
         {
             m_ConnectStatusPublisher.Publish(ConnectStatus.Reconnecting);
-            m_ConnectionManager.ChangeState(ClientReconnecting);
+            StateChangeRequest?.Invoke(ClientReconnecting);
```
SamuelBellomo, line 27: why conditional operator? I'd assume there IS a callback here. If we assume it's there and it's null, I'd fail loudly (and let it throw an exception)

Assets/Scripts/ConnectionManagement/ConnectionManager.cs
```diff
@@ -49,6 +55,21 @@ public class ConnectionPayload
...
     {
         ConnectionState m_CurrentState;
+        ConnectionState CurrentState
+        {
+            get => m_CurrentState;
+            set
+            {
+                if (m_CurrentState != null)
+                {
+                    m_CurrentState.StateChangeRequest -= OnStateChangeRequest;
```
SamuelBellomo, line 65: not sure I understand why this is happening, let's have some comments.
Why is this now a callback?

Assets/Scripts/ConnectionManagement/ConnectionState/StartingHostState.cs
```diff
@@ -0,0 +1,110 @@
...
+            m_ConnectStatusPublisher.Publish(ConnectStatus.Success);
+            StateChangeRequest?.Invoke(Hosting);
+        }
+
+        public override void ApprovalCheck(NetworkManager.ConnectionApprovalRequest request, NetworkManager.ConnectionApprovalResponse response)
+        {
+            var connectionData = request.Payload;
+            var clientId = request.ClientNetworkId;
+            // This happens when starting as a host, before the end of the StartHost call. In that case, we simply approve ourselves.
+            if (m_ConnectionManager.NetworkManager.IsHost && clientId == m_ConnectionManager.NetworkManager.LocalClientId)
```
SamuelBellomo, line 61: Is this check needed, since we're in the "StartingHostState"? Let's make sure we're cleaning up old logic, this current architecture could simplify some of these checks.

Assets/Scripts/ConnectionManagement/ConnectionManager.cs
```diff
@@ -58,6 +79,8 @@ public class ConnectionManager : MonoBehaviour
         NetworkObject m_GameState;
         public NetworkObject GameState => m_GameState;
 
+        public const int k_MaxLobbyPlayers = 8;
```
SamuelBellomo, line 82: as discussed
"max connected players"
and let's have this a serialize field.

Assets/Scripts/Gameplay/UI/IPUIMediator.cs
```diff
@@ -61,6 +65,16 @@ void Start()
...
         }
 
+        void OnDestroy()
+        {
+            m_Subscription.Dispose();
+        }
+
+        void OnConnectStatusMessage(ConnectStatus connectStatus)
+        {
+            DisableSignInSpinner();
```
SamuelBellomo, line 75: for all connect statuses?

Assets/Scripts/ConnectionManagement/ConnectionManager.cs
LPLafontaineB: The goal was to make it more robust and prevent classes that should not be modifying the state from doing it. Now only the current state it is subscribed to can request a state change.

Assets/Scripts/ConnectionManagement/ConnectionState/StartingHostState.cs
SamuelBellomo: I see a thumbs up but no changes related to this?

Assets/Scripts/ConnectionManagement/ConnectionManager.cs
SamuelBellomo: This seems like a good use for restricting access with "internal" and not add complexity to our API. (let your compiler do the work for you).

SamuelBellomo (dismissed):

Assets/Scripts/Gameplay/ConnectionManagement/ConnectionState/ClientConnectingState.cs
SamuelBellomo: So not sure I understand, if this was in the base class, you'd still need to inject sure, but you still wouldn't need to add this publisher to all your state, that's my point. It'd make sense to have all states have a publisher. The Inject attribute is inherited, so this should work in vcontainer.

Assets/Scripts/ConnectionManagement/Unity.BossRoom.ConnectionManagement.asmdef
```diff
@@ -10,7 +10,8 @@
         "GUID:5540e30183c82e84b954c033c388e06c",
         "GUID:59b3bd604d0445f583783872c3e648fc",
         "GUID:4c3f49d89436d478ea78315c03159dcc",
-        "GUID:f2d49d9fa7e7eb3418e39723a7d3b92f"
+        "GUID:f2d49d9fa7e7eb3418e39723a7d3b92f",
```
SamuelBellomo, line 13: Phil changed this for his other assemblies, can we use name based references instead of guids please? guids make it hard to review (and now that we're mostly done with our assembly refactor, we shouldn't need guids anymore)

SamuelBellomo (approved):

Assets/Scripts/ConnectionManagement/ConnectionManager.cs
```diff
@@ -0,0 +1,205 @@
...
+    [Serializable]
+    public class ConnectionPayload
+    {
+        public string playerId;
+        public string playerName;
+        public bool isDebug;
+    }
+
+    /// <summary>
+    /// This state machine handles connection through the NetworkManager. It is responsible for listening to
```
NoelStephensUnity, line 52: 😻 
Love the state machine approach!

Assets/Scripts/ConnectionManagement/ConnectionState/HostingState.cs
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        public override void Exit()
+        {
+            SessionManager<SessionPlayerData>.Instance.OnServerEnded();
+        }
+
+        public override void OnClientConnected(ulong clientId)
+        {
+            m_ConnectionEventPublisher.Publish(new ConnectionEventMessage() { ConnectStatus = ConnectStatus.Success, PlayerName = SessionManager<SessionPlayerData>.Instance.GetPlayerData(clientId)?.PlayerName });
```
NoelStephensUnity, line 46: Like this approach as well.

Assets/Scripts/ConnectionManagement/ConnectionState/HostingState.cs
```diff
@@ -0,0 +1,166 @@
...
+        /// <param name="request"> The initial request contains, among other things, binary data passed into StartClient. In our case, this is the client's GUID,
+        /// which is a unique identifier for their install of the game that persists across app restarts.
+        ///  <param name="response"> Our response to the approval process. In case of connection refusal with custom return message, we delay using the Pending field.
+        public override void ApprovalCheck(NetworkManager.ConnectionApprovalRequest request, NetworkManager.ConnectionApprovalResponse response)
+        {
+            var connectionData = request.Payload;
+            var clientId = request.ClientNetworkId;
+            if (connectionData.Length > k_MaxConnectPayload)
+            {
+                // If connectionData too high, deny immediately to avoid wasting time on the server. This is intended as
```
NoelStephensUnity, line 100: 🥇

Assets/Scripts/ConnectionManagement/ConnectionState/HostingState.cs
```diff
@@ -0,0 +1,166 @@
...
+                return ConnectStatus.IncompatibleBuildType;
+            }
+
+            return SessionManager<SessionPlayerData>.Instance.IsDuplicateConnection(connectionPayload.playerId) ?
+                ConnectStatus.LoggedInAgain : ConnectStatus.Success;
+        }
+
+        IEnumerator WaitToShutdown()
+        {
+            yield return null;
```
NoelStephensUnity, line 162: Just curious:
If the clients disconnect themselves upon receiving the message from the server (do they? still looking around), then would it be over-kill to wait for all of the clients to respond or was that problematic for other reasons (i.e. you want to quickly cancel/end a session and start a new one)?

NoelStephensUnity (commented):
I like the update and love the new state machine approach!
From what I can tell, it looks like a bunch of really good improvements.

Assets/Scripts/ConnectionManagement/ConnectionState/HostingState.cs
LPLafontaineB: In the end, I thought it made more sense for the clients to not disconnect themselves and wait for Netcode to do it for them when the server shuts down, but the other way around would have worked too.

As for waiting for a response from the clients, the host doesn't really care that all clients receive and handle that custom message, it's just something to let them know that this shut down is voluntary so that they don't try to reconnect to the server. After that, the ball is no longer in its court in my opinion.

fernando-cortez (approved):

PR #670, opened 2022-06-01
feat: allowing users to subscribe to a networked message channel while offline [MTT-3684] [MTT-3690]

LPLafontaineB:
### Description

This PR modifies the NetworkedMessageChannel to allow users to subscribe to it before initiating a connection. Now it registers handlers for CustomNamedMessages only when connecting. This also fixes an issue where on ulterior consecutive game sessions, it would not register them again if connecting to a new game after disconnecting from a previous one.

It also includes tests for the NetworkedMessageChannel.

### Issue Number(s)

[MTT-3684](https://jira.unity3d.com/browse/MTT-3684) & [MTT-3690](https://jira.unity3d.com/browse/MTT-3690)

### Contribution checklist


pdeschain (dismissed):

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
```diff
@@ -0,0 +1,204 @@
...
+
+            void OnEmptyMessageReceived(EmptyMessage message)
+            {
+                nbMessagesReceived++;
+            }
+
+            emptyMessageChannelServer.Publish(new EmptyMessage());
+
+            // wait for the custom named message to be sent on the server and received on the clients
+            yield return null;
```
SamuelBellomo, line 82: There should be a "WaitForCondition" in those test utilities somewhere, you can use this in all your tests

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
```diff
@@ -0,0 +1,204 @@
...
+            yield return null;
+
+            Assert.AreEqual(2, nbMessagesReceived);
+        }
+
+        [UnityTest]
+        public IEnumerator EmptyNetworkedMessageIsReceivedByAllSubscribersOnSingleClient()
```
SamuelBellomo, line 89: Since this seems like a copy of the above test, you can parameterize tests and specify "number of clients to spawn" for example and use only one test

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
```diff
@@ -0,0 +1,204 @@
...
+            yield return WaitForClientsConnectedOrTimeOut();
+
+            // Test sending a message a second time
+            emptyMessageChannelServer.Publish(new EmptyMessage());
+
+            // wait for the custom named message to be sent on the server and received on the clients
+            yield return null;
+            yield return null;
+
+            Assert.AreEqual(4, nbMessagesReceived);
```
SamuelBellomo, line 201: to make this clearer, I'd just set nbMessageReceived to 0 before the send, then check for 2 again.

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
```diff
@@ -0,0 +1,204 @@
...
+using Unity.Multiplayer.Samples.BossRoom.Shared.Infrastructure;
+using Unity.Netcode;
+using UnityEngine;
+using Unity.Netcode.TestHelpers.Runtime;
+using UnityEngine.Assertions;
+using UnityEngine.TestTools;
+
+namespace Unity.Multiplayer.Samples.BossRoom.Tests.Runtime
+{
+    public class NetworkedMessageChannelTests : NetcodeIntegrationTest
```
SamuelBellomo, line 11: You're testing for successful cases, you'd normally want to also test for edge cases and failures.

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
LPLafontaineB: When I did, Rider gave me a warning in OnEmptyMessageReceived because the captured variable is updated in the outer scope, but since this won't actually cause issues, I could ignore the warning

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
LPLafontaineB: good point, will add more coverage

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
LPLafontaineB: do you mean like creating a custom yield instruction that waits for two frames so that we can reuse it in all the tests?

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
LPLafontaineB: It's not exactly the same though. Above, we have two clients with one subscription each, and below we have one client with two subscriptions.

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
SamuelBellomo: no, I mean wait for the condition mentioned in your comment

> wait for the custom named message to be sent on the server and received on the clients

AFAIK you're not validating in your test that it takes 2 frames to receive that message, you're just validating it's received at all. So to make sure your test doesn't fail if that behaviour changes, you can yield return new WaitForCondition

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
SamuelBellomo: but if you think it'll add too much complexity, you can disregard this comment and resolve it.

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
SamuelBellomo: yeah, silly rider

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
LPLafontaineB: I should have replied to that comment after updating the PR. In the end, I simply parameterised both the number of clients and the number of subscriptions per client, as well as the message type. So yes this can be resolved

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
SamuelBellomo: wait, why is it a class property now?

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
LPLafontaineB: But the condition is the thing I'm testing for. I'm not sure I understand how that would work here. With a `new WaitUntil(m_NbMessagesReceived==expected)` for example, it would hang indefinitely if the client doesn't receive the expected number of messages and would return early with a success if the client receives more than the expected number of messages

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
LPLafontaineB: I moved it as a field so that I could extract the code creating the message channels and the subscriptions into a reusable method to reduce code duplication

SamuelBellomo: as soon as you add persistent state to your tests, you risk having spreading corruption between test runs. Each test should have its own isolated world as much as possible. Either this or you have setup and teardown methods that'll make sure they run even when a test fails (an assert that fails throws an exception, interrupting the test and not running any teardown you have in the actual test, you need to have teardown methods to make sure they still run at the end)

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
SamuelBellomo: Can you guarantee it should always be two frames of wait time?
you can have timeouts in those wait conditions if it never reaches it or goes beyond.

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
```diff
@@ -0,0 +1,263 @@
...
+                {
+                    Debug.Log($"Received message on server in subscription {numSub}.");
+                    m_NbMessagesReceived++;
+                    Assert.AreEqual(message, expectedValue);
+                }));
+            }
+        }
+
+        [UnityTest]
+        public IEnumerator EmptyNetworkedMessageIsReceivedByAllSubscribersOnAllClientsAndServer([ValueSource(nameof(s_NbClients))] int nbClients, [ValueSource(nameof(s_NbSubs))] int nbSubscribers)
```
SamuelBellomo, line 111: for something as simple as a list of ints, you can use Values
https://docs.nunit.org/articles/nunit/writing-tests/attributes/values.html
ValueSource is mostly for runtime generated values

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
LPLafontaineB: Yes, that is why that field is set to 0 on Setup

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
LPLafontaineB: I cannot guarantee that. Timeouts do prevent indefinitely hanging, but not early returns with successes. For example, some of these tests expect 0 messages to be received, but I still want to make sure to wait enough time to detect if more messages are received so that the test can fail.

To that end, I ended up changing waiting for two frames to waiting for 0.2 seconds, as is the case in NGO's custom named messages tests.

SamuelBellomo: Waiting for times has issues as well. you don't have a guarantee you'll have more than a frame during that time. If there's a hang, that frame could last 0.2+ seconds.

> but I still want to make sure to wait enough time to detect if more messages are received so that the test can fail.

How many of those do you have? Thinking of it, we want our tests to execute as fast as possible. If we have big times like this for a bunch of tests, it'll just increase the time it takes for our PRs to be approved. Are those waits critical enough to add delays to ALL our PRs approvals?

SamuelBellomo (dismissed):
approving after comments are addressed

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
LPLafontaineB: > How many of those do you have?

with all the tests, it amounts to a total wait time of 12.8 seconds.

> Waiting for times has issues as well. you don't have a guarantee you'll have more than a frame during that time. If there's a hang, that frame could last 0.2+ seconds.

Should I go back to waiting two frames then? And adjust these tests if the sdk changes this behaviour in the future?

Assets/Tests/Runtime/NetworkedMessageChannelTests.cs
```diff
@@ -0,0 +1,250 @@
...
+            {
+                m_ClientScopes[i].Dispose();
+            }
+            m_ServerScope.Dispose();
+            return base.OnTearDown();
+        }
+
+        void InitializeNetworkedMessageChannels<T>(int nbClients, int nbSubscribers, T expectedValue, out NetworkedMessageChannel<T>[] emptyMessageChannelClients, out NetworkedMessageChannel<T> emptyMessageChannelServer) where T : unmanaged, INetworkSerializeByMemcpy
+        {
+            emptyMessageChannelClients = new NetworkedMessageChannel<T>[nbClients];
```
SamuelBellomo, line 67: wait, how do you associate your message channel with your client and server NetworkManager? NMC uses NetworkManager's Singleton...

SamuelBellomo: nvm, you inject it

SamuelBellomo (dismissed):

fernando-cortez (commented):
I think the only thing that I would want to see is more verbosity on the error logging for failures of each test. As in, if tests were to eventually fail, would we be able to at a quick glance know where to look for what failed? Especially for someone new to the codebase.

SamuelBellomo (approved):

pdeschain:
this PR needs to get conflicts resolved

LPLafontaineB:
Converting to draft until issues with tests are resolved

pdeschain (approved):

Assets/Scripts/ApplicationLifecycle/ApplicationController.cs
```diff
@@ -30,6 +30,7 @@ protected override void Configure(IContainerBuilder builder)
         {
             base.Configure(builder);
             builder.RegisterComponent(m_UpdateRunner);
+            builder.RegisterComponent(m_ConnectionManager.NetworkManager);
```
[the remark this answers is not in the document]
LPLafontaineB: There are still many places where the singleton access is used. Should I change this as part of this PR or should that be a separate task? cc @SamuelBellomo

fernando-cortez (approved):

Assets/Scripts/ApplicationLifecycle/ApplicationController.cs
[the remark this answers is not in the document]
SamuelBellomo: Why would we want to replace all refs to NetworkManager by injection??
Let's do it as needed. I'd see a reason for replacement say if we want to add automated tests for them. But I'd be surprised if we had a lot of coverage in boss room.

SamuelBellomo (approved):

PR #675, opened 2022-06-08
fix: subscribing to a message channel while unsubscribing is pending [MTT-3756]

LPLafontaineB:
### Description

This PR makes sure that when subscribing to a message channel after having previously subscribed and unsubscribed to it works properly, even if no messages have been published in the mean time. To do so, we keep the list of pending additions and removals from the list of handlers in a dictionary. If we subscribe while a pending unsubscribe is in the dictionary, we simply remove it from the dictionary instead of adding a new pending subscription, and we do the same when unsubscribing if we have a pending subscription.

### Issue Number(s)

[MTT-3756](https://jira.unity3d.com/browse/MTT-3756)

### Contribution checklist


pdeschain (approved):
We discussed it in Slack, implementation looks good 👍

Assets/Scripts/Infrastructure/PubSub/MessageChannel.cs
```diff
@@ -8,11 +8,10 @@ public class MessageChannel<T> : IMessageChannel<T>
     {
```
SamuelBellomo, line 8: What in-game issue is this fixing?
If we need to maintain this code, might was well include a 3rd party messaging library instead.
If we keep this, we need to add tests and release notes for this.

LPLafontaineB: In our current develop, we don't have actual in-game issues with it because of the way we use message channels. We had an issue with a previous iteration of #666 though, when subscribing and unsubscribing to a channel in a state's OnEnter and OnExit events. The latest version doesn't cause this issue anymore but I still think that this is a fix we should do, so that we don't end up having to re-do this whole investigation later when we encounter this issue again.

SamuelBellomo: @pdeschain we talked about how if we had to maintain this code, we'd look into a 3rd party. I consider the above maintaining. What would it take to build networked message publishing with a 3rd party lib?

SamuelBellomo: As discussed offline, there's work to maintain in both cases.
nvm, carry on

SamuelBellomo (approved):

SamuelBellomo (changes requested):
Let's add tests for this case, so there's no regression in the future. This way, even though we don't have a use case in game, we'd have tests to make sure this behaviour doesn't change later.

pdeschain (approved):

PR #692, opened 2022-07-21
feat: connection management tests [MTT-3969]

LPLafontaineB:
### Description

This PR adds tests for connection management.

It also modifies ConnectionManager and the ConnectionStates to allow for easier testing. It passes the NetworkManager via DI and removes usage of NetworkManager.Singleton inside it. It also makes it so the testing assembly can access its internal fields and methods.

Finally, it modifies SceneLoaderWrapper so that it can easily be inherited and its main behavior can be overwritten. This is used in the tests to create a stub so that the tests don't load any scene.

### Issue Number(s)

[MTT-3969](https://jira.unity3d.com/browse/MTT-3969)

### Contribution checklist


LPLafontaineB:
The tests are currently not passing because of a bug with NetworkedMessageChannels that is fixed in #670 . After 670 is merged, this will be fine

Assets/Scripts/ConnectionManagement/ConnectionState/HostingState.cs
```diff
@@ -25,15 +25,17 @@ class HostingState : ConnectionState
 
         public override void Enter()
         {
-            var gameState = UnityEngine.Object.Instantiate(m_ConnectionManager.GameState);
-
-            gameState.Spawn();
+            foreach (var prefab in m_ConnectionManager.NetworkObjectsToSpawnWhenServerStarted)
```
SamuelBellomo, line 28: ConnectionManagement shouldn't be in charge of spawning objects like this. This should be done by the appropriate game state, no?
cc @pdeschain who's working in that area as well.

Assets/Scripts/ConnectionManagement/ConnectionState/HostingState.cs
```diff
@@ -25,15 +25,17 @@ class HostingState : ConnectionState
...
+                var networkObject = UnityEngine.Object.Instantiate(prefab);
+                networkObject.Spawn();
+            }
 
             SceneLoaderWrapper.Instance.AddOnSceneEventCallback();
 
             //The "BossRoom" server always advances to CharSelect immediately on start. Different games
             //may do this differently.
-            SceneLoaderWrapper.Instance.LoadScene("CharSelect", useNetworkSceneManager: true);
+            SceneLoaderWrapper.Instance.LoadScene(m_ConnectionManager.StartScene, useNetworkSceneManager: true);
```
SamuelBellomo, line 38: See my DGS PR for this, connection management shouldn't manage scene names.

Assets/Tests/Runtime/ConnectionManagementTests.cs
```diff
@@ -0,0 +1,593 @@
...
+
+        [Test]
+        public void StartHost_Success()
+        {
+            StartHost();
+            Assert.IsTrue(m_ServerNetworkManager.IsHost);
+            var expectedServerConnectionStateSequence = new List<ConnectionState>();
+            expectedServerConnectionStateSequence.Add(m_ServerConnectionManager.m_StartingHost);
+            expectedServerConnectionStateSequence.Add(m_ServerConnectionManager.m_Hosting);
+            Assert.AreEqual(expectedServerConnectionStateSequence, m_ServerConnectionStateSequence);
```
SamuelBellomo, line 184: Is this test's goal to test internal states themselves or to test the behaviour of the state machine? An issue with testing the internal sequence of a state machine is you're not refactor friendly with this. Usually, when testing state machines, you want to test the inputs and the effects of it, not how it behaves internally. Is it really that important that the state machine went from state 1 to 2 to 3? Usually no as long as the right inputs produced the right outputs.

Assets/Tests/Runtime/ConnectionManagementTests.cs
```diff
@@ -0,0 +1,593 @@
...
+        }
+
+        [UnityTest]
+        public IEnumerator ClientConnectingWithoutHost_ConnectionFailed()
+        {
+            SetUniqueProfilesForEachClient();
+
+            for (var i = 0; i < NumberOfClients; i++)
+            {
+                LogAssert.Expect(LogType.Error, k_FailedToConnectToServerErrorMessage);
```
SamuelBellomo, line 308: you usually don't want to test for string error messages, but for exception types instead. error messages are very likely to change, less for class types.

Assets/Tests/Runtime/ConnectionManagementTests.cs
```diff
@@ -0,0 +1,593 @@
...
+        {
+            StartHost();
+            Assert.IsTrue(m_ServerNetworkManager.IsHost);
+
+            SetUniqueProfilesForEachClient();
+
+            yield return ConnectClients();
+            for (var i = 0; i < NumberOfClients; i++)
+            {
+                Assert.IsTrue(m_ClientNetworkManagers[i].IsConnectedClient);
```
SamuelBellomo, line 198: What else happens after a connection is established? Can we test that more things happened?

Assets/Tests/Runtime/ConnectionManagementTests.cs
```diff
@@ -0,0 +1,593 @@
...
+            }
+
+            yield return ConnectClients();
+
+            Assert.IsTrue(m_ClientNetworkManagers[0].IsConnectedClient);
+            for (var i = 1; i < NumberOfClients; i++)
+            {
+                Assert.IsFalse(m_ClientNetworkManagers[i].IsConnectedClient);
+            }
+
```
SamuelBellomo, line 278: What else can you test with this? Can we test the reason reached the clients? Can we test clients are in the proper state after this?

Assets/Tests/Runtime/ConnectionManagementTests.cs
```diff
@@ -0,0 +1,593 @@
...
+            for (var i = 0; i < NumberOfClients; i++)
+            {
+                var profileManager = m_ClientScopes[i].Container.Resolve<ProfileManager>();
+                profileManager.Profile = $"Client";
+            }
+
+            yield return ConnectClients();
+
+            Assert.IsTrue(m_ClientNetworkManagers[0].IsConnectedClient);
+            for (var i = 1; i < NumberOfClients; i++)
```
SamuelBellomo, line 274: You might want to add some comments around this, from an external point of view, it's hard to read this test without knowing that implicitly this uses the fact that we can connect only once with the same profile.

Assets/Tests/Runtime/ConnectionManagementTests.cs
SamuelBellomo: unless you want to test each states individually, but then you'd want to stub a bunch of things and create unit tests for each state and thoroughly test those.
It's the same thinking as if you were testing a method with inputs and return values. You wouldn't care about it's internal state, as long as it gave you the right output.

Assets/Tests/Runtime/ConnectionManagementTests.cs
```diff
@@ -0,0 +1,593 @@
...
+            for (var i = 0; i < NumberOfClients; i++)
+            {
+                Assert.IsTrue(m_ClientNetworkManagers[i].IsConnectedClient);
+            }
+
+            // Shutting down the server
+            m_ServerNetworkManager.Shutdown();
+            // ConnectionManager should also transition to Offline state normally, but OnClientDisconnectCallback is not
+            // invoked on the host when shutting down, so transitioning here manually
+            m_ServerConnectionManager.ChangeState(m_ServerConnectionManager.m_Offline);
```
SamuelBellomo, line 394: If this doesn't happen boss room side, why are we doing it here?
This seems like cheating. It looks like it's related to the issue where a host doesn't get any callbacks when relay disappears, is that it? I wouldn't workaround this here. I'd add some comment on this and create a test that fails and add a [Ignore] attribute on it to keep track of it.

Assets/Scripts/ConnectionManagement/ConnectionState/HostingState.cs
LPLafontaineB: I agree, however changing this is not in the scope of this PR in my opinion

Assets/Scripts/ConnectionManagement/ConnectionState/HostingState.cs
LPLafontaineB: I agree with this as well, but again I don't think it is in the scope of this PR

Assets/Tests/Runtime/ConnectionManagementTests.cs
LPLafontaineB: Yes that is it. Adding the [Ignore] attribute instead of this workaround.

Assets/Tests/Runtime/ConnectionManagementTests.cs
LPLafontaineB: That is a good point, I'll simplify this to only concern myself with the end results instead of the internal states. That way, I could also remove the AssemblyInfo.cs that I added to make internals visible to the tests

Assets/Tests/Runtime/ConnectionManagementTests.cs
LPLafontaineB: On the client's side, nothing that concerns the ConnectionManager that is not already tested by NGO. On the host, we could test the state of the SessionManager, but I'm not entirely convinced this is pertinent to what is being tested here?

Assets/Tests/Runtime/ConnectionManagementTests.cs
LPLafontaineB: Unless we want to keep testing the final states at the end of the tests? These are internal, but they affect how inputs are handled from the outside

Assets/Tests/Runtime/ConnectionManagementTests.cs
LPLafontaineB: Yes we could test the reception of the reasons messages, and if we want to we could also test the final state

Assets/Scripts/ConnectionManagement/ConnectionState/HostingState.cs
SamuelBellomo: If you agree my DGS PR is the way to go to manage those names vs having them in assets, could you revert this so this makes merging easier with less conflicts? This way I wouldn't have to revert your change you're introducing with this PR

Assets/Tests/Runtime/ConnectionManagementTests.cs
SamuelBellomo: As discussed, testing the state machine is higher prio than individual states.

Assets/Tests/Runtime/ConnectionManagementTests.cs
SamuelBellomo: logs are there to help you debug. Why are we testing for logs? If there are no exceptions and just error logs to help you debug, you shouldn't test for this. If it fails later, because we changed our logs, this will just create low priority noise.

LPLafontaineB: What I meant is that these error logs already happen during those tests. These make tests fail if they are not expected ([see here](https://docs.unity3d.com/Packages/com.unity.test-framework@1.1/api/UnityEngine.TestTools.LogAssert.html)), which is why we must add this here.

Assets/Tests/Runtime/ConnectionManagementTests.cs
```diff
@@ -0,0 +1,514 @@
...
+
+            var lifeTimeScopeGO = new GameObject($"LifeTimeScope - {name}");
+            scope = lifeTimeScopeGO.AddComponent<ConnectionManagementTestsLifeTimeScope>();
+            scope.NetworkManager = networkManager;
+            scope.ConnectionManager = connectionManager;
+            scope.UpdateRunner = updateRunner;
+            scope.Build();
+        }
+
+        void CreatePlayerPrefab()
```
fernando-cortez, line 96: From offline discussion, check if this is still needed.

Assets/Tests/Runtime/ConnectionManagementTests.cs
```diff
@@ -0,0 +1,514 @@
...
+            // Waiting for client to automatically reconnect
+            yield return WaitForClientsConnectedOrTimeOut();
+            Assert.IsTrue(m_ClientNetworkManagers[0].IsConnectedClient, "Client0 failed to reconnect.");
+
+            Assert.AreEqual(1, nbReconnectingMsgsReceived, "No Reconnecting message received.");
+            subscriptions.Dispose();
+        }
+
+
+        [UnityTest, Ignore("Test fails because shutdowns do not invoke OnClientDisconnect on the host, so the ConnectionManager doesn't properly transition to the Offline state")]
```
fernando-cortez, line 361: If this is a Netcode bug that is tracked, we should track on our end to come back to this test and re-enable.

Assets/Tests/Runtime/ConnectionManagementTests.cs
```diff
@@ -0,0 +1,514 @@
...
+            CreatePlayerPrefab();
+
+            StartHost();
+            AssertHostIsListening();
+
+            yield return ConnectClients();
+            AssertAllClientsAreConnected();
+
+            // Switching back references for Client0 and Server
+            (m_ServerNetworkManager, m_ClientNetworkManagers[0]) = (m_ClientNetworkManagers[0], m_ServerNetworkManager);
```
fernando-cortez, line 456: See if this is needed to revert back roles since you also shutdown all clients as well as host.

fernando-cortez (approved):
Left some comments. Mostly double checks and some cleanup. LGTM,

Assets/Tests/Runtime/ConnectionManagementTests.cs
LPLafontaineB: after double-checking, it is still needed. Made the reason clearer in the comments

Assets/Tests/Runtime/ConnectionManagementTests.cs
LPLafontaineB: Not needed anymore, will remove

pdeschain (approved):

Assets/Tests/Runtime/ConnectionManagementTests.cs
fernando-cortez: Create a task to come back to this and we're gtg!

SamuelBellomo (approved):

pdeschain (approved):

fernando-cortez (approved):
Could link that Jira you created to the method summary itself. But it's a nitpick. Gtg.

SamuelBellomo (approved):

PR #695, opened 2022-07-22
Dedicated server support first pass - part 1/2 manual stripping [MTT-1936]

SamuelBellomo:
### Description

Part 1/2 for first pass at adding DGS mode to boss room.

Stripping assets and making client scripts non-networkBehaviour to not mess with NetworkBehaviour indexing when they are stripped. Since stripping involves removing the script from the build, attached components on GameObjects become null, but still occupy a space in NetworkBehaviour's indexing if they were NetworkBehaviours. Things that could become client only now are now always offline and rely on other components to provide netcode hooks.

There are 3rd party tools to strip assets and scripts from builds. This PR shows what's involved to do it manually, for a small portion of boss room. Subsequent work would need to explore this in further details.

This PR strips the following:
- Char select now has it's client only elements in a new separate scene that's loaded additively on clients
- Client only scripts don't rely on NetworkBehaviour anymore
- Tests relying on client side scripts.

#### TODO SAM before merging this
- [ ] After asmdef move, a lot of those scripts are not stripped anymore. Another move to a separate asmdef might be needed. However, there are tools that exist that wouldn't require a move to a separate asmdef and would allow just selecting the scripts to strip. Need to take a decision on this for this first pass.

- [ ] Stripping perf notes:

With release build, VM config: 10GB mem, 8 cores
|       | CPU | Mem |
| --- | --- | --- |
| DGS linux | 35%-55% of one core | 2 players=344MB, 1 player=249MB, 0 player char select=99MB at start or 153MB when going back to it  |
| Client hosted mac | --- | 1 client 1 host char select, host stats: 477MB |

- [ ] why mega memory bump on each player connection? --> could be the UTP queue size we've set, which is about 100MB
- [ ] test with 8 players connected
- [x] test with more reasonable VM settings
- [ ] why no memory bump when going from char select to main room?
- [ ] 20+threads?
- [x] reduce frame rate more (30?)
- [ ] why can't pass through door? Only on DGS??

This is a simplified branch to help with review. Original work branch for history purposes: origin/sam/feat/dedicated-server-support

Dedicated Server build target on linux, with build stripping + extra char select manually stripped:
Client hosted with desktop build target on mac, with no stripping at all:

### Issue Number(s)

[MTT-1936](https://jira.unity3d.com/browse/MTT-1936)

### Contribution checklist


SamuelBellomo:
Tests aren't passing because they need changes in Part 2. Let's review all of this assuming tests pass, I'll merge Part 2 in Part 1 once they are both reviewed. We'll then be able to see tests pass.

PR #716, opened 2022-08-24
feat: NetworkTransform configurations to reduce bandwidth usage [MTT-1203]

LPLafontaineB:
### Description

Currently, we are synchronizing a lot of data through our NetworkTransform components that doesn't need to be synchronized. For example, characters synchronize their position, rotation and scale in their entirety, which leads to updates of 47B each time they are moving.
This PR reduces the bandwidth used by configuring our NetworkTransform components on our different NetworkObjects to only synchronize what is needed. For characters, pickable pots and arrows, this means the (x,z) components of their position and the (y) component of their rotation. For tossed items, this means everything except the scale.
To use the characters as an example again, this reduces the size of the updates to only 23B.

This PR also restores the default thresholds used to determine when a synchronization is needed.

### Issue Number(s)

[MTT-1203](https://jira.unity3d.com/browse/MTT-1203)

### Contribution checklist


pdeschain (dismissed):

fernando-cortez (approved):

Assets/Prefabs/Character/Character.prefab
```diff
@@ -275,15 +278,14 @@ MonoBehaviour:
...
-  SyncScaleY: 1
-  SyncScaleZ: 1
-  PositionThreshold: 0
-  RotAngleThreshold: 0
-  ScaleThreshold: 0
+  SyncRotAngleZ: 0
+  SyncScaleX: 0
+  SyncScaleY: 0
+  SyncScaleZ: 0
+  PositionThreshold: 0.001
```
SamuelBellomo, line 287: Did you test with higher values? This value is basically millimeters of precision, this seems quite a high precision.

SamuelBellomo: same with all others

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ Additional documentation and release notes are available at [Multiplayer Documen
 * NetworkedMessageChannels can now be subscribed to before initiating a connection (#670)
 * Refactored connection management into simpler state machine (#666)
 * Merged GameState bridge classes (the ones that contained no or limited functionality) (#697) This cleans up our sometimes too verbose code split.
+* Configured the NetworkTransform components of every NetworkObject to reduce the bandwidth usage (#716). This prevents the unnecessary synchronization of data that clients do not need, i.e. a character's scale or y position.
```
SamuelBellomo, line 21: It could be interesting to give a quick bandwidth stat. Like "we went from x to y kb/s in this z scenario"

Assets/Prefabs/Character/Character.prefab
LPLafontaineB: I did. I could increase this up to a certain point before it negatively impacted gameplay, but there were no bandwidth gains. However, when testing with the develop branch of NGO, the impact of thresholds was much higher. I would recommend that if we want to update those thresholds, we do so after those changes have been released and integrated into Boss Room

SamuelBellomo: could you test with that branch and check for those thresholds?
You'd see bandwidth gains in cases where there's no that much position or rotation changes (I'm thinking on those situations where imps gang up on you and push you slowly while you don't move yourself)

CHANGELOG.md
SamuelBellomo: I'd still be interested in the bandwidth usage, in addition to the character sizes. At the end of the day, it's total bandwidth we care about.

SamuelBellomo (approved):

PR #719, opened 2022-08-25
fix: game frozen when client leaves in postgame [MTT-4342]

LPLafontaineB:
### Description

This PR changes when we register and un-register to NGO callbacks within ServerBossRoomState, making it simpler and clearer. It is now only done once in OnNetworkSpawn and undone once in OnNetworkDespawn. Before this, it relied on OnLoadComplete to register those callbacks (which happens every time any scene is loaded, including the additive scenes) and on OnUnloadeComplete to un-register them (which only happens when an additive scene is unloaded). This meant that we would register to these callbacks multiple times and we would not un-register properly afterwards.

Here are the repro steps for the bug this is fixing:

1.Start a game with a host and at least one client
2.Progress through the game up to the postgame state
3.As a client, leave the game

No addition to changelog since this bug was introduced after the 1.3.0 release

### Issue Number(s)

[MTT-4342](https://jira.unity3d.com/browse/MTT-4342)

### Contribution checklist

### DIFF
--- a/Assets/Scripts/Gameplay/GameState/ServerBossRoomState.cs
+++ b/Assets/Scripts/Gameplay/GameState/ServerBossRoomState.cs
@@ -79,25 +79,6 @@ namespace Unity.Multiplayer.Samples.BossRoom.Server
 
             m_Subscription = m_LifeStateChangedEventMessageSubscriber.Subscribe(OnLifeStateChangedEventMessage);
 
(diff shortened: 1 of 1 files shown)


SamuelBellomo (approved):
I think I also fixed this the same way in my unmerged DGS PR, thanks for looking at this

fernando-cortez (approved):

PR #737, opened 2022-09-23
feat: upgrade lobby [MTT-3962][MTT-4105][MTT-4052]

LPLafontaineB:
### Description

This PR updates the Lobby package to its latest version (1.0.3) and changes our usage of the deprecated Lobbies with the new LobbyService.

It also uses the new reconnect API to simplify how we deal with reconnecting automatically to a lobby. Now, we no longer need to have the host manually remove the client from the lobby, and the client doesn't have to manually do it either before attempting to reconnect. I also added a small cooldown between each reconnect attempt, so that time can pass between each attempt so that the issue causing the disconnect could have time to resolve itself.

Finally, it also fixes an issue where we would not be sending the proper relay allocationId to the lobby (we simply called .toString() on the byte array that relay returns to us, which simply returns the name of the class). The issue is fixed by using the Guid that Relay returns instead of the byte array.

Those changes remove most of the error logs that were happening, but we still get those when we are rate limited or when we lose connection to the service.

### Issue Number(s)

MTT-3962
MTT-4105
MTT-4052

### Contribution checklist


Assets/Scripts/ConnectionManagement/ConnectionState/ClientReconnectingState.cs
```diff
@@ -81,19 +84,33 @@ IEnumerator ReconnectCoroutine()
             m_NbAttempts++;
             if (!string.IsNullOrEmpty(m_LobbyCode))
             {
-                var leavingLobby = m_LobbyServiceFacade.EndTracking();
-                yield return new WaitUntil(() => leavingLobby.IsCompleted);
-                var joiningLobby = m_LobbyServiceFacade.TryJoinLobbyAsync("", m_LobbyCode);
-                yield return new WaitUntil(() => joiningLobby.IsCompleted);
-                if (joiningLobby.Result.Success)
+                // Attempting to reconnect to lobby
```
SamuelBellomo, line 87:
```suggestion
                // Attempting to reconnect to lobby
                // Timeout configs have been set in Lobby's dashboard
```

Assets/Scripts/ConnectionManagement/ConnectionState/ClientReconnectingState.cs
```diff
@@ -21,6 +21,9 @@ class ClientReconnectingState : ClientConnectingState
         string m_LobbyCode = "";
         int m_NbAttempts;
 
+        const float k_TimeBetweenLobbyReconnectAttempts = 1;
+        const float k_NbLobbyReconnectAttempts = 5;
```
SamuelBellomo, line 25: any reason these are different from UTP reconnect attempts?

SamuelBellomo (dismissed):

Assets/Scripts/ConnectionManagement/ConnectionState/ClientReconnectingState.cs
LPLafontaineB: I could use the parameters we have set to our instance of UTP instead of these constants, but that wouldn't make sense in my opinion, since they don't serve the same purpose

Assets/Scripts/ConnectionManagement/ConnectionState/ClientReconnectingState.cs
LPLafontaineB: I'm not sure I understand what this comment aims to convey

SamuelBellomo: This is mostly to show users where timeouts are set for this, since we can't point anywhere in code

Assets/Scripts/ConnectionManagement/ConnectionState/ClientReconnectingState.cs
SamuelBellomo: I was talking about ConnectionManager's m_NbReconnectAttempts

LPLafontaineB: Oh I see. So m_NbReconnectAttempts is the number of times where we try to reconnect to the game. But for each attempt, if we are using lobby, we try to reconnect to lobby k_NbLobbyReconnectAttempts times, before considering the reconnect attempt as having failed and trying again. The wording is a bit confusing though, I'm open to suggestions to make this clearer.

Assets/Scripts/ConnectionManagement/ConnectionState/ClientReconnectingState.cs
LPLafontaineB: I see, but the timeout set in the dashboard doesn't come into play here specifically. It is the amount of time between when a client is marked as being disconnected from a Relay server and when the Lobby service removes them from the Lobby. I'll write a more detailed comment to explain it

Assets/Scripts/ConnectionManagement/ConnectionState/ClientReconnectingState.cs
SamuelBellomo: So let me get this straight. A disconnect from lobby happens when the heartbeats to lobby stop. Once the clinet comes back, it can call the reconnect api. There's multiple failure situations here:
client internet is down:
- client will try to call reconnect to lobby multiple times, this will fail. service side, this should timeout the relay connection
host is down:
- lobby reconnect passes with no retry
- will attempt multiple time to connect through UTP/Relay, this will fail.
Everything is slow and unstable:
- could reconnect through lobby after 4/5 tries
- could reconnect to relay/utp after 2/3 tries
For this last case, any reason why we're not just counting the number of tries as a whole connection flow? "You have 3 chances to do the whole flow including lobby and utp/relay, else you timeout"? Why the two steps with different retries? Having a single flow would work as well for the above two scenarios too no?

LPLafontaineB: As mentioned in the PR description, the goal was to allow some time between the attempts so that if it is a temporary issue, it could be solved. Otherwise, in your first scenario, we would simply send x number of requests back to back, which would both fail anyways.

For your last scenario, the flow actually is that we have three chances to do the whole reconnect flow. However, for each of those chances, we try to reconnect to lobby multiple times before marking that attempt as having failed and starting a new one, the same way as UTP which tries to connect to the relay server multiple times before marking it as a failed attempt when we call StartClient().

Another simpler solution though could be to only try to reconnect to lobby once per attempt, but adding a cooldown between each attempt. The only thing I'm not sure about with this other solution is for UX, when to update the number of attempts on the popup, after the failed attempt and before the cooldown, or after the cooldown and before the next attempt?

SamuelBellomo: +1 on single reconnect per attempt
regarding cooldown, that's actually good to have a cooldown on relay/utp as well.
I'd also link to this page on exponential backoff
https://en.wikipedia.org/wiki/Exponential_backoff
I don't think we have time to implement this ourselves, but you could add a note that this implementation could be up to users.
Regarding UX, to me cooldown is part of the connection time as a player. So update after cooldown? Before next attempt?
It'd also be great to have some of this discussion summarized in comments, those are good things to think about for users.

LPLafontaineB: Done, and updated the PR description accordingly

SamuelBellomo (dismissed):

SamuelBellomo (approved):

SamuelBellomo (approved):

pdeschain (approved):

PR #788, opened 2022-11-28
feat: bumping RNSM to 1.1.0

SamuelBellomo:
### Description

Bumped RNSM to 1.1.0
Switched x axis units to seconds instead of frames now that it's available. This means adjusting the sample count to a lower value as well to 30 seconds, since the x axis was moving too slowly.

### Contribution checklist
 -  Tests have been added for boss room and/or utilities pack


ian-m-unity:
Minor question, is it common to store .asset files in Git LFS?

SamuelBellomo:
> Minor question, is it common to store .asset files in Git LFS?

@ian-m-unity in my experience, they are hard to merge, you usually want to treat them as binary. But it's true that diffs should be pretty small, so storing them in git LFS might be overkill? We didn't tweak this too much and just said binary==LFS.
It might be worth it to tweak this more for future projects, I agree.

SamuelBellomo:
> Minor question, is it common to store .asset files in Git LFS?
> 
> Edit: I've always thought of these as text files, but apparently some .asset files are binary 😅

And yeah, even the text format is hard to merge anyway

PR #798, opened 2022-12-14
feat: Adding build menus to automatically build all platforms for our platests

SamuelBellomo:
### Description

This adds a new "Boss Room/Playtest Builds" menu that contains different build options and toggles. This will allow to build all required platforms in one go without having to change build target manually and making sure project ID is setup.
This is meant as an internal tool for Unity folks, not for users (usual manual build flow for users must remain valid).

Not adding this to our release note as this is not meant for users.

This is not integrated in CI (yet). This can be called manually while waiting for future CI implementation.

Screen Shot 2022-12-14 at 6 00 52 PM

### Contribution checklist

### DIFF
--- a/Assets/Scripts/Editor/BuildHelpers.cs
+++ b/Assets/Scripts/Editor/BuildHelpers.cs
@@ -168,7 +168,7 @@ internal static class BuildHelpers
         ToggleMenu(k_SkipAutoDeleteToggleName);
     }
 
-    static bool ToggleMenu(string menuName, bool? valueToSet=null)
+    static bool ToggleMenu(string menuName, bool? valueToSet = null)
     {
         bool toSet = !Menu.GetChecked(menuName);
         if (valueToSet != null)
--- a/Assets/Scripts/Editor/BuildHelpers.cs
+++ b/Assets/Scripts/Editor/BuildHelpers.cs
@@ -31,6 +31,7 @@ internal static class BuildHelpers
     const int k_MenuGroupingOtherToggles = 22;
 
     static BuildTarget s_CurrentEditorBuildTarget;
+    static BuildTargetGroup s_CurrentEditorBuildTargetGroup;
     static int s_NbBuildsDone;
 
     static string BuildPathRootDirectory => Path.Combine(Path.GetDirectoryName(Application.dataPath), "Builds", "Playtest");
@@ -86,12 +87,13 @@ internal static class BuildHelpers
     static void RestoreBuildTarget()
     {
         Debug.Log($"restoring editor to initial build target {s_CurrentEditorBuildTarget}");
-        EditorUserBuildSettings.SwitchActiveBuildTarget(s_CurrentEditorBuildTarget);
+        EditorUserBuildSettings.SwitchActiveBuildTarget(s_CurrentEditorBuildTargetGroup, s_CurrentEditorBuildTarget);
     }
 
     static void SaveCurrentBuildTarget()
     {
         s_CurrentEditorBuildTarget = EditorUserBuildSettings.activeBuildTarget;
+        s_CurrentEditorBuildTargetGroup = EditorUserBuildSettings.selectedBuildTargetGroup;
     }
 
     [MenuItem(k_AllToggleName, false, k_MenuGroupingPlatforms)]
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Editor/BuildHelpers.cs
```diff
@@ -0,0 +1,235 @@
...
+using System.Threading.Tasks;
+using UnityEngine;
+using UnityEditor;
+using UnityEditor.Build.Reporting;
+
+/// <summary>
+/// Utility menus to easily create our builds for our playtests. If you're just exploring this project, you shouldn't need those. They are mostly to make
+/// multiplatform build creation easier and is meant for internal usage.
+/// </summary>
+internal static class BuildHelpers
```
fluong6, line 14: Should we wrap this functionality behind a scripting symbols? Otherwise that "feature" will be released to the public and I am not sure if we want to support it

Assets/Scripts/Editor/BuildHelpers.cs
```diff
@@ -0,0 +1,235 @@
...
+{
+    const string k_MenuRoot = "Boss Room/Playtest Builds/";
+    const string k_Build = k_MenuRoot + "Build";
+    const string k_DeleteBuilds = k_MenuRoot + "Delete All Builds (keeps cache)";
```
fluong6, line 18: Is there any benefits in having a "Delete All Builds and cache"? Or we always want to keep the cache?

Assets/Scripts/Editor/BuildHelpers.cs
SamuelBellomo: So this is in the Editor folder, in boss room's project (not in our Utilities package). Editor folders are already excluded from builds. 
We also don't really have the concept of "APIs to support" in samples, at least not the same way Packages do. We write code examples that users study and copy to their own games.
If it was in our Utilities package, I would have agreed though, this one is closer to other packages.

fluong6:
General question: if we want to hook this in the CI, do we eventually have to support command line args?

Assets/Scripts/Editor/BuildHelpers.cs
SamuelBellomo: mmm I'd say we can add this in a followup PR if we need it and if we see there's issues with local build caching? This is for local usage for now with no automation, we can still do things manually if need be. But yes do let me know if you'd like to see improvements for this in the future when using it.

SamuelBellomo:
> General question: if we want to hook this in the CI, do we eventually have to support command line args?

CI allows calling static methods directly. However right now I do rely on Menu.GetChecked which isn't super CI friendly, we'd need to convert this if we want to have those options available in CI. We could also just have a default of "build all" for CI :P which would simplify this.

SamuelBellomo:
> > General question: if we want to hook this in the CI, do we eventually have to support command line args?
> 
> CI allows calling static methods directly. However right now I do rely on Menu.GetChecked which isn't super CI friendly, we'd need to convert this if we want to have those options available in CI. We could also just have a default of "build all" for CI :P which would simplify this.

@fluong6 ^^^

fluong6:
> CI allows calling static methods directly

Oh cool, didnt know that. Thanks for info! ^_^

gregk-unity (dismissed):

Assets/Scripts/Editor/BuildHelpers.cs
```diff
@@ -0,0 +1,243 @@
...
+
+        SaveCurrentBuildTarget();
+
+        try
+        {
+            // deleting so we don't end up testing on outdated builds if there's a build failure
+            if (!skipAutoDelete) DeleteBuilds();
+
+            if (buildiOS) await BuildPlayerUtilityAsync(BuildTarget.iOS, "", true);
+            if (buildAndroid) await BuildPlayerUtilityAsync(BuildTarget.Android, ".apk", true); // there's the possibility of an error where it
```
fluong6, line 68: Incomplete comment here :P

fluong6 (approved):

LPLafontaineB (approved):

Assets/Scripts/Editor/BuildHelpers.cs
fluong6: oh nvm. i missed read

PR #805, opened 2023-02-03
feat: replacing polling for lobby updates with LobbyEvents [MTT-5425]

LPLafontaineB:
### Description

This PR removes the polling we are currently doing to get updates to the lobby we have joined. It instead replaces it with the LobbyEvent feature which uses Wire to create channels that clients can subscribe to. These channels then send updates only when it is needed. In doing so, we save a lot of bandwidth to the Lobby Service.

This PR also fixes a bug in the UpdateRunner that was using the wrong delta time when updating subscribers. This was causing the Heartbeat request we meant to send regularly to the Lobby Service to be sent much less frequently.

### Issue Number(s)

[MTT-5425](https://jira.unity3d.com/browse/MTT-5425)

### Contribution checklist

### DIFF
--- a/Assets/Scripts/Infrastructure/UpdateRunner.cs
+++ b/Assets/Scripts/Infrastructure/UpdateRunner.cs
@@ -14,6 +14,7 @@ namespace Unity.BossRoom.Infrastructure
         {
             public float Period;
             public float NextCallTime;
+            public float LastCallTime;
         }
 
         readonly Queue<Action> m_PendingHandlers = new Queue<Action>();
@@ -56,7 +57,7 @@ namespace Unity.BossRoom.Infrastructure
                 {
                     if (m_Subscribers.Add(onUpdate))
                     {
-                        m_SubscriberData.Add(onUpdate, new SubscriberData() { Period = updatePeriod, NextCallTime = 0 });
+                        m_SubscriberData.Add(onUpdate, new SubscriberData() { Period = updatePeriod, NextCallTime = 0, LastCallTime = Time.time });
                     }
                 });
             }
@@ -90,7 +91,8 @@ namespace Unity.BossRoom.Infrastructure
 
                 if (Time.time >= subscriberData.NextCallTime)
                 {
-                    subscriber.Invoke(Time.deltaTime);
+                    subscriber.Invoke(Time.time - subscriberData.LastCallTime);
+                    subscriberData.LastCallTime = Time.time;
                     subscriberData.NextCallTime = Time.time + subscriberData.Period;
                 }
             }
--- a/Assets/Scripts/UnityServices/Lobbies/LobbyAPIInterface.cs
+++ b/Assets/Scripts/UnityServices/Lobbies/LobbyAPIInterface.cs
@@ -133,5 +133,10 @@ namespace Unity.BossRoom.UnityServices.Lobbies
         {
             await LobbyService.Instance.SendHeartbeatPingAsync(lobbyId);
         }
+
+        public async Task<ILobbyEvents> SubscribeToLobby(string lobbyId, LobbyEventCallbacks eventCallbacks)
+        {
+            return await LobbyService.Instance.SubscribeToLobbyEventsAsync(lobbyId, eventCallbacks);
+        }
     }
 }
--- a/Assets/Scripts/UnityServices/Lobbies/LobbyServiceFacade.cs
+++ b/Assets/Scripts/UnityServices/Lobbies/LobbyServiceFacade.cs
@@ -37,7 +37,6 @@ namespace Unity.BossRoom.UnityServices.Lobbies
 
         public Lobby CurrentUnityLobby { get; private set; }
 
-        LobbyEventCallbacks m_LobbyEventCallbacks;
         ILobbyEvents m_LobbyEvents;
 
         bool m_IsTracking = false;
@@ -90,8 +89,6 @@ namespace Unity.BossRoom.UnityServices.Lobbies
             var task = Task.CompletedTask;
             if (CurrentUnityLobby != null)
             {
-                CurrentUnityLobby = null;
-
                 var lobbyId = m_LocalLobby?.LobbyID;
 
                 if (!string.IsNullOrEmpty(lobbyId))
@@ -105,9 +102,6 @@ namespace Unity.BossRoom.UnityServices.Lobbies
                         task = LeaveLobbyAsync(lobbyId);
                     }
                 }
-
-                m_LocalUser.ResetState();
-                m_LocalLobby?.Reset(m_LocalUser);
             }
 
             m_LobbyEvents?.UnsubscribeAsync();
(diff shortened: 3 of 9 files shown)


Assets/Scripts/UnityServices/Lobbies/LobbyAPIInterface.cs
```diff
@@ -133,5 +133,10 @@ public async void SendHeartbeatPing(string lobbyId)
         {
             await LobbyService.Instance.SendHeartbeatPingAsync(lobbyId);
         }
+
+        public async Task<ILobbyEvents> SubscribeToLobby(string lobbyId, LobbyEventCallbacks eventCallbacks)
+        {
+            return await LobbyService.Instance.SubscribeToLobbyEventsAsync(lobbyId, eventCallbacks);
+        }
```
fernando-cortez, line 140: Not this line, but GetLobby(..) is now unused. I'll leave it up to you if you want to remove it.

Assets/Scripts/UnityServices/Lobbies/LobbyServiceFacade.cs
```diff
@@ -77,88 +81,45 @@ public void BeginTracking()
             if (!m_IsTracking)
             {
                 m_IsTracking = true;
```
fernando-cortez, line 83: This is more of an inquiry. I notice that some connection states begin the tracking of a lobby, and therefore begin the subscription of lobby events. Since lobby create/join/delete requests are queried here, and now we have lobby callbacks, what prevents us from beginning tracking here (and potentially ending tracking)?

LPLafontaineB: You're right, we could remove the BeginTracking stuff and simply do it automatically when joining a lobby. I explored it at first but I decided instead to keep it how it is to keep symmetry with ending tracking that has to be called manually in OfflineState, and to reduce the amount of changes in this PR. If I am to change this, I think I might take the time to refactor a bit of the code around it, i.e. removing the other calls to LobbyServiceFacade in the LobbyUIMediator and doing it internally instead. Thinking back on it, I think it would be worth it and remove some unneeded dependencies

Assets/Scripts/UnityServices/Lobbies/LobbyAPIInterface.cs
LPLafontaineB: Since this is an API interface, it could make sense to keep it? But honestly since we will likely never use it bc LobbyEvents are so much better in our use case I think we can just remove it

Assets/Scripts/UnityServices/Lobbies/LobbyServiceFacade.cs
LPLafontaineB: Do you have an opinion on this @SamuelBellomo ?

LPLafontaineB: Coming back to this, I think that actually this is out of scope for this PR. The goal here is to replace polling with events, and I think it is better to keep this self-contained. I'll create detailed tickets for further improvements so someone could pick it up when they have time.

fernando-cortez: Want to link the ticket? I think that's fair too. I'll approve after!

LPLafontaineB: https://jira.unity3d.com/browse/MTT-5544

Assets/Scripts/UnityServices/Lobbies/LobbyServiceFacade.cs
```diff
@@ -264,6 +225,68 @@ public async Task<(bool Success, Lobby Lobby)> TryQuickJoinLobbyAsync()
...
+        void OnLobbyEventConnectionStateChanged(LobbyEventConnectionState lobbyEventConnectionState)
+        {
+            m_LobbyEventConnectionState = lobbyEventConnectionState;
+            Debug.Log($"LobbyEventConnectionState changed to {lobbyEventConnectionState}");
+        }
+
+        async void SubscribeToJoinedLobby()
+        {
+            var lobbyEventCallbacks = new LobbyEventCallbacks();
+            lobbyEventCallbacks.LobbyChanged += OnLobbyChanges;
```
SamuelBellomo, line 283: mmm I'm scared of those callbacks with no associated -=
I'm assuming the below SubscribeToLobby takes ownership and just handles that for you?
(You can resolve this shouldn't be a blocking comment)

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 Additional documentation and release notes are available at [Multiplayer Documentation](https://docs-multiplayer.unity3d.com).
 
 ## [unreleased] - yyyy-mm-dd
+
+### Changed
+* Replaced our polling to obtain updates to the lobby that a client or host has joined with a subscrition to LobbyEvents (#805). This saves up a significant amount of bandwidth usage to and from the service, since updates are infrequent in this game. Now clients and hosts only use up bandwidth on the Lobby service when it is needed.
```
SamuelBellomo, line 12:
```suggestion
* Replaced our polling for lobby updates with a subscription to the new Websocket based LobbyEvents (#805). This saves up a significant amount of bandwidth usage to and from the service, since updates are infrequent in this game. Now clients and hosts only use up bandwidth on the Lobby service when it is needed.
```

SamuelBellomo: I looked at your back of the napkin calculations in the jira comments. It'd be great to add a summary here.
Quick note, I don't think you need to go as far as calculating game length. Just going with user count, all doing requests every 2 seconds would be enough, no need to calculate number of games. You don't care whether they are the same users, just that on average you'll have x amount of users making poll requests every 2 seconds.
I'd validate my calculations in the lobby channels, just to make sure we didn't miss anything before sharing an actual $ amount.
So 725B * 30 * 100k CCU = 2.175GB per minute. Or 93TB per month.

Assets/Scripts/Infrastructure/UpdateRunner.cs
```diff
@@ -56,7 +57,7 @@ public void Subscribe(Action<float> onUpdate, float updatePeriod)
                 {
                     if (m_Subscribers.Add(onUpdate))
                     {
-                        m_SubscriberData.Add(onUpdate, new SubscriberData() { Period = updatePeriod, NextCallTime = 0 });
+                        m_SubscriberData.Add(onUpdate, new SubscriberData() { Period = updatePeriod, NextCallTime = 0, LastCallTime = Time.time });
```
SamuelBellomo, line 60: is this fixing the useless lobby updates issue you were talking about in your jira comments?

SamuelBellomo: If so I'd try to mention this our changelog, just so users know to update their code for those that took inspiration from our code. They'll need to update their own code to make sure they don't do the same mistake

Assets/Scripts/UnityServices/Lobbies/LobbyServiceFacade.cs
```diff
@@ -77,88 +81,45 @@ public void BeginTracking()
...
-                if (e.Reason == LobbyExceptionReason.RateLimited)
-                {
-                    m_RateLimitQuery.PutOnCooldown();
-                }
-                else if (e.Reason != LobbyExceptionReason.LobbyNotFound && !m_LocalUser.IsHost) // If Lobby is not found and if we are not the host, it has already been deleted. No need to publish the error here.
-                {
-                    PublishError(e);
+                    catch (ObjectDisposedException e)
+                    {
+                        // This exception occurs in the editor when exiting play mode without first leaving the lobby.
```
SamuelBellomo, line 114: only when exiting playmode or actually OnApplicationQuit which could happen in builds as well?

SamuelBellomo (approved):

Assets/Scripts/Infrastructure/UpdateRunner.cs
LPLafontaineB: No, this addresses the issue causing the Heartbeat messages not being sent at the desired frequency to the Lobby service (mentioned in the PR description).
The unneeded updates issue was caused by us receiving the Lobby updates and systematically marking those as changing our local lobbies, so we would also send updates to Lobby each time we received an update, even if no change occurred. This is a nice catch though, because it seems I did not create a bug for this issue. This PR doesn't fix the issue, just greatly reduces its impact. I'll create a Jira ticket for it

Assets/Scripts/UnityServices/Lobbies/LobbyServiceFacade.cs
LPLafontaineB: Only when exiting play mode

Assets/Scripts/UnityServices/Lobbies/LobbyServiceFacade.cs
LPLafontaineB: Yes. The `LobbyEventCallbacks` object is passed to the `LobbyChannel` that is created internally by the Lobby SDK. This is then stored in `m_LobbyEvents`. The callbacks we register here are called by other callbacks managed internally by the `LobbyChannel` object. These are unsubscribed to when we call `m_LobbyEvents.UnsubscribeAsync()` and the `LobbyEventCallbacks` object gets garbage collected. When we subscribe again, we pass in a new `LobbyEventCallbacks` object.

CHANGELOG.md
LPLafontaineB: Good point! I'll add a streamlined version of these calculations here

Assets/Scripts/Infrastructure/UpdateRunner.cs
LPLafontaineB: [here](https://jira.unity3d.com/browse/MTT-5665) it is

SamuelBellomo (dismissed):

fernando-cortez (approved):

.yamato/mobile-build-and-run.yml
```diff
@@ -14,7 +14,7 @@ Build_Player_With_Tests_iOS_{{ project.name }}_{{ editor }}:
 
     commands:
         - pip install unity-downloader-cli==1.2.0 --index-url https://artifactory.prd.it.unity3d.com/artifactory/api/pypi/pypi/simple --upgrade
-        - unity-downloader-cli -c Editor -c iOS -u {{ editor }} --fast --wait
+        - unity-downloader-cli -c Editor -c iOS -u 2021.3.15f1 --fast --wait
```
SamuelBellomo, line 17: I can see this laying around and us forgetting to remove this. What's your plan so we don't forget to revert?

LPLafontaineB: Creating a high priority ticket in our backlog and linking it to the tools ticket for the bug on their side

LPLafontaineB: Here is the ticket: https://jira.unity3d.com/browse/MTT-6201

SamuelBellomo (approved):

PR #816, opened 2023-03-16
Update ConnectionMethod.cs

SamuelBellomo:
### Description

Small PR to add some comments to clarify, following Discord question.

### Contribution checklist

### DIFF
--- a/Assets/Scripts/ConnectionManagement/ConnectionMethod.cs
+++ b/Assets/Scripts/ConnectionManagement/ConnectionMethod.cs
@@ -66,6 +66,11 @@ namespace Unity.BossRoom.ConnectionManagement
             m_ConnectionManager.NetworkManager.NetworkConfig.ConnectionData = payloadBytes;
         }
 
+        /// Using authentication, this makes sure your session is associated with your account and not your device. This means you could reconnect 
+        /// from a different device for example. A playerId is also a bit more permanent than player prefs. In a browser for example, 
+        /// player prefs can be cleared as easily as cookies.
+        /// The forked flow here is for debug purposes and to make UGS optional in Boss Room. This way you can study the sample without 
+        /// setting up a UGS account.
         protected string GetPlayerId()
         {
             if (Services.Core.UnityServices.State != ServicesInitializationState.Initialized)


Assets/Scripts/ConnectionManagement/ConnectionMethod.cs
```diff
@@ -66,6 +66,11 @@ protected void SetConnectionPayload(string playerId, string playerName)
             m_ConnectionManager.NetworkManager.NetworkConfig.ConnectionData = payloadBytes;
         }
 
+        /// Using authentication, this makes sure your session is associated with your account and not your device. This means you could reconnect 
+        /// from a different device for example. A playerId is also a bit more permanent than player prefs. In a browser for example, 
+        /// player prefs can be cleared as easily as cookies.
+        /// The forked flow here is for debug purposes and to make UGS optional in Boss Room. This way you can study the sample without 
+        /// setting up a UGS account.
```
LPLafontaineB, line 73: Not in the scope of this PR, but now that I look at it again, should we throw an exception when the services are initialized but we are not signed in? Instead we currently treat it the same way as when the services are not initialized. Even if I don't think this case ever happens in this sample, I think it would make more sense to throw here.

LPLafontaineB (approved):

Assets/Scripts/ConnectionManagement/ConnectionMethod.cs
SamuelBellomo: With this comment, we're telling users this method is a sample specific flow. I think we can trust our users to implement error handling on their own if they have that flow. I'd also implement it at the authentication call, not here. There might be ways to get signed out randomly with 3rd parties? Or if we had a signout button? But at that point that's starting to get out of scope of Boss Room, we only use anonymous signin, which would be per device. The only way to get signed out unexpectedly is if my account was deleted in some way?

LPLafontaineB: The line I had in mind is this one:
` return AuthenticationService.Instance.IsSignedIn ? AuthenticationService.Instance.PlayerId : ClientPrefs.GetGuid() + m_ProfileManager.Profile;`
I don't know what could cause a user to not be signed in anymore (or even if it is possible), but my point was that this line implied that this was possible and normal, and basically used what seems like a workaround instead of failing. But I agree this doesn't seem like a priority, especially with the above comment as you said, and it is probably not worth spending time on it.

SamuelBellomo, line 73:
```suggestion
        /// setting up a UGS account. It's recommended to investigate your own initialization and IsSigned flows to see if you need 
        /// those checks on your own and react accordingly. We offer here the option for offline access for debug purposes, but in your own game you
        /// might want to show an error popup and ask your player to connect to the internet.
```

SamuelBellomo: added a line explaining potential issues with this

PR #822, opened 2023-04-19
fix: Adding INetworkSerializeByMemcpy where recommended for network serialization [MTT-5912]

fernando-cortez:
### Description
`INetworkSerializeByMemcpy` interface is recommended usage for network messages containing value types. It for allows some bandwidth optimization techniques (byte packing) that are otherwise unavailable.

It is not recommended in places where object types are serialized, nor when FixedStrings are used.

`INetworkSerializeByMemcpy` [when used with FixedStrings] struct is not recommended because it will lose the bandwidth optimization - an empty FixedString128Bytes serialized normally or through `INetworkSerializable` will use 4 bytes of bandwidth, but inside an `INetworkSerializeByMemcpy`, that same empty value would consume 132 bytes of bandwidth.

### Issue Number(s)
[MTT-5912](https://jira.unity3d.com/browse/MTT-5912)

### Contribution checklist

### DIFF
--- a/Assets/Scripts/Gameplay/Action/ActionID.cs
+++ b/Assets/Scripts/Gameplay/Action/ActionID.cs
@@ -7,14 +7,9 @@ namespace Unity.BossRoom.Gameplay.Actions
     /// This struct is used by Action system (and GameDataSource) to refer to a specific action in runtime.
     /// It wraps a simple integer.
     /// </summary>
-    public struct ActionID : INetworkSerializable, IEquatable<ActionID>
+    public struct ActionID : INetworkSerializeByMemcpy, IEquatable<ActionID>
     {
         public int ID;
-
-        public void NetworkSerialize<T>(BufferSerializer<T> serializer) where T : IReaderWriter
-        {
-            serializer.SerializeValue(ref ID);
-        }
 
         public bool Equals(ActionID other)
         {
--- a/Assets/Scripts/Infrastructure/NetworkGuid.cs
+++ b/Assets/Scripts/Infrastructure/NetworkGuid.cs
@@ -3,16 +3,10 @@ using Unity.Netcode;
 
 namespace Unity.BossRoom.Infrastructure
 {
-    public class NetworkGuid : INetworkSerializable
+    public struct NetworkGuid : INetworkSerializeByMemcpy
     {
         public ulong FirstHalf;
         public ulong SecondHalf;
-
-        public void NetworkSerialize<T>(BufferSerializer<T> serializer) where T : IReaderWriter
-        {
-            serializer.SerializeValue(ref FirstHalf);
-            serializer.SerializeValue(ref SecondHalf);
-        }
     }
 
     public static class NetworkGuidExtensions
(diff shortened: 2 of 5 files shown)


Assets/Scripts/Gameplay/Action/ActionID.cs
```diff
@@ -7,15 +7,10 @@ namespace Unity.BossRoom.Gameplay.Actions
     /// This struct is used by Action system (and GameDataSource) to refer to a specific action in runtime.
```
SamuelBellomo, line 7: LobbyPlayerState can't be converted as well? Does the name string property mess with it?
The player name inside, no, but the containing state?

Assets/Scripts/Infrastructure/NetworkGuid.cs
```diff
@@ -3,16 +3,10 @@
```
SamuelBellomo, line 3: adding note here, let's not forget the checklist for PRs, I'd expect a changelog for this

Assets/Scripts/Gameplay/Action/ActionID.cs
fernando-cortez: Converting LobbyPlayerState over to implement `INetworkSerializeByMemcpy`  does work, but it'd introduce bandwidth overhead that I don't think is worth having performant serailzation. Per offline discussion:

> We try to encourage the use of FixedString for all network purposes because of this (but putting FixedString inside an INetworkSerializeByMemcpy struct is not recommended because it will lose the bandwidth optimization - an empty FixedString128Bytes serialized normally or through INetworkSerializable will use 4 bytes of bandwidth, but inside an INetworkSerializeByMemcpy, that same empty value would consume 132 bytes of bandwidth

SamuelBellomo: The fixed string is wrapped by an INetworkSerializable though, do we lose that optim if that wrapper is then inside a INetworkSerializeByMemcpy ?

SamuelBellomo: ah right, it copies the whole struct, byte by byte and won't call the serialize method in the child struct. nvm then

SamuelBellomo: it could be worth adding a quick comment in our LobbyPlayerState struct?

CHANGELOG.md
```diff
@@ -10,13 +10,12 @@ Additional documentation and release notes are available at [Multiplayer Documen
...
+
 ### Cleanup
 * Clarified a TODO comment inside ClientCharacter, detailing how anticipation should only be executed on owning client players (#786)
-* Removed now unnecessary cached NetworkBehaviour status on some components, since they now do not allocate memory (#799)
-
-### Changed
-* Simplified reconnection flow by offloading responsibility to ConnectionMethod (#804). Now the ClientReconnectingState uses the ConnectionMethod it is configured with to handle setting up reconnection (i.e. reconnecting to the Lobby before trying to reconnect to the Relay server if it is using Relay and Lobby). It can now also fail early and stop retrying if the lobby doesn't exist anymore.
+* Removed now unnecessary cached NetworkBehaviour status on some components, since they now do not allocate memory (#799) 
+* Certain structs converted to implement interface INetworkSerializeByMemcpy instead of INetworkSerializable (#822) INetworkSerializeByMemcpy optimizes for performance at the cost of bandwidth usage and flexibility, however it will only work with structs containing value types.
```
SamuelBellomo, line 18:
```suggestion
* Certain structs converted to implement interface INetworkSerializeByMemcpy instead of INetworkSerializable (#822) INetworkSerializeByMemcpy optimizes for performance at the cost of bandwidth usage and flexibility, however it will only work with structs containing value types. For me details see the official [doc](https://docs-multiplayer.unity3d.com/netcode/current/advanced-topics/serialization/inetworkserializebymemcpy/index.html)
```

SamuelBellomo (approved):

CHANGELOG.md
SamuelBellomo, line 18: wanna add an index entry as well?

LPLafontaineB (approved):

PR #836, opened 2023-05-08
feat: NetworkTransforms throughout project optimized where possible [MTT-6312]

fernando-cortez:
### Description
This PR upgrades NetworkTransforms inside the project to reduce bandwidth cost.

Prefabs/ scene objects that had Synchronize Transform enabled inside their NetworkObject component that didn't need to synchronize position have now that flag disabled. This is particularly useful if the NetworkObject is used more for management related tasks and has no spatial synchronization needs. This will reduce initial synchronization payloads.

CharacterPrefab prefab which all PCs & NPCs derive from now have on their NetworkTransform:
* sync only x and z positions
* sync only z rotation
* have half float precision enabled

The savings as seen from a client are as follows:

Pre:
pre-NT-optims

Only syncing x and z positions explicitly:
NT-optims-only-x-and-z-positions

Only syncing x and z positions, y rotation, and half float precision enabled:
NT-optims-only-x-and-z-positions-and-half-precision

Only syncinc x and z positions, half float precision enabled, quat syncing and quat compression enabled:
NT-optims-quat-sync-and-quat-compression-and-half-precision

Bandwidth costs considered, since only the y position is modified per frame in a worst case scenario, it is more cost efficient to only: sync x and z positions, y rotation, and enable half float precision. Individual NetworkTransform packet sizes are reduced from 21 bytes down to 16 bytes with this optimization.

### Issue Number(s)
[MTT-6312](https://jira.unity3d.com/browse/MTT-6312)

### Contribution checklist


NoelStephensUnity:
This looks about right with only synchronizing x and z position axis and the y rotation axis. Just using half float precision for this scenario is the right choice. You should get a 6 byte reduction for the 3 axis (2 bytes per as opposed to 4), but will get an additional 1 byte hit due to the m_BitSet flags and how that is being compressed...which should get you 5 bytes per NT savings.
👍

Assets/Prefabs/Character/Character.prefab
```diff
@@ -231,7 +234,7 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   SyncPositionX: 1
-  SyncPositionY: 1
+  SyncPositionY: 0
```
SamuelBellomo, line 237: for tracking purposes, investigating the impact of cutting y axis

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ Additional documentation and release notes are available at [Multiplayer Documen
 
 ## [unreleased] - yyyy-mm-dd
 
+### Changed
+* NetworkTransform bandwidth optimizations applied to NetworkObject prefabs inside project (#836) Netcode for GameObjects v1.4.0 introduced bandwidth compression techniques to further reduce the bandwidth footprint of a NetworkTransform's message packet. Inside Boss Room, several NetworkTransforms have now their "Synchronize Transform" flag now disabled for their NetworkTransform, meaning that its NetworkTransform properties will not be synced when spawning and/or when late-joining clients connect. This is particularly useful if the NetworkObject is used more for management related tasks and has no spatial synchronization needs. Additionally, the base prefab for PCs and NPCs in-game, Character, had its NetworkTransform modified to now utilize half float precision, ie. "Use Half Float Precision" set to true. Its y position is also explicitly no longer synced. This results in a net 5 byte reduction in a NetworkTransform's sync packet. For more information, see [Netcode for GameObjects' v1.4.0 release notes](https://github.com/Unity-Technologies/com.unity.netcode.gameobjects/releases/tag/ngo%2F1.4.0).
```
SamuelBellomo, line 12: isn't that flag on NetworkObjects and not on NetworkTransform? if you have an NO with no spatial needs, why have a NT at all.
I also wouldn't use "packet" for NT, since UnityTransport will batch stuff in the background. It's not one network packet per state update.

Assets/Prefabs/Game/Arrow.prefab
```diff
@@ -80,17 +83,21 @@ MonoBehaviour:
   SyncPositionX: 1
   SyncPositionY: 0
   SyncPositionZ: 1
-  SyncRotAngleX: 0
+  SyncRotAngleX: 1
```
SamuelBellomo, line 86: do we need those other axis? Does it change anything actually, with the quaternion optims available?

NoelStephensUnity: When quaternion synchronization is set, all axis are synchronized and the NetworkTransformEditor inspector view disables the selection of those axis (i.e. it shouldn't matter what those values are). 
_With the quaternion compression enabled, it reduces the entire per delta update from 12 bytes (i.e. all Euler axis) down to 4 bytes. So, the savings should be from around 21 bytes to roughly ~12 bytes per arrow update depending upon how good of compression the tick and bitset flags are._

SamuelBellomo: ok ok I see. I was just reviewing from github diff, so didn't see the UI change :P 
I'm wondering, if we were to use per axis rot syncing and half floats, wouldn't that be reduced to 2 bytes instead of 4? Since we're only rotating around y?

NoelStephensUnity: If the arrow only uses the y axis for rotation then yes... you would want to do that instead. _(and the savings would be the same as the Character.prefab)_

CHANGELOG.md
fernando-cortez: Oh good catch. That was meant to be `NetworkObjects have now their "Synchronize Transform" properties disabled for their NetworkObject component`. This should be mentioned as an "Additionally,..." comment. I'll fix.

I'll rephrase that as the NetworkTransform synchronization payload.

Assets/Prefabs/Character/Character.prefab
fernando-cortez: No impact! On the server, PCs and NPCs are placed at a y position that is non-zero (0.0003771782) just due to nav mesh bounds of the player. Clients are spawned with 0 for height, and that difference is ultimately negligible to gameplay and look. If we had ramps, or more complex nav meshes, then we would need to sync y positions, but as the level is all flat, there is no need to sync y position.

CHANGELOG.md
SamuelBellomo: still using "packet"?

PR #860, opened 2023-08-15
feat: upgrade to lobby 1.1.0 [MTT-7092] [MTT-5665] [MTT-7101] [MTT-5708]

LPLafontaineB:
### Description

This PR upgrades BR to use Lobby 1.1.0. It leverages a small new lobby feature to lock lobbies at creation and unlock them later when the server is created and Relay is ready to allow clients to join. This prevents clients from joining too early which would lead to an exception about the relay allocation not being found (while possible, this edge-case was still very unlikely, we haven't noticed it in any playtest).

This PR also fixes an issue where we would automatically send back an update request to Lobby whenever updating our local lobby data, even when that update came from the service. Since there are no changes to this data that needs to be updated during a game session, the periodic update mechanism has been removed entirely. The only updates we need to send are when setting up the relay integration, which is done when starting the host and when joining as a client.

It also removes the explicit reference we have to Wire in our package manifest, which we do not need since Wire is a dependency of Lobby.

### Issue Number(s)

[MTT-7092](https://jira.unity3d.com/browse/MTT-7092), [MTT-5665](https://jira.unity3d.com/browse/MTT-5665), [MTT-7101](https://jira.unity3d.com/browse/MTT-7101), & [MTT-5708](https://jira.unity3d.com/browse/MTT-5708)

### Contribution checklist


Assets/Scripts/UnityServices/Lobbies/LobbyAPIInterface.cs
```diff
@@ -42,6 +42,7 @@ public async Task<Lobby> CreateLobby(string requesterUasId, string lobbyName, in
             CreateLobbyOptions createOptions = new CreateLobbyOptions
             {
                 IsPrivate = isPrivate,
+                IsLocked = true, // locking the lobby at creation to prevent other players from joining before it is ready
```
fernando-cortez, line 45: Just to confirm, the service disables this itself when the lobby allocation is ready?

fernando-cortez (commented):
I caught an exception as the host and on clients when hitting stop in the editor while in the character select screen. I'll send the log files offline.

Assets/Scripts/UnityServices/Lobbies/LobbyAPIInterface.cs
LPLafontaineB: Not the service itself, but we do disable this. In LobbyServiceFacade.UpdateLobbyDataAsync() we lock/unlock the lobby depending on if the relay join code is available. In the past this was also how the lobby was locked after creation (because the Lobby SDK didn't include the possibility to lock at creation time).  Although now that we are only sending a single update when the host is ready, we can make this more explicit

Assets/Scripts/UnityServices/Lobbies/LobbyServiceFacade.cs
```diff
@@ -270,7 +266,7 @@ async void SubscribeToJoinedLobbyAsync()
             // The LobbyEventCallbacks object created here will now be managed by the Lobby SDK. The callbacks will be
             // unsubscribed from when we call UnsubscribeAsync on the ILobbyEvents object we receive and store here.
             m_LobbyEvents = await m_LobbyApiInterface.SubscribeToLobby(m_LocalLobby.LobbyID, lobbyEventCallbacks);
-            m_JoinedLobbyContentHeartbeat.BeginTracking();
+            m_UpdateRunner.Subscribe(DoLobbyHeartbeat, 1.5f);
```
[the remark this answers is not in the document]
LPLafontaineB: Nice catch! Indeed clients don't need to send heartbeat pings. Looking at the logs it seems those errors are caused by something else though. I'll investigate

Assets/Scripts/UnityServices/Lobbies/LobbyAPIInterface.cs
fernando-cortez: Ah yeah I couldn't find the reference, but there it is. Awesome.

Assets/Scripts/UnityServices/Lobbies/LobbyServiceFacade.cs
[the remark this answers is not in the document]
LPLafontaineB: Should be fixed now

fernando-cortez (dismissed):
Not seeing the same errors as before. Looks good!

fernando-cortez (approved):

PR #864, opened 2023-08-24
fix: clarifying additive scene synchronization fix for reconnection flow [MTT-3450]

LPLafontaineB:
### Description

This PR clarifies the fix Boss Room is using to handle clients reconnecting to the server while keeping the same active scene and having additional scenes additively loaded.

### Issue Number(s)

[MTT-3450](https://jira.unity3d.com/browse/MTT-3450)

### Contribution checklist


NoelStephensUnity (dismissed):
Looks good LP!
Like the added comments, will help clarify that context.
Now, I just need to update the NGO documentation to reflect that same message.  😸

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoaderWrapper.cs
```diff
@@ -150,7 +150,7 @@ void OnSceneEvent(SceneEvent sceneEvent)
                     }
                     break;
                 case SceneEventType.LoadEventCompleted: // Server told client that all clients finished loading a scene
-                    // Only executes on client
+                    // Only executes on client or host
```
RikuTheFuffs, line 153: nice catch

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoaderWrapper.cs
```diff
@@ -159,14 +159,21 @@ void OnSceneEvent(SceneEvent sceneEvent)
...
+                        if (NetworkManager.SceneManager.ClientSynchronizationMode == LoadSceneMode.Single)
+                        {
+                            // If using the Single ClientSynchronizationMode, unload all currently loaded additive
+                            // scenes. In this case, we want the client to only keep the same scenes loaded as the
+                            // server. Netcode For GameObjects will automatically handle loading all the scenes that the
+                            // server has loaded to the client during the synchronization process. If the server's main
+                            // scene is different to the client's, it will start by loading that scene in single mode,
+                            // unloading every additively loaded scene in the process. However, if the server's main
+                            // scene is the same as the client's, it will not automatically unload additive scenes, so
+                            // we do it manually here.
```
RikuTheFuffs, line 174: just a note: for multi-line comments, it's recommended to use `/* multiline comments */` instead of a series of `// `, as it reduces the visual noise

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoaderWrapper.cs
```diff
@@ -159,14 +159,21 @@ void OnSceneEvent(SceneEvent sceneEvent)
...
+                            UnloadAdditiveScenes();
```
RikuTheFuffs, line 175: Is there a corner case for which this can cause a desync between the scenes loaded here and on the server? If not, all good for me.

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoaderWrapper.cs
LPLafontaineB: In this case I kept it this way for consistency with the rest of the project, but perhaps we could create a task to update all our comments at once?

Packages/com.unity.multiplayer.samples.coop/Utilities/SceneManagement/SceneLoaderWrapper.cs
LPLafontaineB: No. In this case, the flow is:

1. Client receives message from server telling it to start synchronizing. This SceneEvent is invoked
2. This code is executed, starting to unload all scenes that are additively loaded asynchronously
3. Since this is in the case we are using single mode for synchronization, the client will look at the server's main scene and, if it is different than its own, will initiate a load operation in single mode (which would also unload the additively loaded scenes anyways), or if the server's main scene is the same as the client's, the client will do nothing and skip to next step.
4. Once that the main scene is loaded, the client will then load any extra scenes that the server has loaded
5. Once the client has loaded all the scenes loaded by the server, it will invoke OnSynchronizationComplete locally and will tell the server.
6. The server will invoke OnSynchronizationComplete locally once it has received the info from the client. It will also check if some new NetworkObjects have been spawned/despawned during that process and send the right info to the client so that it can update itself.

See [this page](https://docs-multiplayer.unity3d.com/netcode/current/basics/scenemanagement/timing-considerations/index.html#client-synchronization) for more details and diagrams

LPLafontaineB: I suspect that if new scenes were loaded/unloaded on the server during the synchronization process, the server would be informing the client of those changes during step 6, right? @NoelStephensUnity

LPLafontaineB: So to answer more clearly, unloading locally loaded scenes on clients prior to synchronization in single mode will not create desyncs with the server, because the server will tell the client which scene to load after this anyways

RikuTheFuffs: Thanks, all good then!

noellie-velez (approved):

PR #867, opened 2023-08-29
fix: replace null-coalescencing and null-conditional operators [MTT-7125]

LPLafontaineB:
### Description

This PR replaces occurrences of the null-coalesing (`??=`) and null-conditional (`?.`) operators in the project. These operators can cause issues when used with types inheriting `UnityEngine.Object` because that type redefines the `==` operator to define when an object is null. This redefinition applies to regular null checks (`if foo == null`) but not to those operators, thus this could lead to unexpected behaviour.

While those operators were safely used within Boss Room, only with types that were not inheriting `UnityEngine.Object`, we decided to remove most usages for consistency. This will also help avoid accidental mistakes, such as a user reusing a part of this code, but modifying it so that one of those operators are used with a `UnityEngine.Object`.

The only cases where we kept the null-conditional operator were with events/delegates/actions, for which there is no risk of confusion.

### Issue Number(s)

[MTT-7125](https://jira.unity3d.com/browse/MTT-7125)

### Contribution checklist


Assets/Scripts/UnityServices/Lobbies/LobbyServiceFacade.cs
```diff
@@ -212,8 +212,14 @@ public async Task<(bool Success, Lobby Lobby)> TryQuickJoinLobbyAsync()
         void ResetLobby()
         {
             CurrentUnityLobby = null;
-            m_LocalUser.ResetState();
-            m_LocalLobby?.Reset(m_LocalUser);
+            if (m_LocalUser != null)
+            {
+                m_LocalUser.ResetState();
+            }
```
RikuTheFuffs, line 218: There was no null-conditional operator for m_LocalUser, was this change made on purpose?

Assets/Scripts/UnityServices/Lobbies/LobbyServiceFacade.cs
```diff
@@ -212,8 +212,14 @@ public async Task<(bool Success, Lobby Lobby)> TryQuickJoinLobbyAsync()
...
-            m_LocalUser.ResetState();
-            m_LocalLobby?.Reset(m_LocalUser);
+            if (m_LocalUser != null)
+            {
+                m_LocalUser.ResetState();
+            }
+            if (m_LocalLobby != null)
+            {
+                m_LocalLobby.Reset(m_LocalUser);
+            }
```
RikuTheFuffs, line 222: What if m_LocalUser is null? Should this be nested inside the other check?

Packages/com.unity.multiplayer.samples.coop/Utilities/Net/SessionManager.cs
```diff
@@ -56,7 +67,8 @@ public void DisconnectClient(ulong clientId)
                 // Mark client as disconnected, but keep their data so they can reconnect.
                 if (m_ClientIDToPlayerId.TryGetValue(clientId, out var playerId))
                 {
-                    if (GetPlayerData(playerId)?.ClientID == clientId)
+                    var playerData = GetPlayerData(playerId);
+                    if (playerData != null && playerData.Value.ClientID == clientId)
```
RikuTheFuffs, line 71: shouldn't this be `if (playerData != null && playerData.ClientID == clientId)` instead?
In the previous version, the `.Value` wasn't accessed explicitly.

Packages/com.unity.multiplayer.samples.coop/Utilities/Net/SessionManager.cs
```diff
@@ -70,7 +82,8 @@ public void DisconnectClient(ulong clientId)
                 if (m_ClientIDToPlayerId.TryGetValue(clientId, out var playerId))
                 {
                     m_ClientIDToPlayerId.Remove(clientId);
-                    if (GetPlayerData(playerId)?.ClientID == clientId)
+                    var playerData = GetPlayerData(playerId);
+                    if (playerData != null && playerData.Value.ClientID == clientId)
```
RikuTheFuffs, line 86: see comment above

Packages/com.unity.multiplayer.samples.coop/Utilities/Net/SessionManager.cs
```diff
@@ -251,7 +264,8 @@ void ClearDisconnectedPlayersData()
             foreach (var id in idsToClear)
             {
                 string playerId = m_ClientIDToPlayerId[id];
-                if (GetPlayerData(playerId)?.ClientID == id)
+                var playerData = GetPlayerData(playerId);
+                if (playerData != null && playerData.Value.ClientID == id)
```
RikuTheFuffs, line 268: see comment above

Assets/Scripts/UnityServices/Lobbies/LobbyServiceFacade.cs
LPLafontaineB: Yes I added this for consistency. Although now that I think back, this should never be null. It might be better to let it throw an exception if it is. Same with m_LocalLobby

Assets/Scripts/UnityServices/Lobbies/LobbyServiceFacade.cs
LPLafontaineB: see reply to above comment

Assets/Scripts/UnityServices/Lobbies/LobbyServiceFacade.cs
LPLafontaineB: I think this could apply to some of the other null checks as well

RikuTheFuffs: throwing an exception sounds good

noellie-velez (dismissed):

noellie-velez (approved):

PR #886, opened 2024-04-10
feat: ClientPlayerAvatarNetworkAnimator added to handle client synchronizations [MTT-3152]

fernando-cortez:
### Description
This PR rectifies two things:
* A coroutine was used inside `ServerAnimationHandler` to wait until `NetworkAnimator`'s Animator field had been assigned. That is no longer an issue with Netcode for GameObjects, and this `NetworkBehaviour` was also incorrectly playing an animation on newly joining clients on OnNetworkSpawn(). Instead, the `NetworkAnimator`'s synchronize step should have handled applying its sync data to clients. However, players need to instantiate a model before the synchronize data is applied, and so:
* `ClientPlayerAvatarNetworkAnimator` has been created to handle: instantiating the player model based on a networked GUID, rebinding this rig to the player's Animator, and applying synchronize data to said Animator.

To validate this fix:
* Host and connect 1 client to host
* Have either of the host or client faint
* Join with a second client, and see that the fainted client's animation is correctly synchronized via NetworkAnimator

### Issue Number(s)
[MTT-3152](https://jira.unity3d.com/browse/MTT-3152)

### Contribution checklist
 - [ N/A ] Tests have been added for boss room and/or utilities pack
 - [ N/A ] An Index entry has been added in readme.md if applicable


Assets/Scripts/Gameplay/GameplayObjects/Character/ClientPlayerAvatarNetworkAnimator.cs
```diff
@@ -0,0 +1,70 @@
...
+            base.OnNetworkDespawn();
+            m_AvatarInstantiated = false;
+            var avatarGraphics = Animator.transform.GetChild(0);
+            if (avatarGraphics != null)
+            {
+                Destroy(avatarGraphics.gameObject);
+            }
+        }
+
+        protected override void OnSynchronize<T>(ref BufferSerializer<T> serializer)
```
NoelStephensUnity, line 43: Nice way to do this! 👍 
I am planning on adding a prespawn and postspawn sometime in the next few weeks which would allow you to do this in the NetworkBehaviour.OnPrespawn method... but for now it is the best approach (imo).

Assets/Scripts/Gameplay/GameplayObjects/Character/ClientCharacter.cs
```diff
@@ -140,9 +140,9 @@ public override void OnNetworkSpawn()
             {
                 name = "AvatarGraphics" + m_ServerCharacter.OwnerClientId;
 
-                if (m_ServerCharacter.TryGetComponent(out ClientAvatarGuidHandler clientAvatarGuidHandler))
+                if (m_ServerCharacter.TryGetComponent(out ClientPlayerAvatarNetworkAnimator characterNetworkAnimator))
```
NoelStephensUnity, line 143: Minor thing...like super-minor...
If (in the event a user makes changes to a prefab) it cannot get the component...should it log an error/warning?
Does any other code try to access `m_ClientVisualsAnimator`?

The super nit-picky aspect is using TryGetComponent (as opposed to just GetComponent) without any handling of "what happens when the TryGetComponent fails". If there is no handling, then it might be better to just use GetComponent and allow the exception if something tries to use `m_ClientVisualsAnimator` or add an else with a message complaining about the component not being found.

Wouldn't block the PR for that...but figured worth pointing out the logic/usage conundrum/scenario.

NoelStephensUnity (dismissed):
Minor nit picky thing I commented on... 
But it looks good to me from a script perspective!  👍

Assets/Scripts/Gameplay/GameplayObjects/Character/ClientCharacter.cs
fernando-cortez: Thanks for the feedback!
We should definitely return a warning here if that's the case.

I've created [this task](https://jira.unity3d.com/browse/MTT-8378) to keep track of this work and include it in a future release!

Assets/Scripts/Gameplay/GameplayObjects/Character/ClientPlayerAvatarNetworkAnimator.cs
fernando-cortez: Awesome. Let us know when that lands, and we'll take a look! 😄

NoelStephensUnity (approved):
👍

PR #893, opened 2024-12-04
[MTTB-409] Fix a Healer ability doesn't work (#893)

nanho-lee:
### Description
This is the PR that fixed the following [issue](https://jira.unity3d.com/browse/MTTB-409).
To solve the above issue, I did three things.

### 1.
Fixed a bug where the character's orientation was flipped if the target and current location were almost the same when the skill was triggered. In that case, the character's orientation is used.
[ClientInputSender.cs](https://github.com/nanho-lee/Unity-Technologies/blob/bugfix/mttb-585/Assets/Scripts/GamePlay/UserInput/ClientInputSender.cs#L424-L427)
```csharp
//there is a bug where the direction is flipped if the hitPos and current position are almost the same,
//so we use the character's direction instead.
float directionLength = offset.magnitude;
Vector3 direction = float.Epsilon >= directionLength ? (offset / directionLength) : m_PhysicsWrapper.Transform.forward;
```
### 2.
mttb_00
mttb_01
Despite being close enough, the Box cast would often fail depending on the angle. To compensate for this, I added the Sphere cast.
I considered creating a new Action for this, but since the logic is almost identical to the MeleeAction, I decided to only use the Sphere cast when the Radius value is set.
### 3.
If there are multiple targets that are checked for collisions, I've added the GetTotalDamage API to IDamagable to select the most damaged target. I felt that this prioritization was a good one, whether the MeleeAction is attack or recovery.
Note that I kept the ability to prioritize targets with hint targets for targeting.

### Issue Number(s)
MTTB-409
https://jira.unity3d.com/browse/MTTB-409

### Contribution checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ Additional documentation and release notes are available at [Multiplayer Documen
 ### Cleanup
 * Removed ParrelSync from the project
 
+### Fixed
+* Fix a Healer ability doesn't work (#893) 
+
 
 ## [2.5.0] - 2024-04-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,10 @@ Additional documentation and release notes are available at [Multiplayer Documen
 * Removed ParrelSync from the project
 
 ### Fixed
-* Fix a Healer ability doesn't work (#893) 
+* Fix a Healer ability doesn't work (#893)
+  * Changed the way characters are oriented when using skills.
+  * Added the GetTotalDamage API to the IDamagble interface. This number is your maximum health minus your current health.
+  * Changed the way MeleeAction selects a target when there are multiple targets to collide with. The target with the highest GetTotalDamage value (mentioned above) will be selected.
 
 
 ## [2.5.0] - 2024-04-18
--- a/Assets/GameData/Action/Mage/MageHeal.asset
+++ b/Assets/GameData/Action/Mage/MageHeal.asset
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:6c16b018e4bfa01b23bea3e0f6ef366ac529f16a749e38c93384e1e38a5c7ab8
-size 1159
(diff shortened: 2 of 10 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=893) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=893) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/com.unity.multiplayer.samples.coop?pullRequest=893) it.</sub>

Elfi0Kuhndorf (dismissed):
Looks good! Works now as expected. Thank you for fixing this!

fernando-cortez (commented):
All that's left is to merge develop, and add a note in our changelog with this fix. Make sure to add the commit number to the changelog line -- the changelog should have examples of this too.
Looks great!

nanho-lee:
Thank you both for your reviews.
I have modified the log and CHANGELOG files as you said.

Elfi0Kuhndorf (dismissed):

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ Additional documentation and release notes are available at [Multiplayer Documen
 ### Cleanup
 * Removed ParrelSync from the project
 
+### Fixed
+* Fix a Healer ability doesn't work (#893)
```
fernando-cortez, line 24: After linking the PR number, we've tried to give a bit more context on the significance of the change and/or what has been the approach. Could we give a bit more context here on what may have been not working and potentially what was the high-level approach to the fix?

nanho-lee: Yes, of course.
I've been working on it.
If you have any requests, let me know.

fernando-cortez (approved):
Amazing!

